PR #87, opened 2014-09-11
Add support to replace php.ini settings.

jonascarlbaum:
Making it possible to replace **php.ini** settings through **Homestead.yml** like this:

```
php_ini_overrides:
    - max_execution_time: "300"
    - memory_limit: 2048M
    - post_max_size: 100M
    - upload_max_filesize: 100M   
```

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -57,6 +57,20 @@ class Homestead
           end
       end
     end
+    
+    # Override PHP.INI settings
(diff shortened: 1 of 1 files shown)


cnxh:
@taylorotwell the opcache is [enabled](http://php.net/manual/en/opcache.configuration.php) by default,  should it be disabled in homestead since homestead is a develop enviroment?

PR #91, opened 2014-09-12
Hosts file auto-update

LeShadow:
This piece of code will enable vagrant to auto-update the hosts file if
the hostsupdater plugin has been installed. 
If it has not been installed, it will not execute the code that wil add the domains to the hosts file.This can be enabled or disabled in the homestead.yaml file so it's up to the user to if he wants to use this or not.
(Supports: Windows, Mac OS X, Linux)

This takes out the fact that you manually have to edit your hosts file.

If this were to be accepted, you'd have to edit the documentation, take out the part about editing your hosts file and add the following:

> After you installed Vagrant, you will need to install an additional plugin that enables vagrant to automagically update your hosts file so you can access your applications more easily.
> 
> ``` bash
> vagrant plugin install vagrant-hostsupdater
> ```

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -59,7 +59,7 @@ class Homestead
     end
 
     # Updating the hosts file with all the sites that are defined in Homestead.yaml
-    if Vagrant.has_plugin?("vagrant-hostsupdater") && site["hosts_file_additions"] == true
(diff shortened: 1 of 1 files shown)


pariz:
Why wasn't this accepted?

PR #107, opened 2014-11-16
Added default folder map type to be nfs

dmyers:
Without this a simple hello world is very slow (1 second).

Probably due to my mapped code folder being very large with nested folders and this really provides a great experience.


taylorotwell:
NFS has the problem of files delaying to update on the guest, which is equally annoying as the slowness honestly.

dmyers:
@taylorotwell Is it usable out of the box? For me and others, it isn't and I don't get why.

PR #112, opened 2014-11-18
Add HTTPS support

mmeyer2k:
Following up where #83 left off. Added HHVM support.


mmeyer2k:
Because its not useful or what? Seems like pretty basic functionality. Why have Request::secure() without a way to test it locally before deployment?

PR #121, opened 2014-11-25
Listen for EDITOR environment variable if one is set

PatrickRose:
In the Arch Linux community at least, there's the usage of an EDITOR environment variable. It's preferable in my mind to have that be the first option to be listened to since that's pretty much always going to be the preferred editor.

If EDITOR is not set then it defaults back to start/open.

### DIFF
--- a/src/EditCommand.php
+++ b/src/EditCommand.php
@@ -44,6 +44,10 @@ class EditCommand extends Command {
 	 */
 	protected function executable()
 	{
+		if (getenv('EDITOR') !== false)
+		{
+			return getenv('EDITOR');
+		}
 		return strpos(strtoupper(PHP_OS), 'WIN') === 0 ? 'start' : 'open';
 	}
 


dakira:
@hailwood xdg-open will open a text file with the editor that is associated with text files. In my case that's sublime text.

Sequoia:
:+1: c'mon maintainers... throw linux users a bone here ;)

Sequoia:
with @andrewhood125's PR, this prefers `$EDITOR` & supports linux without it as well.

Sequoia:
ay i'll try to fix up these merge conflicts gimme a min...

src/EditCommand.php
```diff
@@ -44,6 +44,10 @@ public function execute(InputInterface $input, OutputInterface $output)
 	 */
 	protected function executable()
 	{
+		if (getenv('EDITOR') !== false)
+		{
+			return getenv('EDITOR');
+		}
 		return strpos(strtoupper(PHP_OS), 'WIN') === 0 ? 'start' : 'open';
```
GrahamCampbell: Please put an empty line before this return statement.

Sequoia:
@PatrickRose played around with this fixing up your PR for merge, but there are problems with Vim...
```
Vim: Warning: Output is not to a terminal
Vim: Warning: Input is not from a terminal
[terminal mayhem]
```
see fix here: http://stackoverflow.com/a/1690319

But I'm not sure how to make it work with `EDITOR=vim` **and** with `EDITOR=something_else`, inside symfony/Process, and adding a special case for vim seems like A Bad Idea. For me (ubuntu user) I reckon xdg is "Good Enough."  Also fwiw `gvim` works fine.

PR #127, opened 2014-11-28
Adding reload command to restart VM

maddhatter:


taylorotwell:
See recently added provision command.

maddhatter:
Reloading without the --provision parameter just reboots the machine, which there isn't really a way of doing other than `homestead halt && homestead up`

PR #131, opened 2014-12-02
Support of "default_server" NGINX config attribute

Claymm:
Here I simply give the ability to the Homestead.yaml file to set a default server. I was looking for a solution to use the ``vagrant share`` command. Vagrant share was working fine but as there is one unique domain name for multiple sites on my Vagrant box, NGINX was using the first server coming in the list by alphabetical order. With the ``default_server`` directive in the config file, I'm able to select which server would be used by default when I hit the box.

Here is how I do it from my ``Homestead.yml`` file now:
```yaml
sites:
    - map: myhomestead.app
      to: /home/vagrant/Code/myhomesteadapp/public
      hhvm: true
      default: true
    - map: anotherhomestead.app
      to: /home/vagrant/Code/anotherhomesteadapp/public
      hhvm: true
```

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -47,11 +47,17 @@ class Homestead
     settings["sites"].each do |site|
       config.vm.provision "shell" do |s|
           if (site.has_key?("hhvm") && site["hhvm"])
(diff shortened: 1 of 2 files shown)


Claymm:
@taylorotwell not interesting? It might be based on v1.x or v2.0 though.

PR #140, opened 2014-12-15
Fix whoops 404 catch by nginx 1.6.2

joecohens:
Since homestead 2.0 the nginx release versión is 1.6.2 for some reason the error_page 404 /index.php is not letting Whoops display. https://laracasts.com/discuss/channels/general-discussion/restore-notfoundhttpexception

Removing the directive makes it work again.

Thanks!


lstables:
+1

PR #144, opened 2014-12-23
Add blackfire configuration to Homestead.yaml file

haampie:
Since Homestead v0.2.2 brings Blackfire.io by default, it might be useful to configure it via the Homestead.yaml file.

Also, xdebug is disabled when Blackfire is enabled via the configuration. See https://blackfire.io/getting-started


taylorotwell:
What is the client ID and token?

GrahamCampbell:
https://blackfire.io/doc/configuration

taylorotwell:
​The word "client" appears once on that page and has nothing to do with
client IDs or tokens.​

haampie:
The client is a CLI tool for triggering the profiler and is useful for API apps in which you cannot use the Chrome extension to profile certain requests. See https://blackfire.io/doc/web-services for reference.

However, I agree it should not be added to Vagrant after all, since you should use it on the client side ;)

PR #147, opened 2015-01-04
Add "homestead share" command

vicdelfant:
This PR adds a `homestead share` command that can be used to share the box using `vagrant share`. Since the name of the share defaults to "share" it is advisable to register a [custom domain] (https://atlas.hashicorp.com/help/vagrant/shares/custom) with Atlas for privacy reasons.

The reason it defaults to "share" is to make it easier to add additional mappings for external access to sites, which would otherwise be pretty much impossible with Atlas' random hostnames. Hostnames are wildcard DNS records, so use `foo.share.custom-domain.tld`, `bar.share.custom-domain.tld` etc. as the hostnames in Homestead.yaml. You basically share your entire Homestead box, not just one specific site. Per-site access is done by using site-specific mappings using these subdomains.

### DIFF
--- a/homestead
+++ b/homestead
@@ -34,6 +34,7 @@ $app->add(new Laravel\Homestead\ResumeCommand);
 $app->add(new Laravel\Homestead\RunCommand);
 $app->add(new Laravel\Homestead\UpCommand);
 $app->add(new Laravel\Homestead\UpdateCommand);
+$app->add(new Laravel\Homestead\ShareCommand);
 $app->add(new Laravel\Homestead\SshCommand);
 $app->add(new Laravel\Homestead\StatusCommand);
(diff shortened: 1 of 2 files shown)


taylorotwell:
Honestly I found homestead share to be a pain in the rear especially when hosting multiple sites on one box. meetfinch.com works much better for Homestead's use case.

vicdelfant:
Thanks for the feedback. meetfinch.com looks nice but isn't our preferred solution since it requires a separate subscription (for static hostnames at least).

I've submitted a second PR (#182) with a far better integration, if you'd like to reconsider. Adding additional sites for sharing is no longer required and `homestead share` now also works with dynamic hostnames generated by Atlas.

PR #148, opened 2015-01-08
Serve script is failing when using oh my zsh under Homestead env

mvpasarel:
The `serve` command is not found when adding homestead aliases to zsh  `echo "source $HOME/.bash_aliases" >> /home/vagrant/.zshrc`:

```bash
/home/vagrant/.bash_aliases:8: parse error near `&&'
➜  ~  serve
zsh: command not found: serve
```

### DIFF
--- a/src/stubs/aliases
+++ b/src/stubs/aliases
@@ -5,7 +5,7 @@ alias h='cd ~'
 alias c='clear'
 
 function serve() {
-	if [[ "$1" && "$2" ]]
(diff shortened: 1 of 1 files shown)


taylorotwell:
Designed to work with bash.

mvpasarel:
This will work with bash as well.

bbashy:
Either way Taylor, it's best to make the code run properly on more shells.

PR #149, opened 2015-01-14
Added the ProvisionCommand class

andre-sa:
I've created the ProvisionCommand class so the Homestead machine could be provisioned without the need for shutdown or restart.


.gitignore
```diff
@@ -1,2 +1,3 @@
 /vendor
 /.vagrant
+/.idea
```
GrahamCampbell: please remove that

GrahamCampbell: ide ignoring should be done at the system level

andrewhood125:
Hey @andre-sa, you might want to do a `git reset --soft` back to 
when you added the ProvisionCommand. Then you can remove `.idea` from the `.gitignore` and redo your original commit so the contents of your `.idea` directory are not included in Homesteads history. Don't forget to system wide ignore `.idea`!

andre-sa:
I believe i've done it now. Thank for your help @GrahamCampbell and @andrewhood125.

milesj:
You misspelled provision, `provisiom`.

PR #150, opened 2015-01-20
Add provider selection parameter

jpwiersma:
- Adds a provider selection parameter for the up command with a default for virtualbox
(fixes issues for people using VMWare Fusion and allows for more standard parameters)

### DIFF
--- a/src/UpCommand.php
+++ b/src/UpCommand.php
@@ -17,7 +17,8 @@ class UpCommand extends Command {
 	{
 		$this->setName('up')
                   ->setDescription('Start the Homestead machine')
-				  ->addOption('provision', null, InputOption::VALUE_NONE, 'Run the provisioners on the box.');
+				  ->addOption('provision', null, InputOption::VALUE_NONE, 'Run the provisioners on the box.')
+				  ->addOption('provider', null, InputOption::VALUE_REQUIRED, 'Back the machine with a specific provider.', 'virtualbox');
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


taylorotwell:
Designed for virtualbox.

jpwiersma:
Understood that it's just for Virtualbox, but if you have VMWare Fusion provider installed with Vagrant, that becomes the default and Homestead doesn't work anymore. This commit makes sure that the Virtualbox provider is chosen by default.

drewhammond:
@taylorotwell Please take another look at this.

I just spent way too much time trying to figure out why this was happening...

``` sh
[19:40:33] drewhammond:~ $ homestead --version
Laravel Homestead version 2.0.9
```
``` sh
[19:40:46] drewhammond:~ $ homestead update -vvv
==> default: Box 'laravel/homestead' not installed, can't check for updates.
```

``` sh
[19:43:44] drewhammond:~ $ vagrant box list
chef/centos-6.5-i386                                                              (vmware_desktop, 1.0.0)
chef/ubuntu-14.04                                                                 (vmware_desktop, 1.0.0)
hashicorp/precise32                                                               (virtualbox, 1.0.0)
laravel/homestead                                                                 (virtualbox, 0.2.1)
laravel/homestead                                                                 (virtualbox, 0.2.2)
```

Turns out it was because my vagrant default provider was set to `vmware_desktop`. I must have hardcoded a fix for this before because I never had an issue until I updated homestead today and all the files were overwritten.

Even if we could just hardcode `--provider=virtualbox` and forget about making it configurable, it would prevent this from happening in the future.

svpernova09:
If the default was set to virtualbox, and we have people testing the other providers to ensure that Homestead works as intended on those providers, I wouldn't see the issue. The case may be supporting those unknown providers may not make sense unless there is reliable feedback.

huntwj:
The issue here, however, is that is I have another provider installed alongside virtual box, I cannot use homestead without hacking the homestead source code to specify virtual box as my provider. If you only want to support virtual box, fine, but don't make it impossible to pass other vagrant options to the command line. `homestead up` should pass along any extra parameters to the `vagrant up` command.

Just my $0.02.

PR #151, opened 2015-01-24
adds configuration option to import sql dump file when creating databases

ammonkc:
This adds a configuration options to automatically import a sql dump file when creating the databases.

Example Homestead.yam config:
databases:
    - db: homestead
      sql: /home/vagrant/Code/Laravel/homestead.sql

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -3,3 +3,7 @@
 DB=$1;
 mysql -uhomestead -psecret -e "DROP DATABASE IF EXISTS $DB";
 mysql -uhomestead -psecret -e "CREATE DATABASE $DB";
+
+if [[ $2 ]]; then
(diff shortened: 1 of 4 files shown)


andrewhood125:
Hmm. Is there something you're trying to do that can't be done with migrations or seeders?

ammonkc:
I realize that migrations and seeders are fine for most cases. But occasionally I'll need to debug an existing database in a dev vm. it may not be a common enough use case to merge it into the project. Thought I would through it out there just in case.

svpernova09:
I see your point, but I'm with @andrewhood125, this would just as easily be an import via mysql or other database frontend tool instead of going into the project.

ammonkc:
Yeah, you're right. It is very much an edge case to need this anyway.

PR #152, opened 2015-01-26
init.sh: dont overwrite Homestead.yaml if already exists

martinlindhe:
Added a check to save myself from world of pain

/m

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -2,7 +2,13 @@
 
 mkdir -p ~/.homestead
 
-cp src/stubs/Homestead.yaml ~/.homestead/Homestead.yaml
+yaml=~/.homestead/Homestead.yaml
+
+if [ -f $yaml ]; then
+	echo "$yaml found, not overwriting it!"
+else
+	cp src/stubs/Homestead.yaml $yaml
+fi;
 cp src/stubs/after.sh ~/.homestead/after.sh
 cp src/stubs/aliases ~/.homestead/aliases
 
--- a/init.sh
+++ b/init.sh
@@ -2,14 +2,28 @@
 
 mkdir -p ~/.homestead
 
-yaml=~/.homestead/Homestead.yaml
+homesteadRoot=~/.homestead
+
+yaml=$homesteadRoot/Homestead.yaml
+after=$homesteadRoot/after.sh
+aliases=$homesteadRoot/aliases
 
 if [ -f $yaml ]; then
 	echo "$yaml found, not overwriting it!"
 else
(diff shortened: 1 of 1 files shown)


andrewhood125:
According to [man cp](http://unixhelp.ed.ac.uk/CGI/man-cgi?cp) you can use the no clobber option to not overwrite an existing file. It might be a slightly cleaner solution. 

`cp --no-clobber src/stubs/Homestead.yaml ~/.homestead/Homestead.yaml`

While we're at it, how about `after.sh`  or `aliases`?

martinlindhe:
@andrewhood125 i don't mind using that instead. but how portable is it. i learned more than once that the system tools differ (gnu vs bsd ones) in command line switches. the If uses bash logic and should work across all OS:es which support bash (as required by line in the script)

for example; on OSX:

```
cp --no-clobber src dst
cp: illegal option -- -
```

I'll update the patch to not overwrite the others as well.

martinlindhe:
PS, it seems the "cp -i" could work?

BSD "cp":

     -i    Cause cp to write a prompt to the standard error output before copying
           a file that would overwrite an existing file.  If the response from
           the standard input begins with the character `y' or `Y', the file copy
           is attempted.  (The -i option overrides any previous -n option.)

GNU cp:
       -i, --interactive
              prompt before overwrite (overrides a previous -n option)

martinlindhe:
I went ahead and cleaned that up, now it runs like;

```
> bash init.sh
overwrite /Users/ml/.homestead/Homestead.yaml? (y/n [n]) n
not overwritten
overwrite /Users/ml/.homestead/after.sh? (y/n [n]) n
not overwritten
overwrite /Users/ml/.homestead/aliases? (y/n [n]) n
not overwritten
Homestead initialized!
```

I'm happy! You happy?

/m

andrewhood125:
@martinlindhe :+1:

andrewhood125:
@martinlindhe when would these files be overwritten through normal usage? Just checked this out to test it and noticed https://github.com/laravel/homestead/blob/master/src/InitCommand.php#L32 which checks if you have done a `php homestead init` already. 

Maybe it would be a good idea to allow the user to proceed but warn them that it's been initialized if new stubs are upstream. What do you think?

martinlindhe:
For me, i quickly skimmed the http://laravel.com/docs/4.2/homestead
under "Manually Via Git (No Local PHP)" (after realizing the homestead php util dont mix well with 'vagrant share'), and ran the suggested "bash init.sh"

PR #154, opened 2015-01-28
Updated src/stubs/aliases to include vendor_bin

gxela:
added vendor_bin alias, so to add vendor/bin to path when run from root of project

### DIFF
--- a/src/stubs/aliases
+++ b/src/stubs/aliases
@@ -16,3 +16,10 @@ function serve() {
 	fi
 }
 
+function vendor_bin() {
+        if [ -d "./vendor/bin" ]; then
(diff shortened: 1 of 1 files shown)


svpernova09:
I think I see what you're trying to do, but you're not using this anywhere. I would include a backwards compatible use case so the end user could see the intended use.

matteoantoci:
Nice!

gxela:
yes, I do use it. once I am inside the homestead machine in the project directory I type vendor_bin and it adds all those great tools into my PATH

gxela:
Is this going to get merged? Over the years I have commited a few changes to Laravel, and you guys don't merge them. If you don't merge this, then you suck and so does Laravel.

PR #155, opened 2015-01-30
Pass environment variables to 'provision' process

onilicious:
Previously, calling 'homestead provision' will not provision the homestead VM because vagrant does not know where to locate the VM.

This commit simply passes those environment variables to `vagrant
provision` so that vagrant will know how to provision the homestead VM.


thelfensdrfer:
+1. Pull request fixes the error that the vm is not created because of the unknown location.

spaceemotion:
+1 As well, this fixes the `homestead provision` error for me as well.

paulrose:
+1

Was just about to create a PR for this :)

PR #156, opened 2015-02-05
Provision command not setting env.

tiran133:
This prevents the command from running.
Vagrant cant find the machine
==> default: VM is not currently running. Please, first bring it up with `vagrant up` then run this command.


tiran133:
Duplicate PR Sorry for that. 4292a79bfb6293c563675671b9398872d9359aa7

PR #158, opened 2015-02-06
UTF-8 charset and collation for MySQL

akurtz:
MySQL's built-in default for new databases is `latin1` / `latin1_swedish_ci`.  Can we make `utf8` / `utf8_unicode_ci` the default for Homestead's MySQL databases?  Specifying it at database creation seems like the simplest solution to me.

We could also/instead modify the Homestead.yaml syntax to specify these.  For example...

```
databases:
    - name: homestead
      charset: utf8
      collation: utf8_unicode_ci
    - name: otherdb
      charset: latin1
      collation: latin1_swedish_ci
```

But I'd still argue that UTF-8 should be the default.

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -2,4 +2,4 @@
 
 DB=$1;
 mysql -uhomestead -psecret -e "DROP DATABASE IF EXISTS $DB";
-mysql -uhomestead -psecret -e "CREATE DATABASE $DB";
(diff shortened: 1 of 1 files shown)


matteoantoci:
+1

IvanDokov:
I think this will be useful too!

jaoued:
:+1:

b3nl:
+1

svpernova09:
IMHO There should be a default and these charset settings should be optional. Just cleans up the config a bit if the user doesn't care. 

Looks like there has been an update thus causing a conflict, you should resolve this in your branch so it would be easier to merge.

PR #159, opened 2015-02-06
Automatically add hosts to hostsfile

ArjandeV:
Using vagrant plugin:
https://github.com/cogitatio/vagrant-hostsupdater

### DIFF
--- a/init.sh
+++ b/init.sh
@@ -6,4 +6,7 @@ cp src/stubs/Homestead.yaml ~/.homestead/Homestead.yaml
 cp src/stubs/after.sh ~/.homestead/after.sh
 cp src/stubs/aliases ~/.homestead/aliases
 
+# Install hostsupdater to automatically add hosts to /etc/hosts or C:\Windows\System32\drivers\etc\hosts
+vagrant plugin install vagrant-hostsupdater
+
 echo "Homestead initialized!"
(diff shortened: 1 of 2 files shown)


mmeyer2k:
PR #91 did exactly the same thing but was recently closed.

erikbelusic:
would love this as well!

vvv (wordpress dev vagrant) has this feature and it is awesome to never have to open up the hosts file. makes starting new projects so much quicker.

svpernova09:
IMHO it's bad form to make Homestead depend on vagrant plugins. It also would be concerning to me that Homestead would be touching my hosts file. If I wanted this functionality I'd do it outside of Homestead.

dmyers:
This is a great feature and I've used it to setup 1 click setup for non devs, but for people who want to contribue to html and css changes. You have to think about those who don't know what `/etc/hosts` is. I would think we would want to drive to near complete automation as possible. I tried to handoff the Laravel Homestead docs and it wasn't enough, if you want an entire team working on a project it has to just be `vagrant up` somehow.

Would you just clone homestead per docs and customize and everything? Problem is then you can't easily upgrade (points to Laravel 5).

It would be awesome if this was possible without changes to homestead.rb as in some kind of after hook (not after.sh), but like provisioned.rb or something which would let anyone add this functionality, or even homestead plugins where people could load up to extend it to whatever you guys don't care to merge.

richdynamix:
Rather than relying on homestead to modify the host machine it would make more sense to use something like DNSmasq (http://www.thekelleys.org.uk/dnsmasq/doc.html) to automatically route your host to the guest. 

```
192.168.10.10 *.dev
```

This can be achieved by setting the configuration to something like

```
address=/dev/192.168.10.10
```

PR #160, opened 2015-02-12
Allows changing of site ports.

zgosalvez:
Useful for us who use external devices (mobile apps) to access the VM.

This is backwards compatible, since I supplied the default port 80.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -54,11 +54,11 @@ class Homestead
     settings["sites"].each do |site|
       config.vm.provision "shell" do |s|
           if (site.has_key?("hhvm") && site["hhvm"])
-            s.inline = "bash /vagrant/scripts/serve-hhvm.sh $1 \"$2\""
-            s.args = [site["map"], site["to"]]
+            s.inline = "bash /vagrant/scripts/serve-hhvm.sh $1 \"$2\" $3"
+            s.args = [site["map"], site["to"], site["port"] ||= 80]
           else
-            s.inline = "bash /vagrant/scripts/serve.sh $1 \"$2\""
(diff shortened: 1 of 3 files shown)


svpernova09:
:+1: 

Just curious: why wouldn't your mobile app use port 80?

zgosalvez:
@svpernova09 It would, by default. But it wouldn't always work, especially with multiple sites on the same VM, since sites are mapped using vhosts on the local machine only. Using ports to map websites is also an alternative to configuring vhosts, as some people have problems with it.

@taylorotwell thanks!

karanits:
Just a heads up. We also have to update the serve command in the vagrant box( when directly using serve on it). Serve script was not working and zeroed in on this update as the reason.

zgosalvez:
@karanits Why the need to do `php artisan serve` on Homestead?

andrewhood125:
@zgosalvez could be wrong but I think @karanits might be talking about `bash /vagrant/scripts/serve.sh` not `./artisan serve`

karanits:
Sorry for late reply. the ```serve``` function defined in ```~/.bash_aliases``` of the VM also needs to be updated

```
function serve() {

        if [[ "$1" && "$2" && "$3" ]]
        then
                sudo dos2unix /vagrant/scripts/serve.sh
                sudo bash /vagrant/scripts/serve.sh "$1" "$2" "$3"
        else
                echo "Usage: "
                echo "  serve domain path port"
        fi
}
```

zgosalvez:
@karanits I don't particularly use this command, so I really can't address this right now. It also appears that this has already been patched https://github.com/laravel/homestead/commit/6e110aef9fb84a670151f776de1ccc0a79545bbe.

PR #161, opened 2015-02-13
Generate SSL-DEV

juliolustosa:
A suggestion to create sites automatically with ssl for testing.
Congratulations on the excellent work, I am a big fan of Laravel


mmeyer2k:
I submitted a similar PR (#112) recently which was closed.

svpernova09:
I think this PR is a bit cleaner than #112. I'd like to see the CSR variables be extracted so they could easily be set or fully generalized.

:+1: this is a decent option to enable SSL

taylorotwell:
Yeah, don't like the hard-coded CSR stuff and all that. Also seems like you could just redirect to the 443 server instead of duplicating blocks.

juliolustosa:
I understand , thank you

PR #163, opened 2015-02-19
Escape database name in MySQL script

felixkiss:
Otherwise it will break, if the database name contains a period.

Breaking `Homestead.yaml`:

```yaml
databases:
    - homestead
    - example.com
```

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
 DB=$1;
-mysql -uhomestead -psecret -e "DROP DATABASE IF EXISTS $DB";
-mysql -uhomestead -psecret -e "CREATE DATABASE $DB";
+mysql -uhomestead -psecret -e "DROP DATABASE IF EXISTS \`$DB\`";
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:+1:, or any other character, like a dash.

svpernova09:
IMHO You shouldn't use dots in database names, ___yes you can___ but you *shouldn't*. You should use underscores instead, it's much less confusing. 

If you're just dead set on using periods in database names, use the after.sh script to create them.

GrahamCampbell:
> IMHO You shouldn't use dots in database names, yes you can but you shouldn't. You should use underscores instead, it's much less confusing. 

Underscores still need escaping don't they?

svpernova09:
No, I was able to have homestead create a database named ```test_database``` without having to escape.

GrahamCampbell:
But you need to escape `test-database` for example.

svpernova09:
Right, dashes and periods must be escaped, and like I said if you're dead set on using those, have those databases created in after.sh. I'd rather have the default behavior not escape to prevent people from doing things that would likely confuse them down the road but still give experienced people the ability to do them if desired.

PR #165, opened 2015-02-26
Support for multiple aliases under one site

brandonroberts:
These changes will allow multiple subdomains to be stored under one site. Each alias will be added as a server name in the nginx configuration.


taylorotwell:
Don't want to complicate things.

PR #167, opened 2015-03-01
Use HOMESTEAD_HOME to define configuration location

nickknissen:
Use HOMESTEAD_HOME enviroment variable instead of HOME so it's possible to move homestead configuration to another location.

### DIFF
--- a/homestead
+++ b/homestead
@@ -15,9 +15,9 @@ else
 
 function homestead_path()
 {
-	if (isset($_SERVER['HOMESTEAD_HOME']))
+	if (isset($_ENV['HOMESTEAD_HOME']))
 	{
-		return $_SERVER['HOMESTEAD_HOME'].'/.homestead';
+		return $_ENV['HOMESTEAD_HOME'].'/.homestead';
 	}
 	else
 	{
--- a/homestead
+++ b/homestead
@@ -1,7 +1,7 @@
 #!/usr/bin/env php
 <?php
 
-$_ENV['HOME'] = getenv('HOME');
+$_ENV['HOMESTEAD_HOME'] = getenv('HOMESTEAD_HOME') ? getenv('HOMESTEAD_HOME') : getenv('HOME');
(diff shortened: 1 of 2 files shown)


homestead
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env php
 <?php
 
-$_ENV['HOME'] = getenv('HOME');
+$_ENV['HOMESTEAD_HOME'] = (getenv('HOMESTEAD_HOME') ? getenv('HOMESTEAD_HOME') : getenv('HOME'));
```
taylorotwell: No need to wrap this in parens.

nickknissen: I removed then parens

taylorotwell:
I'm curious why the variable is put in $_ENV but then referenced out of $_SERVER?

nickknissen:
I more or less just search and replaced. But it properly wold be best to use $_ENV since it's out of request context. I will fix this

homestead
```diff
@@ -15,13 +15,13 @@ else
...
 {
-	if (isset($_SERVER['HOME']))
+	if (isset($_ENV['HOMESTEAD_HOME']))
 	{
-		return $_SERVER['HOME'].'/.homestead';
+		return $_ENV['HOMESTEAD_HOME'].'/.homestead';
 	}
 	else
 	{
-		return $_SERVER['HOMEDRIVE'].$_SERVER['HOMEPATH'].DIRECTORY_SEPARATOR.'.homestead';
```
taylorotwell: These should probably stay $_SERVER since that is what they were before.

nickknissen: But wouldn't `$_ENV` be better since it's a non-server script?

82rules:
:+1:

ThomasBS:
:+1:

homestead
taylorotwell: If it ain't broke don't fix it.

nickknissen: Sure thing :+1:

taylorotwell:
This breaks running "vagrant box update" from the cloned Homestead repository if you are not using the PHP "homestead" helper.

PR #173, opened 2015-03-12
Added local Homestead support for homestead

ThatTomPerson:
homestead init [--local|-l]


taylorotwell:
I think things are OK as is.

ThatTomPerson:
I agree
this just adds support for having all configuration inside of the project to help people who work in teams

there seem to be people who want this and there are forks of homestead to achieve this

just though some people would beneficent if some of these changes were made upstream

if you consider a feature like this I would work on cs to get it a lot cleaner

PR #175, opened 2015-03-14
Updated Dependencies

GrahamCampbell:


taylorotwell:
Why no stable?

taylorotwell:
Ah I guess that's the default :)

GrahamCampbell:
> Ah I guess that's the default

Yep. :)

PR #176, opened 2015-03-18
Add Blackfire client id & client token to config file (Homestead.yaml)

kevinsimard:


taylorotwell:
Why do you need to put the client ID and token here?

kevinsimard:
I used "blackfire config" command to set my client ID & token on the server as they said in their doc, but whenever I run "homestead provision" the  "/home/vagrant/.blackfire.ini" file get overwritten and set empty values. The server ID & token are the only parameters that are passed to the "blackfire.sh" script from the "homestead.rb" file. Adding the client ID & token to the "Homestead.yaml" would add those values to the ".blackfire.ini" file.

```php
if settings.has_key?("blackfire")
  config.vm.provision "shell" do |s|
    s.path = "./scripts/blackfire.sh"
    s.args = [settings["blackfire"][0]["id"], settings["blackfire"][0]["token"]]
  end
end
```
to
```php
if settings.has_key?("blackfire")
  config.vm.provision "shell" do |s|
    s.path = "./scripts/blackfire.sh"
    s.args = [
      settings["blackfire"][0]["sid"],
      settings["blackfire"][0]["stoken"],
      settings["blackfire"][0]["cid"],
      settings["blackfire"][0]["ctoken"]
    ]
  end
end
```

Let me know if you have any additional questions

taylorotwell:
I don't like sid and cid. Can we just name them server-id and client-id

taylorotwell:
Where is this in the Blackfire documentation? I can't find it.

kevinsimard:
That makes sense, ill change the variable names to server-id & client-id.

--
It's in their getting started [documentation](https://blackfire.io/getting-started).

```sh
# Install the blackfire-agent package:
sudo apt-get install blackfire-agent

# Run the config command to initialize the client
# This command will ask for your Blackfire client credentials
blackfire config
```

taylorotwell:
Actually let's just leave the server IDs as id and token otherwise it will be a breaking change, then use client-id and client-token.

kevinsimard:
Done

PR #177, opened 2015-03-20
Only create the database if it doesn't already exist

deringer:
This prevents trashing existing databases when running homestead --provision

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -2,5 +2,4 @@
 
 DB=$1;
 
-mysql -uhomestead -psecret -e "DROP DATABASE IF EXISTS \`$DB\`";
-mysql -uhomestead -psecret -e "CREATE DATABASE \`$DB\` DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_unicode_ci";
+mysql -uhomestead -psecret -e "CREATE DATABASE IF NOT EXISTS \`$DB\` DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_unicode_ci";


deringer:
Just proposing this change, not sure if there was a specific reason why an existing database would be dropped if it already exists.

Given that the docs suggest being able to run `homestead provision` to add an additional site, but don't mention that all existing databases will be removed, it's a bit annoying when you do that only to find you've lost any data you already had when all you wanted to do was add a new site.

Relevant documentation:

> Once your Homestead environment is provisioned and running, you may want to add additional Nginx sites for your Laravel applications. You can run as many Laravel installations as you wish on a single Homestead environment. There are two ways to do this: First, you may simply add the sites to your `Homestead.yaml` file and then run `homestead provision` or `vagrant provision`.

andrewhood125:
#142 :8ball:

deringer:
Alright cool - thanks for that (should have searched first!)

So in that case, should there be a switch to optionally run this *or*
should the docs be updated?

andrewhood125:
@deringer no worries. I think the consensus is you should assume everything will be returned to the starting point when you provision. However we all get in situations from time to time where we have manually entered that sucks to lose. 

I think having `--preserve-databases` flag would be a good compromise, that would retain the default behavior but I'll defer to others thoughts.

vml-pthurmond:
It find that consensus immensely frustrating. That is not something you would assume in any server environment. I feel like this really harms the flexibility and usability of homestead as a dev environment for multiple projects at a time. That said, the "--preserve-databases" flag looks like an excellent idea as a compromise. Though I would think that it would make much more sense to preserve as the default behavior and have a flag to recreate all databases.

I guess it is all personal preference based on how we each work. I will look to see if that flag has been offered up. If not then I will try to figure out how to add it myself and create a pull request.

PR #179, opened 2015-04-03
Change upload max size in PHP

markvaneijk:
The nginx configuration is now set to maximum upload 100 mb files, but PHP config is still on 2 mb. So the php.ini needs to be configurated too. With one single sed command this is possible.


andrewhood125:
I think this is more appropriate for https://github.com/laravel/settler

markvaneijk:
Ah, I did not knew the existence of settler. I've also made a pull request on that package.

andrewhood125:
:100:

PR #180, opened 2015-04-04
Improve file watching for NFS shared folders.

franzliedke:
This should improve watching for file changes inside the VM (e.g. with Grunt etc.).

For reference: http://stackoverflow.com/questions/27035702/grunt-watch-detects-file-changes-only-after-5-seconds-with-vagrant-and-nfs

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -64,7 +64,8 @@ class Homestead
 
     # Register All Of The Configured Shared Folders
     settings["folders"].each do |folder|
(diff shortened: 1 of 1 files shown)


franzliedke:
P.S.: Maybe @fideloper can explain whether [the other options used in Vaprobash](https://github.com/fideloper/Vaprobash/blob/e374d5ac735110fecb7ba79a1ab0fe77d7cf0aac/Vagrantfile#L111) are also useful to add here?

PR #182, opened 2015-04-07
Add "homestead share" command (updated)

vicdelfant:
Improved version of an earlier PR, #147. This PR adds a `homestead share` command that uses Vagrant's built-in sharing capability for making sites externally available.

Usage is pretty straight-forward: create an account with Atlas and run `homestead share`. Doesn't require any additional sites to be added to Homestead.yaml (as opposed to the earlier version of this PR) and can be used with both the default dynamic hostnames generated by Atlas as well as custom domains.


scripts/stop-serve.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+if [ -f "/etc/nginx/sites-available/$1" ]; then
+	rm "/etc/nginx/sites-enabled/$1"
+	rm "/etc/nginx/sites-available/$1"
+
+	service nginx restart
+	service php5-fpm restart
+	service hhvm restart
+fi
```
GrahamCampbell: missing newline

vicdelfant: Before the `then` you mean? Added it in 0dedb02

taylorotwell:
I hate to say it but this is just so much to maintain I would rather avoid it.

vicdelfant:
No prob, decision is up to you of course.

PR #183, opened 2015-04-07
Add artisan and migrate aliases

Stichoza:
Added `art` (`php artisan`) and `migrate` (`php artisan migrate`) aliases.

I know lots of laravel guys are using `art` alias (including @JeffreyWay).

Also, `php artisan migrate` is the most frequently used artisan command on Homestead, so why not simplify things.

### DIFF
--- a/src/stubs/aliases
+++ b/src/stubs/aliases
@@ -4,6 +4,9 @@ alias ...="cd ../.."
 alias h='cd ~'
 alias c='clear'
 
+alias art="php artisan"
+alias migrate="php artisan migrate"
+
 function serve() {
 	if [[ "$1" && "$2" ]]
 	then


taylorotwell:
These can be added to your own aliases file.

Stichoza:
Of course, I know. But I thought it would be nice if Homestead provided that convenience out of the box. What about `art` alias only? (with no `migrate` alias)

Or maybe rename `art` to `artisan`?

^_^

sandrodz:
:+1:  I think `php` in front of artisan command is redundant. Why do we waste 4 keystrokes for nothing?

andrewhood125:
I personally just use `./artisan`.
 Different people are going to have different ways of aliasing. This is easily done in `after.sh` 

#### after.sh
```bash
echo 'alias art="php artisan"' >> ~/.bashrc
echo 'alias migrate="php artisan migrate"' >> ~/.bashrc
```

PR #187, opened 2015-04-11
Added SSL support

JamShady:
Added SSL support with automatically generated certificates per host during the build process

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -75,11 +75,11 @@ class Homestead
     settings["sites"].each do |site|
       config.vm.provision "shell" do |s|
           if (site.has_key?("hhvm") && site["hhvm"])
-            s.inline = "bash /vagrant/scripts/serve-hhvm.sh $1 \"$2\" $3"
-            s.args = [site["map"], site["to"], site["port"] ||= "80"]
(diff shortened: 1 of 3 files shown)


taylorotwell:
So in usage it would look like `ssl: true` in your site declaration?

taylorotwell:
When I test this I'm getting the following Nginx error:

```
400 Bad Request

The plain HTTP request was sent to HTTPS port
```

taylorotwell:
Nevermind. Helps if you put HTTPS in your browser address bar. :smile: :heart:

gpspake:
If I add `ssl: "true"` to Homestead.yaml like:
```
sites:
    - map: laravel.uni.edu
      to: "/home/vagrant/uni/public"
      ssl: "true"
```
and run `vagrant up`, the process completes but I can't get to the site in my browser (with or without https). Simply removing that line and re-running `vagrant up` works fine and I can get to my app. 

I should note that I had to add quotes to "true" because using `ssl: true`, like @taylorotwell mentioned above, returns the error `Shell provisioner 'args' must be a string or array.` on `vagrant up` - this might be part of my problem.

Is there something obvious I'm missing here?

GrahamCampbell:
Please don't comment on closed pull requests with issues.

gpspake:
@GrahamCampbell sorry. Issues seem to be disabled on this repo :/

PR #188, opened 2015-04-11
Allow default ports to be configured via Homestead.yaml settings

JamShady:
The default forwarded ports can be overridden in the Homestead.yaml settings, ala:
````
forwards:
    http: 80
    https: 443
    mysql: 3306
````

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -34,10 +34,10 @@ class Homestead
     end
 
     # Configure Port Forwarding To The Box
-    config.vm.network "forwarded_port", guest: 80, host: 8000
-    config.vm.network "forwarded_port", guest: 443, host: 44300
-    config.vm.network "forwarded_port", guest: 3306, host: 33060
-    config.vm.network "forwarded_port", guest: 5432, host: 54320
+    config.vm.network "forwarded_port", guest: 80,   host: settings.has_key?("forwards") && settings["forwards"].has_key?("http")     ? settings["forwards"]["http"]     : 8000
+    config.vm.network "forwarded_port", guest: 443,  host: settings.has_key?("forwards") && settings["forwards"].has_key?("https")    ? settings["forwards"]["https"]    : 44300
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by taylorotwell** [comment]
My main question here is should we just allow "ports" in general to be configured, and not lock people down to these specific ports. So, basically, allow a ports "to" and "from" setting.

**Comment by JamShady** [comment]
I'd say yes. The current implementation:
 * locks everyone down with inconfigurable port-forwarding
 * locks everyone to use the _same_ ports (i.e. 8000 for http)

I don't see any reason to not open it up, especially if the defaults are set so it won't break anything anyway if someone didn't set them.

A list of to/from would be more flexible than my implementation above. I guess the question is how do you ensure a default for the main ports?

**Comment by andrewhood125** [comment]
```YAML
ports:
  mysql: 
    from: 3600
    to: 36000
  nginx:
    from: 8000
    to: 80
...
````

Maybe?

**Comment by andrewhood125** [comment]
Then you could check if mysql exists and if not then use the default?

**Comment by andrewhood125** [comment]
Wait nvm, I didn't read very carefully. 

If we are going to do port forward in general and not tie it to particular services. We could just check to see if any of them map to the known default listening port for homestead. If there is mapping provided use if not use the default. 

Example:
```YAML
ports:
  from: 3600
    - 35999
```

Then we see they are overriding the default for mysql and we use the provided 35999, but they don't override nginx and so you'd roll with the default.

Thoughts?

JamShady:
I'm inclined toward the position of ignoring named ports, and just process what the user provides. Leave it to the user to dictate the port mapping in their conf.

Of course, that would be a breaking change...

dbpolito:
Following the naming pattern from previous params (https://github.com/laravel/homestead/blob/master/src/stubs/Homestead.yaml#L12), i would suggest:

````yml
ports:
    - map: 80
      to: 8080
    - map: 3360
      to: 33600
````

taylorotwell:
I was thinking basically like the example @dbpolito provides. But, yes, I'm not sure how to do that without a breaking change. I'm going to close this for now as I think the PR would take a fairly different direction. But, it's something I'm interested in if we can figure out a way to do it without breaking backwards compatibility. Perhaps if no ports option is provided at all we can setup the default existing Homestead ports?

PR #192, opened 2015-04-18
Allow overriding default Homestead ports

andrewhood125:
PR #188 discussed having a syntax like this available in `Homestead.yaml` to override the default ports. 
```YAML
ports:
  - map: 80
    to: 8000
``` 

Homestead already has support for the following structure: 
```YAML
ports:
  - guest: 80
    host: 8000
``` 

This PR allows the default ports to be overridden. If they are not overridden then the default port forwarding takes place preserving backwards compatibility.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -37,10 +37,29 @@ class Homestead
     end
 
     # Configure Port Forwarding To The Box
-    config.vm.network "forwarded_port", guest: 80, host: 8000
-    config.vm.network "forwarded_port", guest: 443, host: 44300
-    config.vm.network "forwarded_port", guest: 3306, host: 33060
-    config.vm.network "forwarded_port", guest: 5432, host: 54320
+    # Is HTTP Default Port Overridden?
+    unless settings["ports"].any? { |mapping| mapping["guest"] == 80 }
+      # Set Default HTTP Port Forwarding
+      config.vm.network "forwarded_port", guest: 80, host: 8000
+    end
+
+    # Is HTTPS Port Overridden?
+    unless settings["ports"].any? { |mapping| mapping["guest"] == 443 }
+      # Set Default HTTPS Port Forwarding
+      config.vm.network "forwarded_port", guest: 443, host: 44300
+    end
+
+    # Is MySQL Port Overridden?
+    unless settings["ports"].any? { |mapping| mapping["guest"] == 3306 }
+      # Set Default MySQL Port Forwarding
+      config.vm.network "forwarded_port", guest: 3306, host: 33060
+    end
+
+    # Is PostgreSQL Port Overridden?
+    unless settings["ports"].any? { |mapping| mapping["guest"] == 5432 }
+      # Set Default PostgreSQL Port Forwarding
+      config.vm.network "forwarded_port", guest: 5432, host: 54320
+    end
 
     # Add Custom Ports From Configuration
     if settings.has_key?("ports")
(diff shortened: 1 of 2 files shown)


brunowego:
+1

andrewhood125:
@taylorotwell @GrahamCampbell I know you're busy but any Homestead love <3?

scripts/homestead.rb
```diff
@@ -36,16 +36,32 @@ def Homestead.configure(config, settings)
...
+    default_ports = {
+      80   => 8000,
+      443  => 44300,
+      3306 => 33060,
+      5432 => 54320
+    }
+
+    # Use Default Port Forwarding Unless Overridden
+    default_ports.each do |guest, host|
+      unless settings["ports"].any? { |mapping| mapping["guest"] == guest }
```
dbpolito: no need to if this at each iteration, just do an if, if true, loop the array

scripts/homestead.rb
```diff
@@ -36,16 +36,32 @@ def Homestead.configure(config, settings)
...
+        config.vm.network "forwarded_port", guest: guest, host: host
+      end
+    end
 
     # Add Custom Ports From Configuration
     if settings.has_key?("ports")
       settings["ports"].each do |port|
-        config.vm.network "forwarded_port", guest: port["guest"] || port["to"], host: port["host"] || port["send"], protocol: port["protocol"] ||= "tcp"
+        config.vm.network "forwarded_port", guest: port["guest"], host: port["host"], protocol: port["protocol"]
```
dbpolito: here you're using `host` and `guest`, but at the example you're using `send` and `to`... pick one.


Also, make protocol optional, not sure if this way it is...

dbpolito:
Also, i would like to see an extra config, something like, `default_ports: true`, default to true, instead of checking if you have the port config to add these 4 ports, check this config... if it's true, you would add new ports, not overwrite, if false, you would overwrite...

I think the more usual case is adding new ports, not overwrite, this way we have both possibilities...

Let me know your thoughts...

scripts/homestead.rb
andrewhood125: I'd like to stick to just one but I'm not the maintainer. I've attempted to standardize `homestead.rb` to just `host` and `guest`. 

See Homestead docs.

scripts/homestead.rb
andrewhood125: I'm not sure I understand. Can you elaborate?

andrewhood125:
I disagree with adding `default_ports`. I feel like it's an unnecessary complication. You're still going to be adding complexity to the code checking to see if `default_ports: true` or `false`. Yet the configuration is less intuitive and more complicated. 

Use case: A user wants to change just MySQL port forwarding from `3306 => 33060` to `3306 => 3307`. 

With `default_ports` your `Homestead.yaml` would look like this: 
```YAML
default_ports: false
ports:
  - map: 3306
    to: 3307
  - map: 80
    to: 8000
  - map: 5432
    to: 54320
  - map: 443
    to: 4430
```

With this PR it would be:
```YAML
ports:
  - map: 3306
    to: 3307
```
If you want custom ports and just to change just MySQL, then you would need to manually add all the default port forwarding.

taylorotwell:
Current PR breaks if "ports" is not defined in Homestead.yaml.

http://d.pr/i/aATu/5XojvxDM

taylorotwell:
Fixed it for you.

andrewhood125:
Good catch. Thanks.

PR #196, opened 2015-05-07
GitHub OAuth-token integration

kreischweide:
As GitHub is a core component required for a successful installation of laravel on homestead, it would be nice to have a default setting to specify the GitHub OAuth-token for the default vagrant user. It's quite easy to hit the 60 API call rate limit nowadays.


taylorotwell:
Put this in your after.sh script for now. Thanks

PR #199, opened 2015-05-13
Update serve.sh

ingro:
I don't know if there is a more elegant solution, but this worked for me.

Basically I had to port a old app to Laravel and I needed to keep the old uris, and some of them ended with .php. With default homestead serve.sh the server just returned an `No input file specified` error.

So I Updated serve.sh to redirect properly to the application's index.php main script even for uri that ends with .php


taylorotwell:
Would rather not change this when I know it's working.

PR #201, opened 2015-05-17
Add license

svpernova09:
Add MIT License


GrahamCampbell:
Not formatted correctly. Needs to been the same as the one in the framework repo.

svpernova09:
@GrahamCampbell That was what I started using, then copied this version from the Docs repo. So it seems the docs repo needs to be updated as well.

GrahamCampbell:
:+1:

PR #202, opened 2015-05-23
Check PHP file exists before passing it to PHP

bbashy:
Returns 404 (which is true for this error), if it doesn't exist.

This fixes the issue of the "No input file selected" response.

### DIFF
--- a/scripts/serve-hhvm.sh
+++ b/scripts/serve-hhvm.sh
@@ -27,6 +27,9 @@ block="server {
     sendfile off;
 
     location ~ \.php$ {
+        # don't pass file to PHP unless it exists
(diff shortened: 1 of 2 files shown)


taylorotwell:
No input file specified just means the root web directory isn't setup correctly.

bbashy:
Yes but why keep it like that? It's not an error that explains what's wrong.

I see a lot of questions regarding this and with this change it will allow them to realise the path is wrong?

bbashy:
#199 is exactly the same issue. Shouldn't be ignored.

bbashy:
Forge is even prone to this exploit, fun!

PR #206, opened 2015-06-01
Add support for OCI8 to connect to Oracle DBs

mwolfindale:
Also add extra tools and set up networking (e.g. to connect to Oracle DB VM)


mwolfindale:
Maybe this shouldn't be in core...but here we go anyhow.

taylorotwell:
Not at this time. Thanks

PR #210, opened 2015-06-18
Don't make databases a required setting

WouterJ:
I didn't need any databases on my homestead machine, so I commented the complete `databases` block. This resulted in a Ruby error among the lines of `undefined method each of Nil`.

This PR makes the `databases` setting optional, another solution is to have a better error when it isn't available.


svpernova09:
:+1: Ran into this a few times as well

PR #212, opened 2015-06-19
Add homestead install command

svpernova09:
Add homestead install command to allow users to install homestead into their Laravel project directly.

The use case for this would be to completely isolate the Homestead vagrant environment per project and allow users to ship Homestead environments with their code base making it easier for contributors to get up and running.

### Usage

```
laravel new project
cd project
composer require laravel/homestead
./vendor/bin/homestead install
```

Copies ```Vagrantfile```, ```Homestead.yaml```, ```after.sh```,  and ```aliases``` from the global Homestead folder to the project folder. The other files used by homestead are loaded/used out of the homestead path so they can be easily updated from upstream. These files we copy into the project are the ones I most commonly configure per project.

The command also updates paths in ```Homestead.yaml``` for the project folder.

#### Two options can be used:

Because Virtualbox requires box names to be unique the user can set a unique box name. (this is something we may want to do regardless, possibly use the project folder as the box name by default while allowing this override as well)

```
./vendor/bin/homestead install --name=MyBoxName
```

You can specify a hostname to be used:

```
./vendor/bin/homestead install --hostname=customname
```

#### Vagrant Operations

We have now installed a copy of the Homestead configuration into our project we used native Vagrant command to interact with the box.

```
vagrant up|halt|destroy
```

### DIFF
--- a/src/stubs/LocalizedVagrantfile
+++ b/src/stubs/LocalizedVagrantfile
@@ -2,7 +2,7 @@ require 'json'
 require 'yaml'
 
 VAGRANTFILE_API_VERSION = "2"
-confDir = $confDir ||= File.expand_path("~/.composer/vendor/laravel/homestead")
+confDir = $confDir ||= File.expand_path("vendor/laravel/homestead")
 
 homesteadYamlPath = "Homestead.yaml"
 afterScriptPath = "after.sh"
--- a/src/stubs/LocalizedVagrantfile
+++ b/src/stubs/LocalizedVagrantfile
@@ -1,0 +1,23 @@
+require 'json'
+require 'yaml'
+
+VAGRANTFILE_API_VERSION = "2"
+confDir = $confDir ||= File.expand_path("~/.composer/vendor/laravel/homestead")
+
+homesteadYamlPath = "Homestead.yaml"
+afterScriptPath = "after.sh"
+aliasesPath = "aliases"
+
+require File.expand_path(confDir + '/scripts/homestead.rb')
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+	if File.exists? aliasesPath then
+		config.vm.provision "file", source: aliasesPath, destination: "~/.bash_aliases"
+	end
+
+	Homestead.configure(config, YAML::load(File.read(homesteadYamlPath)))
+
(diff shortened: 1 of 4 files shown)


src/stubs/LocalizedVagrantfile
```diff
@@ -0,0 +1,23 @@
+require 'json'
+require 'yaml'
+
+VAGRANTFILE_API_VERSION = "2"
+confDir = $confDir ||= File.expand_path("~/.composer/vendor/laravel/homestead")
```
taylorotwell: Why would we use ~/.composer here? Wouldn't we just use the project's composer/vendor/ directory?

taylorotwell: It seems like this would still require Homestead to be installed globally.

svpernova09: This PR assumes the user has homestead installed globally and wants to install into $this project folder they are in. 

If we want to build the same / similar thing into the Laravel command,  think that's viable but different scenario. We talked about going either way with it.

taylorotwell: But it seems like just by changing that line to use /vendor/laravel/homestead you could eliminate the need to have Homestead globally installed.

taylorotwell: I think we're kinda thinking about this in two different ways. I'm saying for the local version you should have "laravel/homestead" as a composer dependency of your main laravel app.

svpernova09:
Use Case: Method to install Homestead into a Laravel project to provide isolated Vagrant.

This PR implements this in the Homestead command, assumes it's globally installed and reuses as much as it can from that path.

An alternative implementation would be to put this functionality into the laravel command and have homestead as a dependency to the laravel framework. Then we could copy the minimal amount of files from /vendor/laravel/homestead to the project root. (this is what I'm doing with my HomesteadSkeleton package) This implementation would allow laravel framework to ship with isolated vagrant configs if desired

taylorotwell:
you don't need anything on the laravel command.

taylorotwell:
if you have laravel/homestead as a dependency of a laravel/laravel project, you will be able to access the bin via php vendor/bin/homestead... allowing you to use an install command to copy over some files.

PR #213, opened 2015-06-19
Fix spaces bug on homestead install

svpernova09:
If the path contains a space, vagrant would choke on mounting the shared folder.

Added quotes to escape the paths and made defaultName a property so we could use that sluggified value for our paths inside the vm

### DIFF
--- a/src/InstallCommand.php
+++ b/src/InstallCommand.php
@@ -35,13 +35,13 @@ class InstallCommand extends Command
         $this->basePath = getcwd();
         $this->projectName = basename(getcwd());
 
-        $defaultName = strtolower(trim(preg_replace('/[^A-Za-z0-9-]+/', '-', $this->projectName)));
+        $this->defaultName = strtolower(trim(preg_replace('/[^A-Za-z0-9-]+/', '-', $this->projectName)));
 
         $this
             ->setName('install')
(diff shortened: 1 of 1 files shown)


andrewhood125:
Partially addressed already: c651b5980a8a4db406fda2c2a14eac9429e1269a

svpernova09:
@andrewhood125 Despite c651b59, It's still broken without these commits. Need to wrap strings with spaces in yaml

andrewhood125:
D'oh, split diff made it hard to see. I was worried that preg_replace what getting repeated. Nothing to see here. Carry on.

taylorotwell:
This doesn't seem correct. We shouldn't be using $this->defaultName as a folder mapping because it's not always going to be an actual folder, especially if invalid characters were converted to dashes.

svpernova09:
The only path that has to be real on the host system is the Folder: Map: path. That still uses ```$this->basePath``` and for the mapping of that path to inside the vagrant, we use ```$this->defaultName```

Example:

image

Yields:

image

taylorotwell:
Ah yeah that's true heh. I'll merge when I get home

taylorotwell:
I renamed this command to `local`.

`php vendor/bin/homestead local`.

PR #216, opened 2015-06-22
Added support for Vagrant Host Manager plugin.

ikari7789:
To ease development even further, add support for the vagrant host manager plugin, which can manage the hosts file on the host machine at vagrant up/destroy.
https://github.com/smdahlen/vagrant-hostmanager

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -97,7 +97,9 @@ class Homestead
     end
 
     # Install All The Configured Nginx Sites
+    hosts = Array.new()
     settings["sites"].each do |site|
+      hosts.push(site["map"])
       config.vm.provision "shell" do |s|
           if (site.has_key?("hhvm") && site["hhvm"])
             s.path = scriptDir + "/serve-hhvm.sh"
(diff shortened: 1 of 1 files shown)


svpernova09:
IMHO we shouldn't be dependent on any Vagrant plugins. With this PR we still aren't, but it's a point where we decide to start accepting any PRs that may conditionally check for plugins or continue to not support plugins.

My 2 cents: I'm fine with this sort of plugin support as long as it's kept within reason. I wouldn't want to bloat ```homestead.rb``` to where the file becomes cumbersome.

@taylorotwell Do you think it would be worthwhile to break out into another ruby file to hold this sort of things for the users and delegate this sort of functionality? I'm thinking something similar to what was done for ```after.sh```. It would keep things neat and tidy and give users a designated place to put this stuff.

andrewhood125:
:+1: for extensibility. Having the equivalent of after.sh would be great.

taylorotwell:
I would be open to letting there be a "after" ruby hook.

ikari7789:
My original intention was never to bloat the file down with plugins, and the concept of a hook isn't actually such a bad idea, but I don't think supplying the user with an empty "after" ruby hook is very useful neither. Homestead is a vagrant built to help speed up development, and for some of those users, they may not be familiar yet with PHP let alone Ruby.

Perhaps something like a "plugins" folder with ruby scripts that would be executed would be better than a single "after" ruby hook file. You could separate out the logic for each particular plugin separately, you could enable configuration via the Homestead.yaml file, and you could even include some pre-designed plugin hooks (disabled by default).

PR #217, opened 2015-06-23
Feature after.rb

svpernova09:
As briefly mentioned in #216 this adds an after.rb stub to work in a similar way that we can use after.sh to add extra shell script functionality, after.rb easily allows users to add extra vagrant plugin functionality while still easily remaining up to date with upstream Homestead changes.

I tested this with the plugin mentioned in #216 in an existing and fresh Laravel project and it works as expected. Feedback welcome if there is a better way to handle this. Also my indents may be a bit odd, Phpstorm is confused on files without extensions.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,6 +6,7 @@ confDir = $confDir ||= File.expand_path("~/.homestead")
 
 homesteadYamlPath = confDir + "/Homestead.yaml"
 afterScriptPath = confDir + "/after.sh"
+afterRubyPath = "after.rb"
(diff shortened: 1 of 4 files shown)


taylorotwell:
Honestly the more I think about this people should just modify their Vagrantfile directly. There isn't any need for an extra file.

svpernova09:
:+1: Wasn't excited with how this came out. 

We should add some documentation in regards to "where to put vagrant plugin stuff" that may help people understand where / how to add that stuff VS submitting PRs to add it to the core Vagrantfile

PR #218, opened 2015-06-25
Update first site map to use homestead.test

timgws:
.app is a TLD that should not be used for testing.

We should use **Reserved Top Level DNS Names** ([RFC 2606](http://www.ietf.org/rfc/rfc2606.txt)) for testing.

> The fiercely competed new gTLD .app has sold to Google for a record-breaking $25 million.
( http://domainincite.com/18066-google-buys-app-for-over-25-million )

We should use one of the following TLDs for testing:

* .test
* .example
* .invalid
* .localhost

I selected .test, as it seems a little more appropriate.

### DIFF
--- a/src/stubs/Homestead.yaml
+++ b/src/stubs/Homestead.yaml
@@ -14,7 +14,7 @@ folders:
       to: /home/vagrant/Code
 
 sites:
-    - map: homestead.app
+    - map: homestead.test
       to: /home/vagrant/Code/Laravel/public
 
 databases:


TheShiftExchange:
Personally I think the best TLD is actually `.vm`. That is what I now use.

The reason is there will never be a `.vm` unless there is a new country made (because 2 letter TLDs are reserved for countries).

All other TLDs you've listed could theoretically become a valid TLD at some point.

OlivierParent:
Personally I'm in favor of `.local` since it's ignored by the system's default proxy server settings

taylorotwell:
This is a preference thing. You can change it to whatever you wish in your own config file.

PR #219, opened 2015-06-25
Remove all previously set server environment variables

pavoltanuska:
When I change some env variable specified in `~/.homestead/Homestead.yaml` and then provision the VM, the change isn't reflected on the Homestead instance because the provision script (`scripts/homestead.rb`) just appends the wanted env variable(s) to existing files (`/etc/php5/fpm/php-fpm.conf` and `/home/vagrant/.profile`).

Therefore we need to clean these files from previously set env variables prior to setting the new/current ones.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -124,6 +124,26 @@ class Homestead
         end
     end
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Just destroy the server and recreate it.

PR #222, opened 2015-07-03
explicitly set nfs version on synced folders

pedrofornaza:
I had the error "mount.nfs: access denied by server while mounting" when mounting whitout this.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -91,7 +91,7 @@ class Homestead
     # Register All Of The Configured Shared Folders
     if settings.include? 'folders'
       settings["folders"].each do |folder|
-        mount_opts = folder["type"] == "nfs" ? ['actimeo=1'] : []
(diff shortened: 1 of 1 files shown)


taylorotwell:
@joshfng does this right to you?

pedrofornaza:
If its not a common problem, we can put on Homestead.yaml.

joshfng:
@taylorotwell I've never experienced any issues with the standard nfs setup, but I'm not an expert on the subject.

taylorotwell:
Closing this for lack of other reported issues here.

pawlobanano:
@pedrofornaza In my case adding ",vers=3" just after "actimeo=1" helps too.
Without it NFS couldn't mount shared folders.
Tested on linux Fedora 22.

pedrofornaza:
@pawlobanano Thanks for reporting. We can now set the mount options on Homestead.yml using: mount_opts: ['actimeo=1,vers=3']

PR #223, opened 2015-07-03
using symfony/yaml

timmillwood:
Rather than doing a string replace on Homestead.yaml this patch opts for the more elegant use of symfony/yaml. It parses the Homestead.yaml file into an array and then saves this back to the file after modification.

### DIFF
--- a/src/MakeCommand.php
+++ b/src/MakeCommand.php
@@ -2,10 +2,8 @@
 
 namespace Laravel\Homestead;
 
-use Symfony\Component\Process\Process;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Yaml\Yaml;
@@ -35,8 +33,6 @@ class MakeCommand extends Command
 
     /**
      * Configure the command options.
-     *
-     * @return void
      */
     protected function configure()
     {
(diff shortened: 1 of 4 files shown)


src/MakeCommand.php
```diff
@@ -123,11 +120,16 @@ protected function configurePaths()
...
+        $homesteadFile = $this->getHomesteadFile();
+        // Add name to the 4th position in the array
+        $homesteadFile = array_slice($homesteadFile, 0, 3, true) +
+                    array('name' => $name) +
+                    array_slice($homesteadFile, 3, NULL, true);
+        // Save array back to Homestead.yaml as yaml
+        $this->saveHomesteadFile($homesteadFile);
     }
-
+
```
GrahamCampbell: no trailing whitespace please

src/MakeCommand.php
```diff
@@ -123,11 +120,16 @@ protected function configurePaths()
...
     protected function updateName($name)
     {
-        file_put_contents($this->basePath.'/Homestead.yaml', str_replace(
-            "cpus: 1", "cpus: 1".PHP_EOL."name: ".$name, $this->getHomesteadFile()
-        ));
+        // Get Homestead.yaml as an array
+        $homesteadFile = $this->getHomesteadFile();
+        // Add name to the 4th position in the array
+        $homesteadFile = array_slice($homesteadFile, 0, 3, true) +
+                    array('name' => $name) +
```
GrahamCampbell: Please use short array syntax for all new code.

timmillwood:
@GrahamCampbell Thanks for the review, updates in latest commit.

src/MakeCommand.php
```diff
@@ -123,9 +120,14 @@ protected function configurePaths()
...
     protected function updateName($name)
     {
-        file_put_contents($this->basePath.'/Homestead.yaml', str_replace(
-            "cpus: 1", "cpus: 1".PHP_EOL."name: ".$name, $this->getHomesteadFile()
-        ));
+        // Get Homestead.yaml as an array
+        $homesteadFile = $this->getHomesteadFile();
+        // Add name to the 4th position in the array
+        $homesteadFile = array_slice($homesteadFile, 0, 3, true) +
+                    ['name' => $name] +
```
GrahamCampbell: too much indentation

src/MakeCommand.php
```diff
@@ -136,18 +138,35 @@ protected function updateName($name)
...
     protected function updateHostName($hostname)
     {
-        file_put_contents($this->basePath.'/Homestead.yaml', str_replace(
-            "cpus: 1", "cpus: 1".PHP_EOL."hostname: ".$hostname, $this->getHomesteadFile()
-        ));
+        // Get Homestead.yaml as an array
+        $homesteadFile = $this->getHomesteadFile();
+        // Add name to the 4th position in the array
+        $homesteadFile = array_slice($homesteadFile, 0, 3, true) +
+                    ['hostname' => $hostname] +
```
GrahamCampbell: same

taylorotwell:
If things get more complicated we will consider this but since it is working now going to hold off.

timmillwood:
The main issue that initiated this is a bit of a edge case, when `homestead make` is run twice it create two "hostname" and "name" elements in the yaml file. Parsing the file as yaml and updating an array will resolve this.

Also it just seems odd that Yaml is used but not parsed as yaml. Ruby is parsing it as yaml for Vagrant, but PHP is not parsing it as Yaml for symfony/console. I'd fear this would introduce yet another markup issue.

I strongly feel this is a good improvement for the stability of the project especially when more people start doing Per Project Installation.

svpernova09:
If we don't merge this we should add some functionality to not duplicate the name or hostname and update those values instead.

I noticed in the commits many of the concats have the spaces removed and some extra spaces around parens have also been removed. This is fine if it's fitting with the code style of the rest of the project, otherwise it should be fixed to match the rest of the project. (I don't know if this stuff is covered via PSR-2)

taylorotwell:
OK I'll have to pull it down and test it.

taylorotwell:
Also it needs to be rebased apparently.

svpernova09:
The issue about duplicating names and hostnames on multiple make runs stems from the fact that we check if the yaml file exists before we copy it, when maybe we should overwrite it or we'd have to do some possibly gross things to update the variables. 

I'm fine with this if it covers the issues described & code style aligns with the rest of the package.

taylorotwell:
I definitely think "make" should probably just exit with a message if the
file exists so that it can't be run twice until the user deletes the
current file.

svpernova09:
:+1: on just exiting, force the user to delete it if they really regenerate it.

In that case we can leave this closed and do that in another PR

PR #227, opened 2015-07-10
add before.sh to clear out environment variables and sites

cmgmyr:
Currently when a user re-provisions the VM Homestead will overwrite sites if present, and just append environment variables. The problem is when the user removes a site or environment variable from the yaml file Homestead doesn't remove these from the VM.

This script will look at `/home/vagrant/.profile` and `/etc/php5/fpm/php-fpm.conf` and remove past environment files, as well as remove all sites from `/etc/nginx/sites-enabled` and `/etc/nginx/sites-available/`. Once the `homestead.rb` script runs it will continue to provision the sites and add the environment variables listed in the yaml file as intended. This script will bypass the need to `vagrant destroy` the box.

Please let me know if you have any comments/questions/concerns, or would like me to make any changes.


taylorotwell:
I fixed this.

PR #229, opened 2015-07-22
Add the posibiltiy to use JSON as config file

leonhh:
I have had some problems when using YAML on a windows machine. The ruby YAML parser on windows is not functioning good for me and I like to write my config in JSON(which is even more familiar for most web developers). To use you can just use the make command with the --json flag and a json config file will be copied to the root directory of your project.

### DIFF
--- a/src/MakeCommand.php
+++ b/src/MakeCommand.php
@@ -11,192 +11,192 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 class MakeCommand extends Command
 {
-	/**
-	 * The base path of the Laravel installation.
-	 *
-	 * @var string
-	 */
-	protected $basePath;
+    /**
+     * The base path of the Laravel installation.
+     *
+     * @var string
+     */
+    protected $basePath;
 
-	/**
-	 * The name of the project folder.
-	 *
-	 * @var string
-	 */
-	protected $projectName;
+    /**
+     * The name of the project folder.
+     *
+     * @var string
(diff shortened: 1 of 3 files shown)


w2pc:
I like this idea just because..

taylorotwell:
Better to be opinionated on these things. Less overhead to maintain.

leonhh:
@taylorotwell the reason of this PR is that the YAML parser on windows doesn't work as it does on OSX & linux. See also this http://laravel.io/forum/06-13-2014-ruby-errors-after-vagrant-up

I spend hours adjusting my homestead.yaml file before I started writing this and using the JSON parser it worked like a charm. It is more a bug than a something i am opinionated about.

svpernova09:
Would it be possible to use a command to convert the YAML file to JSON and then have a flag that is used to tell homestead to read from YAML or JSON?

w2pc:
Sometimes overhead is not always a bad thing. As noted there is sometimes
issues with yaml on windows. I don't  think anyone disputes that.

andrewhood125:
Well I think what @svpernova09 is getting at is all Homestead has to do is accept and read from a JSON file. By converting YAML to JSON you don't have to maintain two stubs. It would be minimal overhead. We could even make it look for a Homestead.yaml file first and fallback to a JSON if that's not found and that would be a very minimal change.

andrewhood125:
#230 is an example.

PR #231, opened 2015-07-23
Adding Multi-Connections Bridge Adapters

leonardoDavid:
With *public_network* on **Homestead.yaml**, Homestead creates a new *bridge* type adapter to share the Vagrant machine with a local network. 
This change was tested on Linux and OS X.


taylorotwell:
I would suggest doing this in your own Vagrantfile.

PR #230, opened 2015-07-22
Fallback to JSON

andrewhood125:

PR #232, opened 2015-07-24
Remove dependency on LocalizedVagrantfile

andrewhood125:
This goes along with #230 instead of updating LocalizedVagrantfile just update the regular Vagrantfile.


andrewhood125:
Either way, but the talk has been about reducing overhead.

taylorotwell:
We keep a separate file in case they diverge more.

PR #234, opened 2015-07-26
Add alias option for sites

nicklasmoeller:
Adds the possibility to declare `alias` to the `sites` block, to make sites respond on different domains.

[This StackOverflow question] (http://stackoverflow.com/questions/31630145/homestead-how-would-i-map-multiple-domains-to-the-same-path) asks for a way to do this, and I thought it would come in handy.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -111,10 +111,10 @@ class Homestead
       config.vm.provision "shell" do |s|
           if (site.has_key?("hhvm") && site["hhvm"])
             s.path = scriptDir + "/serve-hhvm.sh"
-            s.args = [site["map"], site["to"], site["port"] ||= "80", site["ssl"] ||= "443"]
+            s.args = [site["map"], site["to"], site["port"] ||= "80", site["ssl"] ||= "443", site["alias"] ||= ""]
           else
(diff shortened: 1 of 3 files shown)


taylorotwell:
I think you could just do *.domain.com.

LasseRafn:
@taylorotwell 

I (the question asker) wanted to map multiple different domains to same site, which for some reason did not work without this neat alias parameter :-)

vinkla:
@LasseRafn don't know if you're using OS X but have you seen [Setup Wildcard DNS on Mac OS X](https://vinkla.com/posts/setup-wildcard-dns-on-mac-osx/)? Works great with Homestead.

nicklasmoeller:
@taylorotwell I actually see a real use case for this. If I were to have several projects on my homestead box, and had a single project, that operates with different names, wildcard subdomains just doesn't do

e.g.

    sites:
      - map: block.app
        alias: square.app
        to: /home/vagrant/Code/Figures

      - map: banana.app
        to: /home/vagrant/Code/Banana

cmgmyr:
Having an alias option would be very helpful. I have a few projects that fit into this use case

w2pc:
The alias would be great :+1:

PR #235, opened 2015-07-28
Add auto correction for port collisions

ddelnano:
I often work with multiple vagrant boxes at the same time and this allows for correction of port collisions.


svpernova09:
:+1:  This is one of the things that bugs me since I'm also always bouncing between multiple Homestead boxes.

Is there any way to suggest/specify what the ports are corrected to?

ddelnano:
I believe homestead allows you to override them (Not 100% sure but I saw something about overriding when I made this change).  I'm not sure if you can tell vagrant to auto correct and specify some defaults to try before randomly picking one.

andrewhood125:
The Vagrant docs are ELI5 and not much help either: http://docs.vagrantup.com/v2/networking/forwarded_ports.html

PR #236, opened 2015-07-29
Enable ssh-config command

simensen:
I'd like to be able to get access to the SSH configuration information for the Homestead vagrant install so that I can setup direct SSH connections from my ssh client(s). This command gets the information that vagrant would normally supply about a vagrant box for SSH clients.

### DIFF
--- a/homestead
+++ b/homestead
@@ -32,6 +32,7 @@ $app->add(new Laravel\Homestead\RunCommand);
 $app->add(new Laravel\Homestead\UpCommand);
 $app->add(new Laravel\Homestead\UpdateCommand);
 $app->add(new Laravel\Homestead\SshCommand);
+$app->add(new Laravel\Homestead\SshConfigCommand);
 $app->add(new Laravel\Homestead\StatusCommand);
 $app->add(new Laravel\Homestead\SuspendCommand);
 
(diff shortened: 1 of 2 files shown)


svpernova09:
:+1: Looks good

Is this to make it easy for Putty? Curious about the workflow

simensen:
@svpernova09 Yep! Putty (Windows) or iterm2 (Mac) or whatever. `homestead ssh` is easy enough if I'm on iterm2 terminal already, but sometimes it would be nice to just be able to jump straight to the ssh in from clicking a button instead.

The big use case is that the other two members on my team are on Windows and right now they are having to use whatever Windows monstrosity passes for a terminal (probably either gitbash or cmd or powershell or something). I'd like to get them using something like putty and so I went looking for the config information (that I knew I have gotten from vagrant in the past). Saw it would be easy to create this passthrough command so I just did that real fast. :)

svpernova09:
Gotcha, that's definitely something that would make it easier on the Windows users.

simensen:
Thanks for the merge, @taylorotwell!

SAMDevelopment:
@simensen for your Windows colleagues. I finally stuck with Cmder ( http://gooseberrycreative.com/cmder/ ) with a Git bash profile. Might be worth checking out.

PR #237, opened 2015-07-29
Avoid ICANN Name Collision

hisman:
It’s trivial, but with **.app** domain being the new gTLD, I think it’s
best practice to change sample site in yaml file to something like
**.local**.

### DIFF
--- a/src/stubs/Homestead.yaml
+++ b/src/stubs/Homestead.yaml
@@ -14,7 +14,7 @@ folders:
       to: /home/vagrant/Code
 
 sites:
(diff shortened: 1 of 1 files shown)


svpernova09:
This has been brought up before: #218 It's a preference thing. Easy to change for each individual user as they desire.

I personally use .dev (another TLD colision) but I'm also overwriting my own hosts file so until someone registers "MyAwesomeLaravel.dev" i'm not worried.

hisman:
Oh, Okay I'll close this request then. Thanks.

PR #238, opened 2015-07-29
Add 'boot_timeout' option

JesusLeon:
Sometimes the programmer's `after.sh` extra provisioning file may contain heavy/long-lasting commands that could make the extra-provisioning process stop.

The PR adds the optional `config.vm.boot_timeout` Vagrant setting, so it is available for the programmer if needed.

More info here: http://docs.vagrantup.com/v2/vagrantfile/machine_settings.html

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -15,6 +15,11 @@ class Homestead
 
     # Configure A Private Network IP
     config.vm.network :private_network, ip: settings["ip"] ||= "192.168.10.10"
+    
+    # Configure A Specific Boot Timeout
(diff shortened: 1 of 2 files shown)


svpernova09:
What are you doing in after.sh that's taking that long, or is it slow internet connection waiting on something to download?

Not sure if this is something that should be added for everyone, or as needed

JesusLeon:
I agree, is an optional setting. In my case is a combination of both: slow connection and large database imports, (not to mention the installation of some additional Node packages).

taylorotwell:
Going to hold off on this.

PR #240, opened 2015-07-31
Update SSL support

KennedyTedesco:
There is no need of two nginx `server{}` for this purpose.


GrahamCampbell:
:+1:

taylorotwell:
You've tested this I assume?

KennedyTedesco:
Sure! :)

KennedyTedesco:
Side note: http://nginx.org/en/docs/http/configuring_https_servers.html#single_http_https_server

taylorotwell:
Do we generate SSL certificates for non-SSL sites?

KennedyTedesco:
Yes, we do. By default. It's a good thing, I guess.

PR #242, opened 2015-08-01
Added a hosts appender command. 

Cyruxx:
ToDo: Test on windows and other operating systems. I could only check it on OSX10

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,32 +1,32 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
     "hash": "e860295f4a8adca4374eedbb18cfba04",
     "packages": [
         {
             "name": "symfony/console",
-            "version": "v2.6.4",
-            "target-dir": "Symfony/Component/Console",
+            "version": "v2.7.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/Console.git",
-                "reference": "e44154bfe3e41e8267d7a3794cd9da9a51cfac34"
+                "reference": "8cf484449130cabfd98dcb4694ca9945802a21ed"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/Console/zipball/e44154bfe3e41e8267d7a3794cd9da9a51cfac34",
-                "reference": "e44154bfe3e41e8267d7a3794cd9da9a51cfac34",
+                "url": "https://api.github.com/repos/symfony/Console/zipball/8cf484449130cabfd98dcb4694ca9945802a21ed",
+                "reference": "8cf484449130cabfd98dcb4694ca9945802a21ed",
                 "shasum": ""
             },
             "require": {
-                "php": ">=5.3.3"
+                "php": ">=5.3.9"
             },
             "require-dev": {
                 "psr/log": "~1.0",
                 "symfony/event-dispatcher": "~2.1",
+                "symfony/phpunit-bridge": "~2.7",
                 "symfony/process": "~2.1"
             },
             "suggest": {
@@ -37,11 +37,11 @@
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "2.6-dev"
+                    "dev-master": "2.7-dev"
                 }
             },
             "autoload": {
-                "psr-0": {
+                "psr-4": {
                     "Symfony\\Component\\Console\\": ""
                 }
             },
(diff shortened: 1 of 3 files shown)


Cyruxx:
Execute as the following:

`sudo ./homestead 192.168.10.10 homestead.dev`

First argument is the IP address to map and second argument is the hostname.

src/HostCommand.php
```diff
@@ -0,0 +1,131 @@
...
+            // ToDo: Someone should test this! I don't have a windows machine.
+            return 'echo "' . $this->ip . ' ' . $this->hostName . '" >> "' . $this->hostFile . '"';
+        } elseif (strpos(strtoupper(PHP_OS), 'DARWIN') === 0
+            || strpos(strtoupper(PHP_OS), 'LINUX') === 0) {
+            return 'echo "' . $this->ip . ' ' . $this->hostName . '" >> "' . $this->hostFile . '"';
+        } else {
+            return false;
+        }
+    }
+}
```
GrahamCampbell: cs

src/HostCommand.php
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Get the command to execute.
+     *
+     * @return bool|string
+     */
+    protected function getCommand()
+    {
+        if (strpos(strtoupper(PHP_OS), 'WIN') === 0) {
+            // ToDo: Someone should test this! I don't have a windows machine.
```
GrahamCampbell: don't include todo comments please in the code

src/HostCommand.php
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Find the correct host file depending on the OS.
+     *
+     * @return string
+     */
+    protected function getHostsFile()
+    {
+        if (strpos(strtoupper(PHP_OS), 'WIN') === 0) {
+            // ToDo: Someone should test this! I don't have a windows machine.
```
GrahamCampbell: again

src/HostCommand.php
```diff
@@ -0,0 +1,131 @@
...
+        $this->hostFile = $this->getHostsFile();
+
+        if ($this->hostFile === false) {
+            $output->writeln('We could not find your host file.');
+            return null;
+        }
+
+
+        if ($this->dialog->askConfirmation($output,
+            "Mapping '" . $this->ip . "' to '" . $this->hostName . "'. Is that okay? (Y/n) ")) {
```
GrahamCampbell: cs

src/HostCommand.php
```diff
@@ -0,0 +1,131 @@
...
+            $output->writeln('Okay! Will now write to hosts file...');
+
+            $command = $this->getCommand();
+
+            if ($command === false) {
+                $output->writeln('We could not find the correct command.');
+                return null;
```
GrahamCampbell: just `return;`, also, you need a newline before this statement

src/HostCommand.php
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Execute the command.
+     *
+     * @param  \Symfony\Component\Console\Input\InputInterface  $input
+     * @param  \Symfony\Component\Console\Output\OutputInterface  $output
+     * @return void
+     */
+    public function execute(InputInterface $input, OutputInterface $output)
+    {
+        /** @var DialogHelper $dialog */
```
GrahamCampbell: no

src/HostCommand.php
```diff
@@ -0,0 +1,131 @@
+<?php
+namespace Laravel\Homestead;
+
+use Symfony\Component\Console\Helper\DialogHelper;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Process\Process;
```
GrahamCampbell: order these by length

composer.lock
```diff
@@ -1,32 +1,32 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
```
GrahamCampbell: Please revert all changes to the lock file.

Cyruxx:
Hey Graham, what do you mean with "cs"?

I'll make another commit if you're fine with this idea.

GrahamCampbell:
Code style.

> I'll make another commit if you're fine with this idea.

Not up to me.

GrahamCampbell:
Only Taylor can decide if he likes this.

Cyruxx:
Allright. Thanks! :-)

src/HostCommand.php
Cyruxx: The statements look good, I've copied this from the EditCommand.

Cyruxx:
Updated.

svpernova09:
IMHO this shouldn't be something Homestead does. There are vagrant plugins that handle this for you automatically. My preference would be to avoid having homestead every do anything that would require sudo.

PR #243, opened 2015-08-02
Add Symfony2 support on serve

vincenttouzet:
Add new script to serve symfony2 applications

On Homestead.yaml file : 
```yml

sites:
    - map: symfony.app
      to: /home/vagrant/Code/Symfony2/web
      symfony2: true # set symfony2 to true to use serve-symfony2.sh
```

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -112,6 +112,9 @@ class Homestead
           if (site.has_key?("hhvm") && site["hhvm"])
             s.path = scriptDir + "/serve-hhvm.sh"
             s.args = [site["map"], site["to"], site["port"] ||= "80", site["ssl"] ||= "443"]
+          elsif (site.has_key?("symfony2") && site["symfony2"])
+            s.path = scriptDir + "/serve-symfony2.sh"
+            s.args = [site["map"], site["to"], site["port"] ||= "80", site["ssl"] ||= "443"]
           else
             s.path = scriptDir + "/serve.sh"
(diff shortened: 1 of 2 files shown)


vincenttouzet:
Why this PR cannot be accepted ?

GrahamCampbell:
I assume because we're only interest in hosting laravel apps.

taylorotwell:
Re-opening so I can think about this more.

taylorotwell:
To me I would rather have a "type": "symfony2" or something like that so we can have "type": "wordpress", etc.

yellowmamba:
It would be great to have this new feature documented somewhere. I just got my Symfony2 app working on the newest version of Homestead, but with a bit of effort in updating to `2.1.6` and figuring out how to provision with `serve-symfony2.sh` script. Can you guys consider having a `CHANGELOG.md` for each release?

andrewhood125:
This feature is hot of the press.  I imagine these things and other recent additions will be documented in due time.

weaverryan:
Per documentation on the Symfony side, we're chatting about this over at symfony/symfony-docs#4676

Thanks for the added feature!

WouterJ:
Can we rename `symfony2` to `symfony`, given that Symfony 3 is going to be released quite soon and is 100% compatible with Symfony 2?

taylorotwell:
It actually does already allow `symfony` or `symfony2`.

WouterJ:
Ah, great. I missed the fact that you changed it to `type:`. :+1:

PR #244, opened 2015-08-06
Provide the ability to disable port forwarding for default ports.

simensen:
If the default Homestead port forwarding options conflict with someone else, the only option is to set another port mapping for that port. Which is fine if you are leveraging the port forwarding options to work with homestead. In my case I am not, so it is far more preferable to simply disable the default port forwarding options instead of having me select another port to forward from.

This patch allows you to disable the default port forwarding options on a guest-port-by-guest-port basis. Want to disable the HTTP forwarding? Disable the 80 default port forwarding. Updated the stub to make it easier to remove the 8000 & 44300 port forwarding options.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -47,6 +47,10 @@ class Homestead
       settings["ports"] = []
     end
 
+    if (!settings.has_key?("disabled_default_ports"))
+      settings["disabled_default_ports"] = []
+    end
+
     # Default Port Forwarding
     default_ports = {
       80   => 8000,
@@ -57,9 +61,9 @@ class Homestead
 
     # Use Default Port Forwarding Unless Overridden
     default_ports.each do |guest, host|
-      unless settings["ports"].any? { |mapping| mapping["guest"] == guest }
-        config.vm.network "forwarded_port", guest: guest, host: host, auto_correct: true
-      end
+      next if settings["ports"].any? { |mapping| mapping["guest"] == guest }
+      next if settings["disabled_default_ports"].include?(guest)
+      config.vm.network "forwarded_port", guest: guest, host: host, auto_correct: true
     end
 
     # Add Custom Ports From Configuration
--- a/src/stubs/Homestead.yaml
+++ b/src/stubs/Homestead.yaml
@@ -36,3 +36,7 @@ variables:
 #     - send: 7777
 #       to: 777
 #       protocol: udp
+
+# disabled_default_ports:
+#     - 80
+#     - 443


simensen:
I had to force push this branch as I had failed to include one of my changes. :) All should be good now, though.

svpernova09:
I like the idea of disabling specific port forwards.

Is this related to Homestead & Sculpin port conflicts? 

#244 was tagged in 2.1.6 and should automatically resolve port conflicts.

simensen:
@svpernova09 Yes this was for conflicts with Sculpin. Which issue had that? I think you typed the wrong tag. :)

I'm not sure how automatically resolving port conflicts works. Maybe that would be sufficient. I know as-was (I actually had a rebase conflict w/ that auto correct stuff so I wasn't seeing that yet) it didn't work very well.

I don't ever do anything with the port forwarding. I'd just as soon turn it off as hope something magically resolves it (somehow) or manually pick another port that I'm not going to use and hope it won't conflict with another service down the line.

svpernova09:
[Release v2.1.6](https://github.com/laravel/homestead/releases/tag/v2.1.6) includes commit 2b450997247be1bac6c216a23115cf5fc7aaec8f which should auto correct port collisions. I've not yet tried it and I believe vagrant just randomizes the port when it finds a conflict, the docs aren't very clear there. I would assume that would also solve the conflict, but if you're running into issues, we may need to look into it more.

I'm :+1: for this, I also never use the port forwards unless I'm showing Homestead to someone new.

simensen:
Yeah, if that is how auto correction works it would be less useful for me. Sculpin, for example, comes up on an as-needed basis instead of always running by default. Homestead, on the other hand, is always running for me. So the chances that I have Sculpin running at the time that Homestead starts up are pretty small. And as I discovered today working through this, simply reprovisioning won't change the ports, I'd actually have to `halt` and `up` Homestead in order for the auto correction stuff to kick in.

svpernova09:
Yep, Makes sense. 

Any of the changes to Ports, Folder Shares, CPU, Ram would require a halt and up to take effect. That's a Vagrant limitation as far as I know.

simensen:
@svpernova09 Yep, it was obvious after the fact. Do you know if `suspend` would work or just `halt`? I found `halt` worked and had wasted enough time that I didn't bother trying things with `suspend` too. :)

Might be worth updating the docs around these to say that changes to these particular things require a `halt` / `up` vs just a `provision`. The stuff in `Homestead.yml` contains configuration that is split between stuff that will rebuild on `provision` and stuff that wont'.

svpernova09:
I don't *think* suspend will pick up the changes because I believe that's just resuming the suspended image and not going through any of the VM boot process.

taylorotwell:
I'm honestly not a big fan of this. It adds more noise to the configuration file for something that doesn't come up very often in the history of Homestead (this being the first time).

simensen:
@taylorotwell What about an option to disable default ports entirely? It would be slightly less noisy, perhaps? Or finding some way to provide a way for advanced users to override this stuff some other way?

I'd just as soon not maintain a fork of Homestead that simply comments out the `80` and `443` lines, which is what I had done before. I did that on my first install of Homestead and had to figure out how to disable it again when I upgraded my host operating system and installed Homestead again from scratch. I figure if I hack something a second time I might as well try and find a long-term solution. :)

So if you have any suggestions for what I could do that you might like better, let me know. If you're not interested in me looking for a long-term solution for this at all, let me know. Thanks!

andrewhood125:
```yaml
ports: 
  - send: 80
    to: disabled
```

Would keep the config nice and clean, if not the most intuitive in the world.

Then when setting the default ports we could verify the `to` is an integer. Which would be good to have anyways, if it's not then skip.

simensen:
@andrewhood125 This is actually the first approach I took except that the whole "to: null" or "to: disabled" looked kinda backwards to me. If @taylorotwell liked this approach better and would consider merging it I'd happily rewrite this code to do that instead.

andrewhood125:
Yeah I hear you @simensen. 

```yaml
ports:
  - disabled: 80
```

Would be another alternative.

PR #245, opened 2015-08-14
Add PHPUnit and PHPSpec aliases

ShaunaGordon:
Alias the call to phpunit and phpspec in the vendor folder so users don't have to type the full path all the time (or add their own alias).

This current aliases do have the weakness of requiring being run at the project root, instead of anywhere in the project tree. Not sure how far down the "flexibility" rabbit hole we'd want to go with this, though.

### DIFF
--- a/src/stubs/aliases
+++ b/src/stubs/aliases
@@ -5,6 +5,9 @@ alias h='cd ~'
 alias c='clear'
 alias artisan='php artisan'
 
+alias phpspec='vendor/bin/phpspec'
+alias phpunit='vendor/bin/phpunit'
+
 function serve() {
     if [[ "$1" && "$2" ]]
     then


GrahamCampbell:
:-1:

GrahamCampbell:
It's very common for people to ONLY have global phpunit installs.

ShaunaGordon:
Then why doesn't Homestead ship with a global PHPUnit install by default?

GrahamCampbell:
Not sure. I think it should though. Ping @taylorotwell.

GrahamCampbell:
All symfony projects for example don't include phpunit in require-dev for this reason.

andrewhood125:
:+1:

piotr-cz:
This breaks usage of global installation of phpunit:

```sh
$ phpunit
-bash: vendor/bin/phpunit: No such file or directory
```

ShaunaGordon:
This PR was for Homestead, which isn't for production, to my understanding. Perhaps that's changed over the past year? Or perhaps Taylor uses the same config stuff for Forge, etc? I'm not sure. However, it would probably be best to open a separate issue and provide details of your implementation/setup to allow for proper troubleshooting.

piotr-cz:
Unfortunately Issues are disabled for Homestead repo.

To better describe my problem:

I'm not adding phpunit to composer.json, but running phpunit globally.
This alias overwrote path to global installation by path local installation which is not available.

The reason behind is that I'm deploying to production environment whole `/vendor` folder and but I don't see a point of deploying dev tools (PHPUnit, PHPSpec) there.

ShaunaGordon:
Why are you deploying the `/vendor` folder, instead of running `composer install` on your production server with the Composer lock file (as is intended)? Using the `composer install` method will install the production-necessary dependencies, without the development ones (assuming you have things like PHPUnit under the `require-dev` heading).

andrewhood125:
@ShaunaGordon I imagine they don't have cli access on this machine.
@piotr-cz All issues for homestead should be filed on the laravel/framework. Use this as an example: https://github.com/laravel/framework/issues/14815

PR #246, opened 2015-08-15
Add edit aliases option to edit command

OlivierParent:


src/EditCommand.php
```diff
@@ -16,7 +17,10 @@ class EditCommand extends Command
      */
     protected function configure()
     {
-        $this->setName('edit')->setDescription('Edit the Homestead.yaml file');
+        $this
```
w2pc: i don't $this should be on a line by itself.

src/EditCommand.php
```diff
@@ -5,6 +5,7 @@
 use Symfony\Component\Process\Process;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
```
svpernova09: Ensure your use statements are shortest to longest top to bottom. (like they were before you added InputOption

src/EditCommand.php
svpernova09: +1 Should be ```$this->setName('edit')``` then line break

OlivierParent: It's the same as in `src/MakeCommand.php` and `src/UpCommand.php`.

svpernova09: Hrm, I would defer to @GrahamCampbell or @taylorotwell input then.

src/EditCommand.php
OlivierParent: @svpernova09 Is there a reference to this rule. Couldn't find it in PSR-2. I thought it would make sense to order them alphabetically.

svpernova09: It's not part of PSR-2, it's part of Laravel's code style (which is a modified PSR-2 afaik) Other Laravel repos (like framework) [have a .php_cs file](https://github.com/laravel/framework/blob/5.1/.php_cs) that will tell phpcs how to style code.

PR #247, opened 2015-08-16
Support parallels as a provider

adeubank:
Hi, thank you so much for creating an awesome VM for hacking php stacks. You have saved me so much time.

I started using parallels on my mac and found a significant performance boost. I then found that the homestead box isn't supported on parallels.

I poked around and found the laravel/settler repo. I put together this pull request to support parallels as a homestead provider. It's working well enough for me that I thought you might want to take a look at.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,6 +11,10 @@ aliasesPath = confDir + "/aliases"
 
 require File.expand_path(File.dirname(__FILE__) + '/scripts/homestead.rb')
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
We would normally build the entire box file from the settler repo. That's the point since we don't want to run the whole provision script every time we boot up the box. However, I still prefer to keep it limited to VirtualBox and VMWare since those are the two I have access to.

PR #250, opened 2015-08-19
Allow wildcard domain names

EspadaV8:
I'm building an app that needs to work with wildcard sub-domains. In the homestead.yaml file using something like `*.espadav8.dev` doesn't work because the `*` doesn't get escaped correctly in a number of places causing issues.

Instead, it will now check for a `*` within the domain and if it finds one will create a hash and replace the `*` with it (to make a safe filename), e.g.

```
*.espadav8.dev
```
becomes
```
7c5b7a74289c8ee8976770bae86deebc07b43520.espadav8.dev
```

### DIFF
--- a/scripts/serve-hhvm.sh
+++ b/scripts/serve-hhvm.sh
@@ -1,9 +1,17 @@
 #!/usr/bin/env bash
 
+FILENAME="$1"
+
+if [[ $1 == *"*"* ]]
+then
+    HASH=`echo $1 | md5sum | cut -f1 -d" "`
+    FILENAME="${1/\*/$HASH}"
+fi
+
 mkdir /etc/nginx/ssl 2>/dev/null
-openssl genrsa -out "/etc/nginx/ssl/$1.key" 1024 2>/dev/null
-openssl req -new -key /etc/nginx/ssl/$1.key -out /etc/nginx/ssl/$1.csr -subj "/CN=$1/O=Vagrant/C=UK" 2>/dev/null
-openssl x509 -req -days 365 -in /etc/nginx/ssl/$1.csr -signkey /etc/nginx/ssl/$1.key -out /etc/nginx/ssl/$1.crt 2>/dev/null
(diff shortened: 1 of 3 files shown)


taylorotwell:
This seems like a kind of heavy handed change for something pretty simple. Should the serve blocks just always allow wildcards?

EspadaV8:
The server block does allow wildcard domains (`server_name *.espadav8.dev;`) but without this the rest of the script creates `*.espadav8.dev` files, e.g.

```
/etc/nginx/sites-enabled/*.espadav8.dev
/var/log/nginx/*.espadav8.dev-error.log
/etc/nginx/ssl/*.espadav8.dev.crt
/etc/nginx/ssl/*.espadav8.dev.key
```

With the SSL crt containing a `*` nginx won't start

```
2015/08/19 01:54:49 [emerg] 2048#0: BIO_new_file("/etc/nginx/ssl/*.espadav8.dev.crt") failed (SSL: error:02001002:system library:fopen:No such file or directory:fopen('/etc/nginx/ssl/*.espadav8.dev.crt','r') error:2006D080:BIO routines:BIO_new_file:no such file)
```

EspadaV8:
It would be easier to just replace the `*` with a predefined string (e.g. `laravel_wildcard.espadav8.dev`) instead of calculating the md5 of the domain, but I thought that could possibly (although unlikely) have a collision with a domain someone actually wanted to use.

It would then become `FILENAME="${1/\*/laravel_wildcard}"` and result in filenames like `/var/log/nginx/laravel_wildcard.espadav8.dev-error.log`.

vinkla:
@EspadaV8 have you seen [Setup Wildcard DNS on Mac OS X](https://vinkla.com/posts/setup-wildcard-dns-on-mac-osx/)?

EspadaV8:
@vinkla yeah, this isn't about resolving the DNS but about setting up the nginx server so that it responds to requests for the wildcard domain.

PR #252, opened 2015-08-20
Fix support for `vagrant global-status --prune`

nedmas:
When running the command `vagrant global-status --prune` (used by Vagrant Manager) you receive an error that `homestead.rb` can not be loaded. This appears to be because the file path is relative to where you run the vagrant command. In order to fix this issue we can pass the containing directory of the vagrant file to the `expand_path()` method when we make the default `confDir`.

### DIFF
--- a/src/stubs/LocalizedVagrantfile
+++ b/src/stubs/LocalizedVagrantfile
@@ -2,7 +2,7 @@ require 'json'
 require 'yaml'
 
 VAGRANTFILE_API_VERSION = "2"
-confDir = $confDir ||= File.expand_path("vendor/laravel/homestead")
+confDir = $confDir ||= File.expand_path("vendor/laravel/homestead", File.dirname(__FILE__))
 
 homesteadYamlPath = "Homestead.yaml"
 homesteadJsonPath = "Homestead.json"


cooperaj:
Could not find any reports of this issue in the wild at *all* and the lack of vagrant manager is killing me at this point. I'm glad someone was able to find what the cause was. Thanks :)

nedmas:
@cooperaj glad that someone else found this useful :) I'm interested to find out why you needed to add the dir command. Would you be able to post the contents of your `Vagrantfile` and the version of homestead please?

cooperaj:
I had to do this https://gist.github.com/cooperaj/790c2fd77fa33342e968 in the end. Vagrant manager now works at least though.

nedmas:
@cooperaj I've had a play with version 2.1.4 and the bigger issue is that `Homestead.configure(config, YAML::load(File.read(homesteadYamlPath)))` isn't wrapped in an if statement to check it exists. This was introduced in v2.1.6. This PR is targeted at v2.1.6 but there's no reason I can see that prevents you from updating your `Vagrantfile` and still using v2.1.4 if you need.

PR #253, opened 2015-08-22
Add default_site setting

wilsonpinto360:
This is an optional setting that will set nginx default site. It will be useful if you wanna use external dynamic dns forwarding to homestead.


taylorotwell:
I really don't want to clutter things with so many options.

PR #255, opened 2015-08-24
Add a default wildcard subdomain for all provisioned site

EspadaV8:
I think this is what you meant in this https://github.com/laravel/homestead/pull/250#issuecomment-132413771 about allowing wildcard (sub-)domains.


taylorotwell:
Going to hold off on this and prefer to just have a default way to change the generated Nginx file used by Homestead.

PR #256, opened 2015-08-26
Convention for supplementing nginx site configs

collegeman:
I'm working on a project that incorporates a large number of rewrites. I'd like to be able to version those rewrites along with the rest of the codebase, and deploy those changes in a standard way across all of my projects. So this is the concept I came up with: for Laravel-like projects in which the root of the app/site is a subfolder in the main project, a folder called `/nginx` contains the supplemental configuration files; for projects like WordPress in which the root of the site is the also the root of the project, a folder called `/.nginx` serves the same purpose. What do you think? Cheers!

### DIFF
--- a/scripts/serve.sh
+++ b/scripts/serve.sh
@@ -49,6 +49,9 @@ block="server {
 
     ssl_certificate     /etc/nginx/ssl/$1.crt;
     ssl_certificate_key /etc/nginx/ssl/$1.key;
+    
+    include \"$2/.nginx/*.conf\";
+    include \"$2/../nginx/*.conf\";
 }
 "
 


taylorotwell:
Personally I think it would be better to just have a standard way to change the default generated nginx file by Homestead rather than bake-in this specific use case.

collegeman:
I would be happy to implement that so it's available in a standard way. Can you give me some basic requirements? I don't want to waste my time building something that is dramatically misaligned with your vision for the product.

taylorotwell:
The basic idea of how I picture it working is having something like a ~/.homestead/nginx/server, ~/.homestead/nginx/symfony-server, etc... We would check for the existence of those files, and if they exist use their contents in the serve scripts.

w2pc:
I think Taylor's solution would be the right way to implement this as well.
This structure could lead to over riding several other files that might
come up.

collegeman:
Breaking from my original proposal, I still think that keeping the config file that is custom to my project versioned with my project is the solution that makes the most sense from the standpoint of easing collaboration, i.e., if I'm going to collaborate on a project, I fork the project codebase, and I make Homestead aware of my new project, from which configuration for the site, database, etc. are optionally pulled from my project's codebase instead of from a central config file. 

This would be patterned on the way the VVV project provides for [auto site setup](https://github.com/Varying-Vagrant-Vagrants/VVV/wiki/Auto-site-Setup). Their approach to adding new projects to a virtual server is a conventions-based system whereby files in the project cue the vagrant provisioning process. 

Homestead configuration, in comparison, is currently centralized: I am to use `homestead edit` to edit the YAML config file in `~/.homestead`, but then these steps must also be performed by anyone who might want to collaborate with me on my project, and so must exist in documentation instead of riding along with the project—onboarding new collaborators would be made easier by the following. 

I propose taking this modular approach we have been discussing a step further with this convention: a `.homestead/Homestead.yaml` optionally included in each project, combined with a new `projects` convention/section in the centralized configuration file, thereby allowing me to "plug in" other projects to an existing Homestead machine as simple as adding something like this to the centralized config:

    projects:
      - map: ./path/to/my/project

Then on provisioning, the primary provisioning script `homestead.rb` would walk something like the following:

    if settings.include? 'projects'
      settings["projects"].each do |project|
        # {project[map]}/.homestead/Homestead.yaml is primary config
        # load custom instructions and/or follow these conventions:
        # {project[map]}/.homestead/nginx/site.conf is the nginx site config
        # {project[map]}/.homestead/database/backups/mysql.sql is a database snapshot to load
        # ...
      end
    end

The potential collision with other files in a project is an edge case worth considering, but given the addition of a `.homestead` namespace, it's probably highly unlikely that these collisions would occur, especially in Laravel projects.

PR #257, opened 2015-08-27
Database upgrades

mikemand:
Allow developer to specify not only the database’s name, but a username and password to use as that database’s owner/default login.

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
 DB=$1;
+USER=$2;
+PASS=$3;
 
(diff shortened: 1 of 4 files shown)


mikemand:
I just tested the Postgress stuff (I think...) and it did not work.

```
create: command not found
grant: command not found
```

taylorotwell:
Prefer to stick to the conventions we have.

mikemand:
Would you mind my asking to which conventions you are referring? This repository has absolutely no information regarding contributions.

than:
Just did the same thing, wanting to avoid db noise in my local homestead environment.

PR #267, opened 2015-10-01
Add homestead file

mikebronner:
Wasn't sure if this was an intentional omission. As I use the homestead command all the time, I was at first confused why it no longer worked after updating to the php-7 branch.


taylorotwell:
Won't be included in the next version of Homestead. We prefer and document the basic `vagrant` commands instead of using PHP helpers.

PR #269, opened 2015-10-06
laravel-create shortcut

pavelrogala:
Added a shortcut to ~/.homestead/aliases for running `composer create-project laravel/laravel --prefer-dist`. Accepts any options accepted by composer create-project including path and version. 
Example use: `laravel-create Blog 5.1`


taylorotwell:
Check out the Laravel installer in the docs. :)

pavelrogala:
I use the laravel installer on my host machine but it's not included in Homestead and even if I add it it wouldn't persist through a `homestead destroy`. Either way, was worth a try :). And @taylorotwell Laravel is AWESOME!

PR #270, opened 2015-10-16
Add ability to configure SMB shares

ZackWard:
Add the ability to configure SMB shares in Homestead.yaml in the following format: 

```
folders:
    - map: ~/Laravel
      to: /home/vagrant/Laravel
      type: "smb"
      smb_username: "Windows User"
      smb_password: "Password"
```

I'm very new to this, and I don't even know Ruby. =P There is probably a much much better way to do this, but this is what worked for me. Using SMB on a windows host solves a lot of issues for me, primarily the long file path issue that is caused by Virtualbox shares not using UNC paths for the deep dark recesses of node_modules.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -112,7 +112,11 @@ class Homestead
             mount_opts = folder["mount_opts"] ? folder["mount_opts"] : ['actimeo=1']
         end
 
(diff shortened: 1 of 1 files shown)


ZackWard:
On second thought, it seems like a bad idea to stick your windows password in a config file just so that you don't have to be prompted for it in a Windows command prompt.

PR #271, opened 2015-10-19
add a laravel-installer script provisioner.

lesstif:
laravel-installer wasn't included in Homestead machine and this PR is to install it on provision by running command below .

```sh
laravel new project
```

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -191,6 +191,11 @@ class Homestead
       s.inline = "/usr/local/bin/composer self-update"
     end
 
+    # Install Laravel global installer
+    config.vm.provision "shell" do |s|
+        s.path = scriptDir + "/laravel-installer.sh"
+    end
+
     # Configure Blackfire.io
     if settings.has_key?("blackfire")
       config.vm.provision "shell" do |s|
--- a/scripts/laravel-installer.sh
+++ b/scripts/laravel-installer.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+sudo -u vagrant -i composer global require "laravel/installer=~1.1";
+
+sudo -u vagrant -i laravel list >/dev/null 2>&1 || sudo -u vagrant -i echo 'export PATH=$PATH:$HOME/.composer/vendor/bin' >> /home/vagrant/.profile;


andrewhood125:
Seems a better fit for settler.

svpernova09:
Not sure if encouraging users to create new Laravel apps from within Vagrant is a good idea or not. They could easily create something and lose it. IMHO it's better to create new instances of Laravel on the host machine and then map to that new folder via Homestead.yml file.

Also agree with @andrewhood125 this is probably a better fit for [laravel/settler](https://github.com/laravel/settler)

jonnywilliamson:
@svpernova09 - fair point but:

Take me for example. I use windows to develop but I HATE having any development software installed on my windows machine. 

This is what makes Homestead so attractive, a virtual environment that has PHP/mysql/etc all set up so I can just SSH in and get going.

However, how do I get a new Laravel app installed on my main machine so that Homestead can use it in the shared folder?

Lets look at the Laravel install docs (http://laravel.com/docs/5.0/installation)...on the very first line:

>Laravel utilizes Composer to manage its dependencies. So, before using Laravel, you will need to make sure you have Composer installed on your machine.

>Install Laravel
>Via Laravel Installer
>First, download the Laravel installer using Composer.
>`composer global require "laravel/installer=~1.1"`


Oh ok, so I need composer to install Laravel.

Let's take a look at the install requirements for composer (https://getcomposer.org/doc/00-intro.md#system-requirements)

>Composer requires PHP 5.3.2+ to run. A few sensitive php settings and compile flags are also required, but when using the installer you will be warned about any incompatibilities.



Mmm.

So now, on my clean windows machine, that has a fully running virtual machine with everything working perfectly, I now have to:
* install PHP on my windows machine.
* Then install composer. 
* Then install the laravel package globally to composer. 
* Then create a new laravel app in the vagrant share folder.
* Finish.



It seems excessive. It seems messy, running windows versions of PHP etc never seem to be as clean as the unix/mac alternatives. 

Wither its in homestead or settler, I think this is a useful addition.

svpernova09:
@jonnywilliamson 

Yeah, that's the tradeoff of using Windows. It's not a terrible process to install enough things into Windows to get composer working. 

If you don't want to install the tools in Windows, you should just download the zip file from here: http://cabinet.laravel.com/latest.zip which is the same place the installer downloads from: https://github.com/laravel/installer/blob/master/src/NewCommand.php#L97

That way you can extract that outside of Vagrant and then add that path to your homestead config.

jonnywilliamson:
>Yeah, that's the tradeoff of using Windows. It's not a terrible process to install enough things into Windows to get composer working.

The thing is, it's not just Windows. You have to have a local version of php and composer on a mac too if you want to install laravel. The difference is that it's just much slicker installing those in a mac environment.

The main point still remains:

Why do we need to install dev tools on our main machine, when Homestead, created especially to deal with development stuff, would allow us to easily bypass this requirement.

The bottom line I suppose is, this is just a convenience PR. I already do my laravel installs inside my Homestead machine (not via my main windows machine) for the reasons I outline above.

The only person who's opinions matters is of course Taylor's. But I think it only proper to explain why I believe this is a reasonable idea.

taylorotwell:
Do agree this is probably more of a laravel/settler thing.

jonnywilliamson:
@lesstif - Perhaps doing a PR on the https://github.com/laravel/settler repo will be more successful! 

:)

lesstif:
@jonnywilliamson  i was created new PR on https://github.com/laravel/settler/pull/54. thanks :)

PR #273, opened 2015-10-20
Hide output of terminal and remove parameter null

anteriovieira:
Hide output of terminal when editing Homestead.yaml


taylorotwell:
Hmm, I think it's OK.

anteriovieira:
I made a mess here, It has something that can be taken advantage of?

PR #274, opened 2015-10-22
Migrate and seed aliases

martinbean:
Add helpful `migrate` and `seed` aliases.

### DIFF
--- a/src/stubs/aliases
+++ b/src/stubs/aliases
@@ -4,6 +4,8 @@ alias ...="cd ../.."
 alias h='cd ~'
 alias c='clear'
 alias artisan='php artisan'
+alias migrate='artisan migrate'
+alias seed='artisan db:seed'
 
 alias phpspec='vendor/bin/phpspec'
 alias phpunit='vendor/bin/phpunit'


taylorotwell:
I agree that this is helpful but just don't want to open the door to a lot more alias PRs. :smile:

svpernova09:
@martinbean 

fwiw, if you're using homestead per project you can easily add all the aliases you want by using the aliases flag via the make command.

martinbean:
@taylorotwell Fair enough. Understandable! :smile: 

@svpernova09 Already aware, thanks. I just find myself adding the same aliases for each project, so thought I’d try and get them merged.

jonnywilliamson:
@svpernova09 Could you explain that a little more please? I don't quiet follow what you mean.

Thank you.

svpernova09:
@jonnywilliamson 

Have a look at the docs about [per project install](http://laravel.com/docs/5.1/homestead#per-project-installation) When you run `php vendor/bin/homestead` add `--aliases` and it will create an aliases file in the folder you can add all the aliases you want there.

jonnywilliamson:
@svpernova09 Ah ok!

I haven't had to do per project installs, so I hadn't noticed that option. Thank you.

PR #275, opened 2015-10-22
Command self update added

anteriovieira:
Updating homestead:

```
homestead self-update
```


taylorotwell:
We're sort of shying away from these Homestead commands overall because they are a little more magical and it's not clear to many people how they are actually working.

PR #276, opened 2015-10-25
Update aliases

scottpnelson:
I was encountering the following error whenever I updated my ~/.bash_aliases file and ran "source ~/.bash_aliases".

/home/vagrant/.bash_aliases:8: parse error near `&&'

This prevented the serve command from working. Changing how the arguments get checked fixed the problem.


scottpnelson:
Oops! Turns out I was trying to add the serve function using the zsh shell, not bash! All good now. Closed this PR.

PR #277, opened 2015-10-29
Add a vagrant plugin to update the hosts when booting the VM.

vtalbot:
By adding the plugin `vagrant-hostsupdater`, if the user has set a `hostname` to something like `laravel.app`, it would be automatically added to the `/etc/hosts` file.

### DIFF
--- a/src/stubs/LocalizedVagrantfile
+++ b/src/stubs/LocalizedVagrantfile
@@ -8,10 +8,15 @@ homesteadYamlPath = "Homestead.yaml"
 homesteadJsonPath = "Homestead.json"
 afterScriptPath = "after.sh"
 aliasesPath = "aliases"
+vagrantPlugins = %w(vagrant-hostsupdater)
 
 require File.expand_path(confDir + '/scripts/homestead.rb')
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+    vagrantPlugins.each do |plugin|
+        system "vagrant plugin install #{plugin}" unless Vagrant.has_plugin? plugin
+    end
+
     if File.exists? aliasesPath then
         config.vm.provision "file", source: aliasesPath, destination: "~/.bash_aliases"
     end


andrewhood125:
https://github.com/laravel/homestead/pull/91

taylorotwell:
Going to hold off on this.

QWp6t:
@taylorotwell This would be a fantastic addition to Homestead. I hope you'll reconsider it.

jonnywilliamson:
What about me on a windows development machine? Why would you only add a plugin that helps MacOs/unix systems?

QWp6t:
The plugin works on Windows...

jonnywilliamson:
I would tend not to agree.

https://github.com/cogitatio/vagrant-hostsupdater/issues/40

QWp6t:
Well, you're wrong. It works on Windows. I develop on Windows. I use the plugin, literally every day, on Windows. It works fine. 

Here are a few options that you have...
* Set `vagrant.exe` to open as administrator. 
* Turn off UAC. 
* Add `trustInfo` element to `vagrant.exe` application manifest resource. It looks like this...
```xml
<trustInfo xmlns="urn:schemas-microsoft-com:asm.v3"><security>
<requestedPrivileges><requestedExecutionLevel level="requireAdministrator" uiAccess="false"></requestedExecutionLevel></requestedPrivileges>
</security></trustInfo>
```
Any of these methods will work, and there are other more advanced methods that will work as well.

But this isn't a technical support ticket for little old Johnny Williamson. This is a PR for Homestead. And if the Vagrant plugin in question didn't work on Windows (which it does), I still don't see why a lack of Windows support for an *optional plugin* would preclude it from being added. Your objection is quite frankly absurd.

jonnywilliamson:
Hehe. I enjoyed that @QWp6t 

So basically what you've said is, it works on windows. IF

1) You give a relatively random *plugin* administrator rights on your computer OR
2) You disable the security layer of the computer OR
3) You have to manually edit an application manifest resource.

Sounds brilliant. Wait no, the other thing...it sounds like a very un-Laravel like support nightmare.

I think Taylor has the right idea here.

PR #278, opened 2015-10-31
Adding message to the end of script

simonardejr:


taylorotwell:
That might not be exactly correct depending on what port they are using.

PR #281, opened 2015-11-05
Fixed access denied error on app_dev.php/

OlivierParent:


rndstr:
:+1: this is also how the [nginx recipe](https://www.nginx.com/resources/wiki/start/topics/recipes/symfony/) does it (for sf2)

PR #282, opened 2015-11-06
Use 2048 for certificates

bbashy:
Even though it’s local dev, should use up-to-date bit-length.
https://en.wikipedia.org/wiki/Key_size#Asymmetric_algorithm_key_lengths


GrahamCampbell:
:+1:

PR #283, opened 2015-11-11
Add reload command with --provision option

lucasmichot:
* Add `reload` command with `--provision` option
* Also sort commands


taylorotwell:
We're deprecating all these commands. I don't want to rebuild the Vagrant CLI in PHP.

PR #287, opened 2015-11-17
Updated parameters for folder sharing

anteriovieira:
```
mount_options (array) - A list of additional mount options to pass to the mount command.
```


taylorotwell:
Wouldn't this be a breaking change? Did we just have it wrong to begin with?

PR #289, opened 2015-11-25
Added "down" alias for "halt" command

martinssipenko:
I often forget the `halt` command and TBH find it strange that it's not `down` by default. In any case I think this won't harm.


svpernova09:
This looks solid.

I believe the plan is to depreciate these commands since they're just wrappers around vagrant commands.

taylorotwell:
Deprecating soon.

PR #296, opened 2015-12-10
MakeCommand now includes various enhancements.

AdenFraser:
Includes:
 - Argument to define the IP address for the box to use.
 - Argument to determine if a Homestead.yaml.example should be created.
 - When `php vendor/bin/homestead make` is run, it will check for a
   Homestead.yaml.example file in the projects basePath. If it exists
   it will update the folder paths for that, rather than use the default
   stub file.

### DIFF
--- a/src/MakeCommand.php
+++ b/src/MakeCommand.php
@@ -44,10 +44,12 @@ class MakeCommand extends Command
         $this
             ->setName('make')
             ->setDescription('Install Homestead into the current project')
-            ->addOption('name', null, InputOption::VALUE_OPTIONAL, 'The name the virtual machine.', $this->defaultName)
-            ->addOption('hostname', null, InputOption::VALUE_OPTIONAL, 'The hostname the virtual machine.', $this->defaultName)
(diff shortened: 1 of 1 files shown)


AdenFraser:
Let me know your thoughts. My only consideration at the moment is that perhaps the execute method is now a little bloated.

svpernova09:
What is the use case for specifying the IP? Are you using the Homestead.yaml.example file like you would the .env.example? - a placeholder for options?



-- 
- Joe Ferguson
JoeFerguson.me
MemphisPHP.org

AdenFraser:
If you use Homestead in a collaborative environment on a per project basis, you probably define a different IP address for each different project and box. 

With this addition you can commit the Homestead.yaml.example to the your repository / version control with the defined project values (Hostname, name, ip address) and subsequent developers who `php vendor/bin/homestead make` will receive an identical environment to all other developers on the project.

Thanks to the way this is setup, it means you could configure custom ports, configure blackfire and define extra memory and cpu in your Homestead.yaml.example and all developers you provision later will receive the same setup and allocation.

svpernova09:
@AdenFraser 

:+1: That's exactly what I was hoping for. I've had this same issue before regarding tweaking ports, even hosts going to the VM.

PR #297, opened 2015-12-10
Remove cron job if schedule=false

peterpan666:
At this time you can't disable schedule unless you remove manually the file in /etc/cron.d

Now we can simply specify that schedule is false so homestead provision will handle that for us.


peterpan666:
Nice thanks! When do you think this feature will be added to the new version?

PR #298, opened 2015-12-11
Add support for symfony ~3.0 on Homestead 2.1.x

crynobone:
Branch `php-7` does support it at the moment and hoping that the current stable release can support it as well.


vinkla:
:+1:

PR #299, opened 2015-12-15
Set the laravel/homestead box as default.

MatthewLilley:
In some terminals, the default box is not recognised. Setting the default box to "laravel/homestead" fixes this issue. 

I noticed this when using http://babun.github.io/ on windows.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,4 +25,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     if File.exists? afterScriptPath then
         config.vm.provision "shell", path: afterScriptPath
     end
+    
+    config.vm.box = "laravel/homestead"
+    
 end


jonnywilliamson:
Just out of interest - as I don't know the answer myself, but would this have any interference if someone was trying to use a PHP-7 Box.



In the [instructions](http://laravel.com/docs/5.1/homestead#first-steps) it says:
image


So wondering if having it hard coded would stop someone doing that?

taylorotwell:
Indeed can be set in the YAML file.

MatthewLilley:
Ah, I didn't notice the box option as it was in the upgrading to php 7 part of the documentation.

Would you agree that a default box should be set? Using bash when I vagrant up, it defaults to laravel/homestead, but zsh needs the box to be set.

Maybe the default box should be set in the copied yaml file, to laravel/homestead. And then in the upgrade guide, instead of adding the box option, simply change it to laravel/homestead-7 instead.

PR #303, opened 2015-12-22
Add support for linked clone on VirtualBox

ngocphamm:
Requires Vagrant 1.8 and later. Do have check to see if user has Vagrant 1.8 to include the config, so it won't break people's boxes.


taylorotwell:
Going to hold off on this for a bit.

PR #308, opened 2015-12-28
Nginx remove trailing slash

pYr0x:
homestead is using nginx as webserver. so the .htaccess in the public folder has no impact.

i added 
```
if (!-d $request_filename) {
rewrite ^/(.*)/$ /$1 permanent;
}
```
to all the bash script files. hopefully the right one


pYr0x:
@taylorotwell why no merge?

PR #315, opened 2016-01-23
Xdebug and tweak for Symfony2

sanis:
Hello,

I was looking for PHP7 ready vagrant box for my Symfony projects. After quite a long search and a lot of tries I came up with using Laravel homestead which is great and has great setup for project start. But it lacked of debugging tool like xdebug. So I decided to make a small script to integrate optional xdebug to homestead. 

And now you can just add ```xdebug: true``` to the homestead yaml file and xdebug will be installed automatically.

Maybe implementation is not in the best way, but maybe after some suggestions it can be improved.

Also I have added serve-symfony2 function to serve symfony projects easily.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -219,5 +219,11 @@ class Homestead
         ]
       end
     end
+
+    if settings.has_key?("xdebug")
+    	config.vm.provision "shell" do |s|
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Please send two separate PRs with exactly one commit in each.

PR #316, opened 2016-01-23
Function to serve Symfony projects easily.

sanis:
Just an fuction in aliases added to serve Symfony projects easily.


src/stubs/aliases
```diff
@@ -32,3 +32,15 @@ function serve-hhvm() {
         echo "  serve-hhvm domain path"
     fi
 }
+
+function serve-symfony2() {
```
WouterJ: please use `symfony` instead of `symfony2`

DonPrus:
We really need Symfony 2 & 3 support at some projects.
Every time I just create a copy of app_dev.php and name it as index.php :)

scripts/homestead.rb
```diff
@@ -141,7 +141,7 @@ def Homestead.configure(config, settings)
       end
 
       if (type == "symfony")
```
WouterJ: you should change this to `symfony2`, to keep backwards compatibility with people using `symfony2` as type

WouterJ:
@DonPrus there already is Symfony 2/3 support in homestead. See the (not yet merged) documentation here: http://pr-5655-6qmocelev2lwe.eu.platform.sh/cookbook/workflow/homestead.html

PR #317, opened 2016-01-23
Xdebug for PHP7

sanis:
Hello,

I was looking for PHP7 ready vagrant box for my Symfony projects. After quite a long search and a lot of tries I came up with using Laravel homestead which is great and has great setup for project start. But it lacked of debugging tool like xdebug. So I decided to make a small script to integrate optional xdebug to homestead.

And now you can just add xdebug: true to the homestead yaml file and xdebug will be installed automatically.

Maybe implementation is not in the best way, but maybe after some suggestions it can be improved.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -219,5 +219,15 @@ class Homestead
         ]
       end
     end
+
+    # Setup Xdebug if it has been set for this machine
+    if settings.has_key?("xdebug")
+      config.vm.provision "shell" do |s|
+        if (settings["xdebug"])
+          s.path = scriptDir + "/xdebug.sh"
+        end
+      end
+    end
+
   end
 end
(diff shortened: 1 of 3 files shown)


taylorotwell:
It will be installed on the next settler release. I will build it today.

baseten:
Has XDebug been reintroduced to homestead? Doesn't appear to be in the latest release, even though PHP 7 support seems solid now?

jonnywilliamson:
@baseten 

Yes it's installed

https://github.com/laravel/settler/blob/e153e338936db471b5e06d943ab5028e909e730b/scripts/provision.sh#L55

And options were added to config files
https://github.com/laravel/settler/commit/e153e338936db471b5e06d943ab5028e909e730b

But it is deactivated for the COMMAND LINE only as it slows composer massively.
https://github.com/laravel/settler/commit/001653413046660f96451d4a73f59a313eaea49f

baseten:
Ah OK, I guess IntelliJ reports that XDebug is not installed because it's looking via CLI. Once I'd updated the XDebug config everything works as expected. Thanks.

jonnywilliamson:
You can also enable it easily like this when running tests from phpstorm/intellij

image

PR #319, opened 2016-02-16
Reverse Proxy Script/Alias

CodyBohn:
Added a script and relevant alias to make provisioning a reverse proxy using nginx fast/simple.

### DIFF
--- a/scripts/serve-proxy.sh
+++ b/scripts/serve-proxy.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+mkdir /etc/nginx/ssl 2>/dev/null
+
+PATH_SSL="/etc/nginx/ssl"
+PATH_KEY="${PATH_SSL}/${1}.key"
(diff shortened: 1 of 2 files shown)


taylorotwell:
What do you use this for on Homestead?

CodyBohn:
I manually wrote configs for self-serving software (e.g. node) like redis-commander. The process was very similar to the existing serve scripts so I just adapted it.

taylorotwell:
So if you set one up for redis-commander what does that then allow you to do? Sorry just trying to understand the use case.

CodyBohn:
I completely understand. It might be uncommon but it's convenient for me so I thought I would share.

In essence it allows you to treat services that expose themselves directly on a port (e.g. 8080) as another project accessible at http://redis.app rather than 192.168.10.10:8080 or homestead.app:8080.

I reference it in my after.sh for serving redis-commander. I also use it for front end projects that serve themselves and do hot reloading during development. The proxy portion is native to nginx.

I hope that does a better job of explaining it.

PR #322, opened 2016-03-11
Make the config directory compliant with the XDG specification.

borela:
The vagrant box (0.4.2, virtual box) has this problem too, at the end of the .profile file, the composer bin directory is added as follows:

PATH="/home/vagrant/.composer/vendor/bin:$PATH"

but composer is following the XDG specification too, the actual directory is:

/home/vagrant/.config/composer/vendor/bin

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,7 +2,9 @@ require 'json'
 require 'yaml'
 
 VAGRANTFILE_API_VERSION = "2"
-confDir = $confDir ||= File.expand_path("~/.homestead")
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
Nothing seems to be broken at the moment.

borela:
The virtual box image is, laravel commands don't work without modifying the PATH variable because it doesn't have the right path to composer's bin. Most apps are moving to the XDG standard to prevent polluting the user's home directory, that's why I changed homestead's config path.

PR #330, opened 2016-04-05
Adds option to ignore files added by homestead.

drecken:


svpernova09:
You shouldn't ignore some of these files, if you add Homestead to your project others will need these files to match your environment.

.vagrant should be ignored globally on you system or in your project. The Homestead.yaml file will be ignored by Laravel (if it's not in .gitignore you're on an older version of 5.x)

svpernova09:
You shouldn't ignore some of these files, if you add Homestead to your project others will need these files to match your environment.

.vagrant should be ignored globally on you system or in your project. The Homestead.yaml file will be ignored by Laravel (if it's not in .gitignore you're on an older version of 5.x)

drecken:
I was ignoring the `Homestead.yaml.example` file, which probably should not be ignored (kinda like `.env` and `.env.example`). 

The reason for ignoring the other files is because I use vagrant per project, while most of my coworkers have one machine. This option allows for quick `vendor\\bin\\homestead make --ignore` and `after.sh` and `aliases` could be committed as `after.sh.example` and `aliases.example` if needed.

I usually have to add those files to my `.gitignore` on each project and figured this would also be the case for others.

andrewhood125:
I generally try to avoid the situation where you have an `after.sh.example` commited and need to copy to `after.sh` on checkout. If you have bc incompatible changes it can become a mess, are people overwriting `after.sh` every time there is a change?

Also I would err on the side of less ignores and letting people handle untracked files rather than hiding something they potentially meant to commit.

svpernova09:
@drecken 

`./vendor/bin/homestead make` is designed to be run without overwriting existing `after.sh` or `aliases` files. If you are seeing behavior otherwise please let me know and I'll fix it. (Or feel free to submit a PR to fix it)

Your coworkers should easily be able to run the make command and get the exact same environment as you without having to create their own files. The `Homestead.yaml` file will always be recreated and thus shouldn't be committed because it's unique per user per path of the project folder where the make command is run.

Also +1 to less ignores @andrewhood125 especially when adding ignored versions of files you do actually need will only create confusion.

drecken:
It doesn't change the default behavior of the command, just adds an option to remove an extra step I find myself repeating on each new project. Thought others might find it useful. Thanks for having a look.

svpernova09:
@drecken Thanks for the PR. Hopefully my comments help clarify a better / alternative workflow regarding sharing your per project config.

PR #331, opened 2016-04-11
add hostupdater support in localized vagrant

gperdomor:
This PR is similar to #328, but in this case is for use homestead per project :D


svpernova09:
+1 looks good

PR #332, opened 2016-04-11
Adding default_ports option

dbpolito:
Default Port forwarding makes hard to use multiple homestead instances at the same time. With new introduced option, now we're able to simply do not forward any port at all, using direct VM IP instead of port forwards.

With new introduced option, it only disables if `default_ports: false`

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -40,7 +40,7 @@ class Homestead
     # Configure A Few VMware Settings
     ["vmware_fusion", "vmware_workstation"].each do |vmware|
       config.vm.provider vmware do |v|
-        v.vmx["displayName"] = "homestead"
+        v.vmx["displayName"] = settings["name"] ||= "homestead-7"
         v.vmx["memsize"] = settings["memory"] ||= 2048
         v.vmx["numvcpus"] = settings["cpus"] ||= 1
         v.vmx["guestOS"] = "ubuntu-64"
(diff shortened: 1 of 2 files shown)


scripts/homestead.rb
```diff
@@ -40,7 +40,7 @@ def Homestead.configure(config, settings)
     # Configure A Few VMware Settings
     ["vmware_fusion", "vmware_workstation"].each do |vmware|
       config.vm.provider vmware do |v|
-        v.vmx["displayName"] = "homestead"
+        v.vmx["displayName"] = settings["name"] ||= "homestead-7"
```
taylorotwell: Why this change? Default homestead is already on PHP 7.

dbpolito: I just added the same logic from virtual box provider:

https://github.com/laravel/homestead/blob/master/scripts/homestead.rb#L32

taylorotwell: Hmm, ok :)

taylorotwell:
So is this different than just removing the ports from the configuration?

dbpolito:
Yes because if you just remove the ports from config file the default_ports are going to be added anyway...

My initial idea was just, if there is a `port` config and it's a empty array, don't add default ports at all... Just add default ports if the `port` config is commented out, which is the default.

taylorotwell:
Personally I think this should maybe be called port_forwarding: false / true instead of default_ports: false

dbpolito:
In my opinion `default_ports` makes more sense because if you add ports to your config, and set `default_ports: false`, your custom ports are still going to be added...

`port_forwarding: false` makes me think no ports will be added at all... But i'm totally ok with that change...

PR #333, opened 2016-04-12
Update create-mysql.sh

george-oakling:
Create MySQL DB with UTF-8MB4 for emoji support and other special characters.


taylorotwell:
I do worry this will break quite a few people's migrations which use a default string length of 255 and if I recall correctly the maximum length for a utf8mb4 string index is in the 190s. This can probably be done when the default string size of Laravel string migration columns changes.

PR #334, opened 2016-04-20
Added `CHANGELOG.md`

tillkruss:
Only contains release notes for Homestead `v3.0.2` so far.


nelsonking:
hi
i get a question on branch master ,when i use vagrant up ,it's hit me can't find laravel/homestead,but i checkout 2.0 it's ok and the most is php-fpm7 i want use it ,but default whitch site-aviable php-fpm5 
what i can do ~

PR #335, opened 2016-04-27
append `privileged: false` for `afterScript` provisioner

justinmoh:
Realistically the shell provisioner script should be ran as `vagrant`, so that it could do things specific to `$USER` (which is only `vagrant` in homestead).

Using `sudo` to gain privilege is easy, but switching from `root` to another user is a mess -- and I couldn't get it working.

Based on this reason I think this `afterScript` should be ran as `privileged: false` by default.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -25,7 +25,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     Homestead.configure(config, settings)
 
     if File.exists? afterScriptPath then
-        config.vm.provision "shell", path: afterScriptPath
+        config.vm.provision "shell", path: afterScriptPath, privileged: false
     end
 
     if defined? VagrantPlugins::HostsUpdater


andrewhood125:
:+1: but this is going to be a backwards incompatible change.

justinmoh:
@andrewhood125 Ah, indeed. I'm new to open source collaborations, thanks for pointing out!

justinmoh:
It could be improved someday in the future, or just have both simultaneously, checks for:
- `after.sh` for `root` and,
- `user.sh` for `vagrant`.

A workaround for starters.

jonnywilliamson:
Just for info

Prefixing any command with `sudo -H -u vagrant` would have run the command as user `vagrant`

justinmoh:
@jonnywilliamson 

Thanks for the information, I've seen some of these examples in the previous day.

But I'm not really good in using unix system, so I had some hard time with these...

```
#!/usr/bin/env bash
# content of after.sh

sudo -H -u vagrant echo $HOME;    # /root
sudo -H -u vagrant npm config set prefix ~/.npm-global;
sudo -H -u vagrant npm config get prefix;    # /root/.npm-global

exit 0;
```

I couldn't target *myself* in vagrant's shell provisioner, perhaps it matters with how the provision script was run.

jonnywilliamson:
@justinmoh Strange, as this is working for me (prior to this PR)

```sh
    #!/usr/bin/env bash
    # content of after.sh


    #This script runs as ROOT. For most commands we need to run as vagrant
    #Prepending each command with this sudo command does this for us.
    #Maybe there is an easier way.
    USER="sudo -H -u vagrant"

<...snip...>

    # Make some MC ini settings.
    $USER touch /home/vagrant/.ssh/known_hosts   //File is created by user vagrant
```

PR #338, opened 2016-05-24
use with symfony 2 and 3

Basster:
for use with symfony 3 but homestead box 0.3.3 and php 5.6. otherwise composer install refuses to install homestead per project.

### DIFF
--- a/scripts/serve-symfony2.sh
+++ b/scripts/serve-symfony2.sh
@@ -11,12 +11,12 @@ block="server {
     server_name $1;
     root \"$2\";
 
-    index index.html index.htm index.php app.php;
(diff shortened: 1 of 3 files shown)


taylorotwell:
These should be two separate commits / PRs.

scripts/serve-symfony2.sh
```diff
@@ -11,12 +11,12 @@ block="server {
     server_name $1;
     root \"$2\";
 
-    index index.html index.htm index.php app.php;
+    index index.html index.htm index.php app_dev.php;
```
GrahamCampbell: I don't think this change is right?

PR #339, opened 2016-05-30
Extended the previous "artisan" alias to take in arguments...

jjsee:
Also made it shorter by adding the "art" alias for quick access.

### DIFF
--- a/src/stubs/aliases
+++ b/src/stubs/aliases
@@ -3,7 +3,7 @@ alias ...="cd ../.."
 
 alias h='cd ~'
 alias c='clear'
-alias artisan='php artisan'
+alias art=artisan
 
 alias phpspec='vendor/bin/phpspec'
 alias phpunit='vendor/bin/phpunit'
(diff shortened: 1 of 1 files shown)


src/stubs/aliases
```diff
@@ -44,3 +44,7 @@ function serve-proxy() {
         echo "  serve-proxy domain port"
     fi
 }
+
+function artisan() {
+    php artisan $1 $2
```
taylorotwell: I believe there is a way in Bash to just expand out all arguments instead of limiting it to the first 2.

taylorotwell:
This can be changed in your own aliases file I believe.

jjsee:
@taylorotwell There is currently an alias for php artisan that does not work with arguments (on line 6). This is basically to fix that so that it would accept arguments. 

Or if the direction is to keep this in our own aliases file, then perhaps that line should be removed? 

Either way I will be happy to submit a new PR to either update the artisan alias as per your feedback or to remove it entirely.

taylorotwell:
Sure feel free to update it here.

jjsee:
@taylorotwell Done. The artisan() function will now include all arguments instead of just the first two.

PR #340, opened 2016-06-02
Allowing server_name to differ from the map name

jessentp:
Currently the sites map name is used to define the server_name in the nginx config. If you have a subdomain setup within the application adding something like "map: site.tld subdomain.site.tld" creates a bad nginx server config.

This change would allow for a new property in the Homestead config file "server_name" so that additional server names can be listed so that you can test any subdomain routes within the same application/site configuration. If no server_name is given it will default to the map property.


taylorotwell:
Why not just use a wildcard subdomain which should already be be supported?

PR #341, opened 2016-06-05
Automatically set the users /etc/hosts or Windows hosts file with the 

alnutile:
help of https://github.com/cogitatio/vagrant-hostsupdater

I post here https://alfrednutile.info/posts/184 some very easy steps to using this library. Also I link to http://sherriflemings.blogspot.ca/2015/03/laravel-homestead-on-windows-8.html which shows it being used in Windows (and I used it there a bit).

Just thought if this was part of the Homestead install docs it might be a nice addition.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -16,7 +16,10 @@ class Homestead
     config.vm.box = settings["box"] ||= "laravel/homestead"
     config.vm.box_version = settings["version"] ||= ">= 0.4.0"
     config.vm.hostname = settings["hostname"] ||= "homestead"
-
+    
+    # Using https://github.com/cogitatio/vagrant-hostsupdater Alias are done automatically for you
+    config.hostsupdater.aliases = settings["aliases"]
+    
     # Configure A Private Network IP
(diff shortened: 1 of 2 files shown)


taylorotwell:
It causes errors if the plugin isn't installed.

Ingramz:
@alnutile some suggestions, because I'd like this to end up in homestead.

Surround your code with this, so that it's not required to have it installed
```rb
if Vagrant.has_plugin?("vagrant-hostsupdater")
  # plugin specific code here
end
```

and instead of duplicating aliases in config, capture them from `sites` -> `map` values ([usage](https://github.com/laravel/homestead/blob/master/scripts/homestead.rb#L155-L189)), unless there is a reason why one should define them separately.

alnutile:
Odd thing seems this is already in Laravel Homestead just not documented?

https://github.com/laravel/homestead/blob/master/Vagrantfile#L34-L36

@taylorotwell would it be helpful for me to do a PR and update the docs?

jonnywilliamson:
I think this was related to https://github.com/laravel/homestead/pull/159

alnutile:
@jonnywilliamson thanks for the link

PR #342, opened 2016-06-10
Use a patched Vagrantfile in favor of LocalizedVagrantfile

Ingramz:
It looks like there has been some useful pull requests that modify either `Vagrantfile` or `LocalizedVagrantfile`, but not both (even if the other party could benefit from that). Since the files aren't too different, it would be enough to just overwrite the default values for the configuration directory and homestead directory of the original `Vagrantfile`.

Also changed single quotes to double quotes in `Vagrantfile`, so it would look more uniform. In addition to that I moved the `vagrant-hostsupdater`-specific code to `homestead.rb` so it wouldn't fail if `sites` key was not defined in `Homestead.yaml`.


Ingramz:
I already see a way to improve the `make` command. Instead of copying and then patching, we could just read the file, modify it and then save it to the new location.

taylorotwell:
I specifically didn't want to get into modifying the VagrantFile. Would rather just have two files.

Ingramz:
@taylorotwell sorry to hear that. It seems a little odd, because rest of the configuration files already get modified programmatically and the differences between localized and regular vagrantfile can be reduced to two simple lines.

Would you like me to do the same PR sans the modification part, so there'd be feature parity?

PR #343, opened 2016-06-20
Symlink support for VirtualBox shares with Windows host

Swanty:
This adds symlink support for VirtualBox shares with Windows host.

After adding this and running `vagrant up` **with administrator privileges** it is possible to run something like `npm install` without using `--no-bin-links`


taylorotwell:
Holding off on this.

PR #346, opened 2016-06-28
Use MariaDB Xenial PPA

Zaghyr:
Hi !

Today I updated my Homestead box to 0.5.0 and activated MariaDB.
This didn't work because the MariaDB PPA is still for Ubuntu Trusty !

I updated to Xenial MariaDB PPA and now it works after a vagrant provision !


taylorotwell:
Thanks!

PR #347, opened 2016-07-05
Update mariadb mirror.

djohnston08:
image

After an upgrade to 0.5.0 the mariadb was consistently failing install.  I have modified the source list to pull from a good mirror.


djohnston08:
Closing, seems I was still on trusty repo and hadn't pulled in new change.

PR #348, opened 2016-07-11
Add Vagrant.require_version to Vagrantfile

jeremyjousse:
Forces to use Vagrant 1.8.4 and greater


taylorotwell:
Interesting. Didn't know you could do this. Thanks!

PR #350, opened 2016-07-16
Http2 Support

TheNodi:
Http2 is now supported by [major browsers](http://caniuse.com/#search=http2) and it should be taken into consideration when developing for the web. I wanted to test some laravel application under http2 and since homestead already supports https it is really easy to setup.

This changes allows you to setup http2 support from the homestead setting file. I added an "httpsOnly" flag to automatically redirect to https version of the site (since http2 works only on https for the majority of browsers).

Example:
```yaml
sites:
    - map: homestead.app
      to: "/home/vagrant/laravel/public"
      http2: true
      httpsOnly: true
```

I have to apologies for the quite bad bash syntax, I would like to tight them up but I'm not good enough in bash (I know only the basics). I commented some of the statement that might look confusing, let me know if I have to edit something, I'll be happy to do so.

Note:
Requiring homestead >= 0.5 is optional. On Ubuntu 14.04 http2 won't work in Chrome because of ALPN protocol. More detailed reason can be found [here](http://serverfault.com/a/733556).

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -164,10 +164,21 @@ class Homestead
           type = "symfony2"
         end
 
+        http2 = (site.has_key?("http2") && site["http2"]) ? 1 : 0
+
+        httpsOnly = (site.has_key?("httpsOnly") && site["httpsOnly"]) ? 1 : 0
+
         config.vm.provision "shell" do |s|
           s.name = "Creating Site: " + site["map"]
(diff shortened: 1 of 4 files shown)


taylorotwell:
I would just make it the default. No need to introduce a new configuration option.

TheNodi:
That sounds awesome.
I'm gonna close this PR. Let me know if you want me to send another PR with http2 as default.

PR #351, opened 2016-07-25
add ability to serve apache sites

browner12:
as much as I prefer Nginx, the company I work at has a lot of sites that run on apache (lots of redirects and such in their `.htaccess` files). while I am trying to move the sites to be more webserver agnostic, for the time being I need an easy way to setup apache sites on Homestead to be able to get them running locally and edit them.

in the past they have been building separate VMs for each site, so anything I can do to lower the barrier to entry and get them running everything through Homestead is a big help.

btw, i realize the SSL portion doesn't actually do anything yet. i figured this was a good starting point.

thanks,

### DIFF
--- a/scripts/serve-apache.sh
+++ b/scripts/serve-apache.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+
+mkdir /etc/apache2/ssl 2>/dev/null
+
+PATH_SSL="/etc/apache2/ssl"
+PATH_KEY="${PATH_SSL}/${1}.key"
+PATH_CSR="${PATH_SSL}/${1}.csr"
+PATH_CRT="${PATH_SSL}/${1}.crt"
+
+if [ ! -f $PATH_KEY ] || [ ! -f $PATH_CSR ] || [ ! -f $PATH_CRT ]
+then
+  openssl genrsa -out "$PATH_KEY" 2048 2>/dev/null
+  openssl req -new -key "$PATH_KEY" -out "$PATH_CSR" -subj "/CN=$1/O=Vagrant/C=UK" 2>/dev/null
+  openssl x509 -req -days 365 -in "$PATH_CSR" -signkey "$PATH_KEY" -out "$PATH_CRT" 2>/dev/null
+fi
+
+block="<VirtualHost *:80>
+       # The ServerName directive sets the request scheme, hostname and port that
+       # the server uses to identify itself. This is used when creating
+       # redirection URLs. In the context of virtual hosts, the ServerName
+       # specifies what hostname must appear in the request's Host: header to
+       # match this virtual host. For the default virtual host (this file) this
+       # value is not decisive as it is used as a last resort host regardless.
+       # However, you must set it for any further virtual host explicitly.
+       #ServerName www.example.com
+
+       ServerAdmin webmaster@localhost
+       ServerName $1
+       ServerAlias www.$1
+       DocumentRoot $2
+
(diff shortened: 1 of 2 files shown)


josecybertill:
@browner12 how can I use this feature? I tried adding 'type: apache' to my Homestead.yaml 'sites' configuration, is that right?

browner12:
I don't believe there is a 'type' option for the sites.

i'm not sure if the Virtualhosts are created automatically on provisioning or not. If not, you will need to log into your Homestead machine (`homestead ssh`) and then type

```sh
serve-apache(domain.app, /home/vagrant/code/domain_public_directory)
```

You will also need to install Apache, because Homestead does not do that automatically.

Add this to your `after.sh`

```sh
# install apache
sudo apt-get update
sudo apt-get install -y apache2

# install apache php module
sudo apt-get install libapache2-mod-php7.0

# enable apache rewrite module
sudo a2enmod rewrite
```

You can then cycle between using Apache and Nginx by using

```sh
sudo service nginx stop
sudo service apache2 start
```

and

```sh
sudo service apache2 stop
sudo service nginx start
``` 

Hope this helps.

svpernova09:
@josecybertill 

Yes sites type: apache should be what you want.

@browner12 

We break out site types to support multiple vhost types via https://github.com/laravel/homestead/blob/master/scripts/homestead.rb#L178 Apache support was fixed in #420 

Homestead users should be ready for Apache support to end soon™ To continue using a feature Homestead removes just lock your Homestead version to the previous release.

browner12:
oh cool! didn't know there was a `type` option. didn't see it in the documentation anywhere.

@svpernova09 what does ending support entail? I feel like we don't really *support* it now other than the `serve-apache` helper function.  I still have to install and configure Apache on my own. 

As I've said before, I'm not a fan of using Apache, but sometimes I have to.  Are there other Apache features that I'm unaware of that will be removed?

svpernova09:
@browner12 You don't need to install apache, it's sneakily hidden in the serve script: https://github.com/laravel/homestead/blob/master/scripts/serve-apache.sh#L3 

The only reason it's still there is because someone shared a PR to add SSL support to apache #419 and in the process of testing that I realized the functionality was completely broken, which resulted in #420 

My primary goal for Homestead is to support Laravel users the best way we can in a vagrant environment. Thousands (likely more) of PHP developers rely on Homestead to solve a lot of problems they'd otherwise have to solve themselves. I'm not against supporting stuff Laravel doesn't natively if the overhead for me to do so isn't too great. Obviously that's subject to change at any given time for the greater good of the Laravel user.

Case in point: Laravel no longer supports HHVM. Homestead still supports HHVM because I believe that should warrant a major version number change but also HHVM isn't a pain point for me to build Homestead at the moment so it's just hanging around for a bit. Any day now I or @taylorotwell could get a wild hair and go nuts gutting stuff we don't need anymore and likely HHVM and Apache would be at the top of the chopping block. I'm confident that our versioning strategy with Homestead and now also Settler will make it relatively painless for users to lock themselves (or a project) at a specific version of Homestead to retain any features we may remove in the future.

Also, the sites `type:` feature isn't documented (like a few other things aren't) because these features usually start out as a PR from someone and it's sometimes a bit of a hack job, but it works, and it doesn't break anything so it gets merged. Usually these are edge case changes that go undocumented. For sites `type:`, maybe it's time to document that.

(Sorry for the wall of text)

browner12:
thanks for the info. 

I agree, if it becomes a pain point to maintain, maybe we get rid of it.  I've just been installing it manually so far, anyway.

PR #352, opened 2016-07-28
Added site queue key to setup supervisor workers.

davidrushton:
To save having to manually set up queue worker processes using supervisor, this PR adds a new `queue` key which triggers an automatic worker config file and starts the process with supervisor.

```
sites:
	- map: homestead.app
	  to: /home/vagrant/Code/Laravel/public
	  queue: true
```

The worker conf file is setup in `/etc/supervisor/conf.d` and follows the following stub:

```
[program:$1-worker]
process_name=%(program_name)s_%(process_num)02d
command=php $2/../artisan queue:work --sleep=3 --tries=3 --daemon
autostart=true
autorestart=true
user=vagrant
numprocs=8
redirect_stderr=true
stdout_logfile=$2/worker.log
```

Tested on homestead 3.0.2 and it appears to work ok :)

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -185,6 +185,19 @@ class Homestead
           end
         end
 
+        # Configure Supervisor Queue Workers
+        if (site.has_key?("queue"))
+          config.vm.provision "shell" do |s|
+            if (site["queue"])
+              s.path = scriptDir + "/queue-worker.sh"
+              s.args = [site["map"].tr('^A-Za-z0-9', ''), site["to"]]
+            else
+              s.inline = "rm -f /etc/supervisor/conf.d/$1-worker.conf"
+              s.args = [site["map"].tr('^A-Za-z0-9', '')]
+            end
+          end
+        end
+
       end
     end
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Going to hold off on setting up supervisor in dev. Typically would just run queue workers in console window.

davidrushton:
@taylorotwell That's a shame - I think this would be useful, particularly on project-specific homestead boxes, and it would be completely optional.

Thanks anyway.

marcelogarbin:
Hello, have you been able to automatically run queues in homestead (without having to type manually)? Thanks

davidrushton:
@marcelogarbin I've occasionally added a queue daemon to Homestead's `/etc/supervisor/conf.d`, but since the release of Horizon I just run `php artisan horizon` in a console tab when testing the queue works as expected on local. All other times I just use the sync driver instead.

marcelogarbin:
@davidrushton Thanks for the feedback.

I did not know Laravel Horizon, I tested last night, very interesting, I'm using REDIS. However I had a doubt, for example I tested sending 10 emails using mailtrap, as they were simultaneous sends it ends up blocking and the job failing. My question is, will these failed jobs be added back for submission or do I have to manually submit? Which is more suitable, REDIS or Database? In production I have a machine with some projects sharing the same hardware, can this be a problem?

PR #355, opened 2016-08-15
Updated homestead.rb to run commands directly.

Utkarsh-vishnoi:
Updated homestead.rb to run shell commands directly form homestead.yaml file.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -88,6 +88,16 @@ class Homestead
         config.vm.network "forwarded_port", guest: port["guest"], host: port["host"], protocol: port["protocol"], auto_correct: true
       end
     end
+    
+    # Run Custom Commands After Booting Up
+    if settings.include? 'commands'
+      settings["commands"].each do |command|
+        config.vm.provision "shell",:run => 'always' do |s|
+          s.name = command["name"]
+          s.inline = command["command"]
+        end
+      end
+    end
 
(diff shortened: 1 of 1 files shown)


andrewhood125:
So `after.sh` gets run only on `vagrant provision` or initial `vagrant up` and this would get run after`vagrant up` only? I'd rather rename `after.sh => provision.sh` and then add a `up.sh`.

svpernova09:
What is the use case for running shell commands from the `homestead.yaml` that seems very odd. If you need to run shell scripts add them to `after.sh`, if you need them to run _every_ time you boot vagrant than they should be added to the base box instead of Homestead. (this may mean creating your own box it's a one off thing that doesn't fit Homestead)

Utkarsh-vishnoi:
How exactly I am gonna run a command for each and every-time I do `vagrant up`. I wanna do this Because setting up homestead is easy then entirely configuring another base box.
If you got any other method that is not odd. I would love to hear it. :+1:

svpernova09:
@Utkarsh-vishnoi If you are running a shell script every time you do `vagrant up` I'm saying that's not common in the typical vagrant workflow. The better option (Better than changing how Homestead works for everyone) would be to use `after.sh` to add a cron job to run on every boot (You'd want `@reboot` I believe for this) that runs your script from your project folder.

Utkarsh-vishnoi:
OK

PR #356, opened 2016-08-17
Vagrant >= 1.8.4 is not required anymore

taliptako:
We dont need to force people to use Vagrant >= 1.8.4

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,8 +14,6 @@ aliasesPath = confDir + "/aliases"
 
 require File.expand_path(File.dirname(__FILE__) + '/scripts/homestead.rb')
 
-Vagrant.require_version '>= 1.8.4'
-
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     if File.exists? aliasesPath then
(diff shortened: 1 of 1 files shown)


morloderex:
I believe that the issues with vagrant versions piror to vagrant 1.8.4 will still have issues when running homestead.

taliptako:
Many offical repositories of Linux distributions have older version from 1.8.4 so i also have and i'm manually removing this line always ...

jonnywilliamson:
One of the biggest issues was this:
https://github.com/mitchellh/vagrant/pull/6493

Basically, trying to run gulp and npm in the homestead VM, (whilst running on windows host), caused a lot of issues because the filepaths created during NPM installation where too long.

It causes lots of errors, but finally the guys at Vagrant found a fix to that solution.

It might have arrived in vagrant 1.8.2 or similar, but there was a further bug that was squashed by version 1.8.4.

It now works reliably and as such, I think requiring a minimum of vagrant 1.8.4 is a good idea,

Utkarsh-vishnoi:
Yeah i had the same problem too, but when i installed gulp on both in homestead and my host OS(windows). It worked.

PR #358, opened 2016-08-24
Allow to install updated symfony component when working with PHP 5.6.

crynobone:
Still stuck with PHP 5.6 for a few months, would be nice if we can have this to the existing `2.2.x` releases.

Signed-off-by: crynobone <crynobone@gmail.com>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,8 +3,8 @@
     "description": "A virtual machine for web artisans.",
     "require": {
         "php": ">=5.4",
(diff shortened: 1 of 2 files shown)


crynobone:
@taylorotwell could we have `v2.2.2` release soon, here the only different:

https://github.com/laravel/homestead/compare/v2.2.1...2.0

taylorotwell:
Done.

crynobone:
Thanks <3

PR #359, opened 2016-08-26
Adds configuration options for SMB and NFS versions

acmitch:
Adds configuration options for easy SMB and NFS versioning. Specifically, Homestead cannot run via SMB on Windows 7 due to SMB needing 2.01 version.


taylorotwell:
You can use the existing `mount_options` variable for this I believe.

PR #360, opened 2016-09-08
Install HHVM if the user wants, when using MariaDB.

blaise-zaga:
During the course of uninstalling MySQL when the user selects to use MariaDB, HHVM is also uninstalled.

This script that is added here is used to get HHVM back.

### DIFF
--- a/install-hhvm.sh
+++ b/install-hhvm.sh
@@ -1,0 +1,12 @@
+# install HHVM
+sudo apt-get -y install hhvm
+
+# ensure webserver can talk to HHVM over FastCGI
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can we figure out what HHVM is uninstalled when MySQL is and somehow just prevent that from happening in the first place?

blaise-zaga:
Hi Taylor,

This section of code `apt-get remove -y --purge mysql-server mysql-client mysql-common
apt-get autoremove -y
apt-get autoclean`

in homestead/scripts/install-maria.sh is responsible for removing HHVM.

PR #361, opened 2016-09-15
Install HHVM if the user wants, when using MariaDB.

blaise-zaga:
During the course of uninstalling MySQL when the user selects to use MariaDB, HHVM is also uninstalled.

This script that is added here is used to get HHVM back

Responded to your concerns here - https://github.com/laravel/homestead/pull/360


taylorotwell:
I don't want to "re-install HHVM", I want to not remove it in the first place.

PR #364, opened 2016-10-09
Vagrantfiles coherence

Djuuu:
Base `Vagrantfile` and per-project `LocalizedVagrantfile` underwent "asymmetric" modifications.
However, most changes would be pertinent on both sides.
This puts back both files as close as possible, applying all "missed" changes.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,7 +4,7 @@
 require 'json'
 require 'yaml'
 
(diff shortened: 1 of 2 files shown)


Djuuu:
To avoid this kind of problem in the future, I think we could try and use only one `Vagrantfile`.
That's what I tried in 0692418, but I'm not sure it's elegant enough.

PR #366, opened 2016-10-24
Run composer self-update from vagrant user

Swanty:
Before this fix `composer self-update` was run directly as root and didn't add the missing keys to `/home/vagrant/.composer` (`keys.dev.pub` and `keys.tags.pub`)

if you try run `composer diagnose` before this fix then you will see that it shows error of missing keys


Swanty:
Probably another solution is to just simply add the missing `keys.dev.pub` and `keys.tags.pub` to the homestead vagrant box itself

PR #367, opened 2016-11-28
Add gui to vmware and virtualbox

jordanramstad:
This just allows a new homestead option to use the "gui" right from the `Homestead.yaml` file. I have run into problems more specifically to vmware where the errors are not descriptive enough from the vagrant API alone to let you diagnose the problem.

This allows a new option:

```
gui: true
```

That will tell homestead to enable the gui option, allowing for better debugging without needing to manually edit the homestead package.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -36,6 +36,9 @@ class Homestead
       vb.customize ["modifyvm", :id, "--natdnsproxy1", "on"]
       vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
       vb.customize ["modifyvm", :id, "--ostype", "Ubuntu_64"]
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for the PR. Will you please squash your 3 commits into 1. Once done looks good from here

👍

jordanramstad:
Thank you, I have done the adjustment.

svpernova09:
Thanks! Tested here and it's ready to go.

jordanramstad:
Just a note this does not add the option to parallels since its mac only and I do not have access to a mac nor am I familiar with parallels.

svpernova09:
Good thing to point out. Shouldn't be an issue since it's isolated to the Virtualbox and VMWare providers.

PR #372, opened 2016-12-02
Fixed the issue of expired password

marabesi:
Hello guys,

Recently I've through a problem in my vagrant environment related with expired password. The problem was in the provision, here is the error that I was getting:
pr1

After a while trying to fix it I've dived into the code and did a quick change, but then a new error had arrived

pr2

And one more time I did a little change in the scripts folder to fix this issue

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -9,4 +9,5 @@ EOF
 
 DB=$1;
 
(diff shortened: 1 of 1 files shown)


svpernova09:
I'm not seeing this issue, can you open an issue and fill out the template: https://github.com/laravel/homestead/issues/new

Thanks!

marabesi:
@svpernova09  https://github.com/laravel/homestead/issues/373

: )

svpernova09:
Thanks for the issue, I replied there. Closing this PR since it's not an issue if you're keeping versions of Homestead and the box up to date.

Thanks for the PR.

PR #379, opened 2016-12-06
Link to 5.3 documentation

edcs:
Updated link in readme to point to latest version of the docs.


svpernova09:
The 2.0 branch is old and should point to older versions. Possible we could just nuke this branch since I don't think any 5.x version of Laravel would run on 2.x version of homestead.

Thanks for the PR.

PR #381, opened 2016-12-06
Block for review prod environment in local

nietzcheson:


nietzcheson:
@svpernova09 Yes. I know it looks ugly. Ugly. But, for now is the only way that works for me and I think it could serve.

In Symfony the dev environment skips many validations. If you rely on developing locally, everything in dev, and then launches to production there could be many problems ...

If there is a better way, please share it.

svpernova09:
@nietzcheson Do these 2 blocks not do what you're wanting?

https://github.com/nietzcheson/homestead/blob/99eb3539e6ab7c446f1ea788f62066cf22edd48e/scripts/serve-symfony2.sh#L44-L66

nietzcheson:
@svpernova09 Modify the script so that by default nginx over write /app.php and if the dev environment is required the developer type /app_dev.php. This avoids commenting on lines of text

taylorotwell:
Couldn't we just have like symfony and symfony_dev blocks be separate site types?

taylorotwell:
I would be more interested in moving that direction rather than this approach.

svpernova09:
@taylorotwell THat's what I was hoping for, I just haven't had time to test it. (Or a modern demo Symfony app)

PR #386, opened 2016-12-07
allow trailing slash

tuanalumi:
change `+` to `*` to allow trailing slash handling


svpernova09:
Is this more of an application level decision than a server level?

tuanalumi:
Hhhmmm. I think it's better for general use when it can handle more cases.

PR #388, opened 2016-12-07
Update after.sh

lioannou:
I find this a useful command - so I thought I would share.

Having this as part of the `after.sh` means my Git is always configured correctly after any `destroy` and `provision`.

### DIFF
--- a/src/stubs/after.sh
+++ b/src/stubs/after.sh
@@ -3,3 +3,7 @@
 # If you would like to do some extra provisioning you may
 # add any commands you wish to this file and they will
 # be run after the Homestead machine is provisioned.
+
+# By using the two commands below - you can have your Git profile added automatically after any vagrant provisioning
+# git config --global user.email "Your Email Here"
+# git config --global user.name "Your Name Here"


taylorotwell:
Hmm, I'm not sure it's totally typical to use Git inside of the virtual machine. Thanks though!

lioannou:
@taylorotwell - @svpernova09 mentioned something similar to me as well.

But on a Windows box it makes perfect sense. I dont have Git installed on Windows. I simply install Homestead, and then do *everything* inside the virtual machine. Otherwise you run into other issues with Windows when you dont do it inside.

Otherwise, if we should not be using Git inside the virtual machine, then why have Git as part of Homestead?

jordanramstad:
I have never really run into a problem using git on windows. I guess it depends on your workflow though. Most GUI's make Git Bash easily accessible and Sourcetree for instance lets you open Git Bash right in the repo directory. I find although using the git commands works, its quicker and more beneficial to my workflow to use a GUI and I always have the ability to do it manually if need be.

Anyway the point being that although git may be handy to use through the virtual machine its much easier to use it in the host (windows). I have actually never run into any issues with git specific to windows.

Anyway just my 2 cents on it as a fellow windows user.

Issue #384, opened 2016-12-06
[Proposal] Set UTF8mb4 as the default database character set / collate


jonnywilliamson:
With a lot of web apps these days, storing data such as emojis in messages etc has become a common requirement.

Unfortunately the default character set for MySQL in homestead seems to be UTF-8. This doesn't allow the full range of emoji characters (and others) to be stored, and results in a lot of database errors if these types of characters are in the data to be saved.

By using the `UTF8mb4` character set, this limitation is bypassed.

I was unable to raise this issue in Laravel settler, but would the project team be at all interested in changing the (default) character set for the homestead database to `UTF8mb4`?

It would be a simple edit of this line https://github.com/laravel/settler/blob/master/scripts/provision.sh#L204

to 

`mysql --user="root" --password="secret" -e "CREATE DATABASE homestead character set UTF8mb4 collate utf8mb4_bin;"`

Just wanted some thoughts before making the PR.

Some other links to this: 
https://twitter.com/adamwathan/status/756569942397419520
http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci/766996#766996

thecrypticace:
It should probably be changed at the database config level (during provision in settler); not just for the create database statement. If you do not change it in the configs new databases could still default to latin1 or utf8 instead of utf8mb4.

svpernova09:
Will have to revisit after Laravel 5.4 since this would default people into having issues since the default string length is 255. https://github.com/laravel/homestead/pull/391

PR #391, opened 2016-12-07
:sparkles: :snowflake: Use character set UTF8mb4 and collate utf8mb4_…

svpernova09:
…bin when creating databases

Addresses #384


taylorotwell:
Will cause problems until Laravel 5.4 is released I believe since the default string length is 255.

svpernova09:
Ah, I see: https://laracasts.com/discuss/channels/general-discussion/migrations-configure-a-default-string-length-utf8mb4 

If we accept this now, we'll need big warning signs when people create tables. Not sure that's worth it atm. Will close a revisit after 5.4

PR #397, opened 2016-12-15
:ambulance: :checkered_flag: Fix line endings on aliases file if needed, fixes #396

### LINKED ISSUE #396: Alias file is not used in per project installs
In order to update the aliases for a per project install of homestead you need to manually run this script after starting up the vm for the first time.  Is there a way to have this run during the initial provision similar to global installs of homestead.  The after.sh bash file is used, but putting this script into that file doesn't work.

```
tr -d '\r' <~/.bash_aliases >~/tmp
mv ~/tmp ~/.bash_aliases
unalias -a
source .bash_aliases
```

## Versions
  - Vagrant: 1.9.1
  - Provider: Virtualbox 5.1.10
  - Homestead: 0.6.0

## Host operating system
Windows 10 (64-bit)

## Homestead.yaml
Not Applicable

## Vagrant up output
Not Applicable

## Expected behavior
What should have happened?
The `alias` file included in the root folder of a per project install should be picked up and applied to bash aliases on provision similar to a global install of homestead.

## Actual behavior
What actually happened?
No aliases are included and instead a bunch of command not found errors are show when first entering the vm.

## Steps to reproduce
1. Install per project Laravel project with homestead dependency as per docs
2. Add alias file to root of the project
3. vagrant up and aliases aren't added

svpernova09:
:ambulance: :checkered_flag: Fix line endings on aliases file if needed, fixes #396

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,6 +19,9 @@ Vagrant.require_version '>= 1.8.4'
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     if File.exist? aliasesPath then
         config.vm.provision "file", source: aliasesPath, destination: "~/.bash_aliases"
+        config.vm.provision "shell" do |s|
+          s.inline = "awk '{ sub(\"\r$\", \"\"); print }' /home/vagrant/.bash_aliases > /home/vagrant/.bash_aliases"
+        end
     end
 
     if File.exist? homesteadYamlPath then


andrewhood125:
Very much against this. It's not the responsibility of homestead to handle `\r\n => \n` conversion. What if they edit `after.sh` or any other file that we have or may have in the future. We'll have to do a `awk '{ sub(\"\r$\", \"\"); print }'` for each one. 

Possible alternatives:
  - What if the edit the file from the vm? Add a warning to the top of both files. `# Don't edit on your windows box`. 

  - Git can detect windows line endings and convert. 

  - Check all homestead files once somewhere in our code and look for `\r\n` `grep`? Warn the user to handle it.

svpernova09:
It's not the solution I'd prefer. I'm open to other ideas that aren't "Just tell users to only edit files that respect line endings".

I believe `aliases` is the only file we'd have to do this with since it's the only file that a user should be editing that we copy from the host to the guest and then executed on the guest side (causing line endings issue)

This is certainly a bug in Windows because it's something that works fine in MacOS and Linux, yet fails on Windows.

It's an inelegant solution to make it easier on Windows users. Otherwise they'd have to manually add conversion scripts into their `after.sh`.

jordanramstad:
I think the problem stems from them being different OS's. Mac is a closet linux box so it uses a similar kernel, and that kernel is the same as homesteads (probably oversimplifying this point though). So its not a bug in Windows, just a result of different operating systems.

Anyway I am a windows user, and with the right IDE/editor this never really becomes a problem. I understand some people need to use certain editors that might not have this feature so I would have to +1 this idea, it brings the functionality that "recommended" editors have, making Homestead more accessible and "magical" 👍 

Think in the mind of someone just starting out with laravel and uses windows. If they get locked up in this problem and find that the solution is to change the programs they use they are likely to not get far before giving up. I originally had some problems when first starting out as well and DID give up, it was when my job decided to use it I took another look and am thankful I did.

The entire point of using VM's is to run the application is a dedicated container, the host should not matter. If there is a solution that removes a barrier on a specific host I think its a good idea to go for it.

PR #401, opened 2016-12-21
Show DB name when creating while provision

caiquecastro:
Show which database is being created while the provision.


svpernova09:
Thanks for the PR!

PR #404, opened 2016-12-21
The .my.cnf could be available for vagrant user

JuanDMeGon:
Currently, this configuration is only available for the root user. Making it available for the vagrant user, may provide easy access to the database command line inside the virtual machine.

With this, after SSH into the virtual machine, the developer can easily issue: `mysql `to get into the database command line, instead of `mysql -u homestead -p`

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -7,6 +7,8 @@ password = secret
 host = localhost
 EOF
 
+cp ~/.my.cnf /home/vagrant/.my.cnf
+
 DB=$1;
 
 mysql -e "CREATE DATABASE IF NOT EXISTS \`$DB\` DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_unicode_ci";
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -1,13 +1,11 @@
 #!/usr/bin/env bash
 
-cat > ~/.my.cnf << EOF
+cat > /home/vagrant/.my.cnf << EOF
 [client]
 user = homestead
 password = secret
 host = localhost
 EOF
-
-cp ~/.my.cnf /home/vagrant/.my.cnf
 
 DB=$1;
 


svpernova09:
Hi @JuanDMeGon I'll give this a spin. Most likely it's something we'll want to implement in https://github.com/laravel/settler/blob/master/scripts/provision.sh rather than Homestead.

svpernova09:
@JuanDMeGon Actually, instead of Settler, lets keep this here, but instead of adding the line `cp ~/.my.cnf /home/vagrant/.my.cnf` lets instead change https://github.com/laravel/homestead/pull/404/files#diff-098793c438d3da71b20f99ccc5c29c91L3 to be `cat > /home/vagrant/.my.cnf << EOF`.

If you want to make this change go for it.

JuanDMeGon:
Yeah, I agree with that. I just used copy, if someone wanted to keep that for root user too. I'll change that so.

JuanDMeGon:
Ok, already modified the path. Should work correctly :)

svpernova09:
@JuanDMeGon Thanks, will you squash the commits please. then we're good to go.

svpernova09:
Merged. Thanks for the PR!

JuanDMeGon:
Thanks for merging. Sorry for the delay, I just squashed my commits, but too late, my bad.

svpernova09:
No worries at all, I always forget I can do it when I merge the PR. :D

JuanDMeGon:
Hello @svpernova09. I just realized that if we do not keep the .my.cnf for the root user, the Homestead virtual machine deployment fails with mariadb enabled.

Mariadb installation, try to establish the connection without credentials, wich result in an "Access denied for user 'root'@'localhost'"

It seems like we can keep the .my.cnf file for both vagrant and root, but it seems a little inconsistent with the install-maria.sh file, because it uses the credentials directly in the mysql command.

The good way to go (in my opinion) is to use the credentials in the create-mysql.sh for the mysql command as usual.

I'll make  PR with this proposal to solve the problem.

svpernova09:
Sounds good. Thanks for catching this.

PR #406, opened 2016-12-22
Using credentials in mysql command as usual

JuanDMeGon:
To avoid "Access denied" problem specifically for maridb installations, it is required to use the database credentials directly on the command.


svpernova09:
Thanks for the PR.

Issue #407, opened 2016-12-22
MySql Warnings (Red Text During Vagrant Up)


svpernova09:
With #406 and #404 we introduced a change that now causes red text:

```
==> homestead-7: mysql:
==> homestead-7: [Warning] Using a password on the command line interface can be insecure.
```

During `vagrant up` Now while this is a minor issue and nothing is really "broken" I'd prefer to not show the user any red text at all. (To prevent casting any doubt that everything is ok)

The likely fix is to go back to copying `/root/.my.cnf` to `/root/.` and then copying it to `/home/vagrant/.my.cnf` as well. 

We need to check and make sure this works with maria as well. Currently we're using passwords on the command line with MariaDB as well, I haven't checked yet to see if those also show up in red. If they do we should take steps to resolve that as well.

/cc @JuanDMeGon

JuanDMeGon:
Got it. I will be back to the original idea. Create and then copy the .my.cnf file from root to vagrant.
I'll check if the mariadb credentials can be removed from the .sh file.

svpernova09:
👍 Sounds good. Thanks, sorry for the run around

JuanDMeGon:
Ok. Playing a little with Mariadb enabled and disabled it seems like when mariadb is enabled there is not any warning about the usage of credentials in the command prompt. However when mariadb is disabled, it means MySQL is used, the warning comes out.

So it seems like the warning can be removed, only using the original proposal. Cat the /root/.my.cnf and then copy to /home/vagrant/.my.cnf.

svpernova09:
Ok, lets do that, but lets use full paths so we're explicit where files are: `cp /root/.my.cnf /home/vagrant/.my.cnf`

JuanDMeGon:
Just sent the PR to solve the warning while using MySQL: #408.

In the way. Remove the credentials from the install-maria.sh is not trivial.

The .my.cnf file is created after the mariadb installation, so if the credential are removed, the authentication will fail.

A possible solution is to create this file in the install-maria.sh at the beginning and then in the create-mysql.sh create the same file if does not exists, with the respective copy for the vagrant user.

PR #408, opened 2016-12-22
Solving warning for credentials usage

JuanDMeGon:
Removing the credentials from the mysql command #407 . A copy of the .my.cnf file is created for the vagrant user, to allow easy access to the database.

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -1,12 +1,14 @@
 #!/usr/bin/env bash
 
-cat > /home/vagrant/.my.cnf << EOF
+cat > /root/.my.cnf << EOF
 [client]
 user = homestead
 password = secret
 host = localhost
 EOF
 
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks @JuanDMeGon !

scripts/create-mysql.sh
```diff
@@ -1,12 +1,14 @@
...
 
-cat > /home/vagrant/.my.cnf << EOF
+cat > /root/.my.cnf << EOF
 [client]
 user = homestead
 password = secret
 host = localhost
 EOF
 
+cp /root/.my.cnf /home/vagrant/.my.cnf
```
andrewhood125: I'm curious what the file permissions for this is after creating it this way. 🤔 Any issues with `mysq`l reading it?

JuanDMeGon:
All these commands are executed as root. MySQL and MariaDB shouldn't have any problem reading this. I don't know if permissions are an issue here, it is for development environment.

PR #416, opened 2016-12-30
Remove the config.ssh.shell directive. This breaks the ability to run

cwandling:
vagrant ssh -c command

This was added to prevent an error message from appearing when
sshing into a vagrant box.  It was a red herring because there
was no actual problem.

If users wand to get rid of the message then the can add

    config.ssh.shell = "bash -c 'BASH_ENV=/etc/profile exec bash'"

to their Vagrantfile

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -6,9 +6,6 @@ class Homestead
     # Configure Local Variable To Access Scripts From Remote Location
     scriptDir = File.dirname(__FILE__)
 
-    # Prevent TTY Errors
-    config.ssh.shell = "bash -c 'BASH_ENV=/etc/profile exec bash'"
-
     # Allow SSH Agent Forward from The Box
     config.ssh.forward_agent = true
 


svpernova09:
Thanks for the PR @cwandling Can you open an issue with the problem so I can verify the bug: https://github.com/laravel/homestead/issues/new

Thanks!

cwandling:
You bet…

cwandling:
Done.  Thanks!

svpernova09:
@cwandling 

We can't accept as is. The reason that line exists is to hide the red messages you see here:

screenshot 2016-12-30 15 05 19

We need to either find another way to hide those messages before this could be accepted.

jonnywilliamson:
Perhaps this long thread might help @cwandling 

https://github.com/mitchellh/vagrant/issues/1673#issuecomment-168205206

cwandling:
Thanks.

Tried to get it to work but failed…. :-(

cwandling:
Sorry about that.

Am I correct that the messages are a red herring?

Cory

svpernova09:
Essentially yes.  We don't want to expose all those messages and have to field a ton of error reports when the scripts are running fine.

cwandling:
Understandable, just wanted to make sure that if I hack my local copy I am not screwing things up.

PR #419, opened 2017-01-05
SSL support is added to Apache virtual hosts

umutkirgoz:
SSL support is added to Apache virtual hosts

### DIFF
--- a/scripts/serve-silverstripe.sh
+++ b/scripts/serve-silverstripe.sh
@@ -1,0 +1,122 @@
+#!/usr/bin/env bash
+
+mkdir /etc/nginx/ssl 2>/dev/null
+
+PATH_SSL="/etc/nginx/ssl"
+PATH_KEY="${PATH_SSL}/${1}.key"
+PATH_CSR="${PATH_SSL}/${1}.csr"
+PATH_CRT="${PATH_SSL}/${1}.crt"
+
+if [ ! -f $PATH_KEY ] || [ ! -f $PATH_CSR ] || [ ! -f $PATH_CRT ]
+then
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure we should even be supporting Apache. I know nothing about it and it's not supported on Forge or Valet.

svpernova09:
@umutkirgoz Thanks for the PR! I also realized that Apache was totally broken and I'm surprised it was working at all for you. Apache is fixed and working again. This will probably be the last update we'll see on Apache since Forge doesn't support it, (as Taylor mentioned). Homestead will likely stop supporting it sometime in the future.

browner12:
as much as I wish I could do everything on nginx, I work on some projects that require apache (more specifically require their htaccess file) to work.  While I am pushing towards refactoring these sites to be web server agnostic, it is a slow process.

it is really nice to have Apache available on Homestead, so I can switch over quickly for the projects that require it.  I would vote to not remove it as long as it's not hurting anything.

I try and be an advocate for Laravel and all its great tools whenever I can, and having support for things like apache, in addition to the preferred and suggested software, helps smooth the transition for people reluctant to change.

my 2c. thanks.

umutkirgoz:
[We](https://github.com/orgs/mynetcom/people) are using Laravel and its homestead dev environment when building the future of [Mynet](http://www.mynet.com/) (Biggest portal of Turkey with ~1 billion pageviews per month)

I plan to build a homestead fork with multiple web server support (or maybe default nginx, and optional apache) 

This was the first step of fixing the provisioning problem of Apache virtual hosts. Vagrant box needs to be customized...

tHanks for interest...

Regards.

PR #422, opened 2017-01-06
Use fastcgi-php.conf snippet for nginx

vigneshgurusamy:


taylorotwell:
AFAIK nothing is broken in this area.

vigneshgurusamy:
@taylorotwell Sorry my bad. I have not explained about this PR.

In settler, `fastcgi_params` file is updated because of missing fastcgi parameter

```sh
fastcgi_param SCRIPT_FILENAME $request_filename
```

But it has been added in `fastcgi.conf` which is included in this snippet.

If we are using this snippet, **we don't need to update `fastcgi_params`** file in settler.

All homestead Nginx host files are using similar fastcgi configurations which are already present in this snippet. So I thought using this snippet will make it clean & consistent.

vigneshgurusamy:
@taylorotwell ping...

PR #424, opened 2017-01-09
Add MongoDb

studiomado:


sensorario:
+1

jordanramstad:
I would say since laravel does not use mongo in the core it might be best to keep this out of default homestead.

That being said the ability to support it I feel would be good, just that it might be best to somehow do it though add-ons rather then in the main package.

The reason I think this if its supported in the core it needs to be maintained in the core, and its one more application that will need to be tested/checked in every version on every OS/Provider.

Just my thoughts on it anyway.

svpernova09:
Thanks for the PR!

PR #430, opened 2017-01-19
Improve install sh scripts

martinssipenko:
- Touch after install completes, otherwise it might fail, and on next run `vagrant up --provision` it thinks that mariadb was installed
- dump command executions of install scripts
- exit if something fails during install
- init.sh should have `+x`


scripts/install-maria.sh
```diff
@@ -1,14 +1,15 @@
 
 # Check If Maria Has Been Installed
 
+set -e
+set -x
```
andrewhood125: Would love if these were documented. I think all but the most hardcore scripters remember all the flags.

andrewhood125: I'm in favor of leaving off `-x`.

martinssipenko: Possibly, I find it helpful when I can see which command exactly failed, if something fails. But happy to remove it if others think it's too verbose.

andrewhood125: Yeah it's great for debugging, but to the precedent set in other places in homestead is not to output unless there's a failure.

svpernova09:
```
set -e
set -x
```

This isn't something I'd want to do every run. I prefer to have as little red text scrolling past the user as possible. This could be something that's run behind a debug check. You'd have to check a debug flag in the `Homestead.yaml` and I'm not even sure how useful that would be for end users. (We haven't needed such extensive debugging so far, but doesn't mean we would never need it)

scripts/install-maria.sh
```diff
@@ -53,3 +55,5 @@ mysql --user="root" --password="secret" -e "GRANT ALL ON *.* TO 'homestead'@'0.0
 mysql --user="root" --password="secret" -e "GRANT ALL ON *.* TO 'homestead'@'%' IDENTIFIED BY 'secret' WITH GRANT OPTION;"
 mysql --user="root" --password="secret" -e "FLUSH PRIVILEGES;"
 service mysql restart
+
+touch /home/vagrant/.maria
```
svpernova09: This isn't a terrible idea, but we have root access when we're running these, lets drop this somewhere outside of `/home/vagrant`. This would be useful as it's own PR outside of the `set -e` as well.

svpernova09:
Going to close this, but open to another PR that implements ```Touch after install completes, otherwise it might fail, and on next run vagrant up --provision it thinks that mariadb was installed```

martinssipenko:
Ok, will do.
However, I agree with `set -x`, that it would dump a lot of output.
But having script to continue if one of it's commands fails is useless. I've personally had times when installing mariadb fails, and then i need to destroy and try again. without `set -e` moving touch is useless, because it would be executed even if above commands fail.

svpernova09:
```
But having script to continue if one of it's commands fails is useless
```

When I tested your branch, the script exited as you'd expect, but vagrant reported it as a long string (hashed I believe from running it from /vagrant) which was also a useless error to see. I'd rather see the error in the console from vagrant which will clue me into what actually failed

martinssipenko:
Ok, I'll try it out.

svpernova09:
```
I've personally had times when installing mariadb fails, and then i need to destroy and try again. without set -e moving touch is useless, because it would be executed even if above commands fail.
```

I understand, but there's only so many factors Homestead can account for. maybe move the touch to the last thing that happens?

martinssipenko:
@svpernova09 I tested the `install-maria.sh` script from this pull request with inserting `/bin/false` right before installing mariadb (around line 25), and the output of running `vagrant up` was..

```
...
==> homestead-7: Reading package lists...
==> homestead-7: Building dependency tree...
==> homestead-7: Reading state information...
went wrong.
```

Can you give an example to hashes you were mentioning above?

Otherwise, I think that the pull request in the current state would attempt to install maria/mongo if it failed.

PR #432, opened 2017-01-20
Script to allow Homestead to be globally accessible in Windows

axelitus:
Also the init will ask if you want to add Homestead to the PATH environment variable and take care of this for you.

### DIFF
--- a/bin/homestead.bat
+++ b/bin/homestead.bat
@@ -1,0 +1,12 @@
+@echo off
+
+set cwd=%cd%
+set homesteadVagrant=%~dp0
+set homesteadVagrant=%homesteadVagrant:~0,-5%
+
+cd /d %homesteadVagrant% && vagrant %*
+
+cd /d %cwd%
+
+set cwd=
+set homesteadVagrant=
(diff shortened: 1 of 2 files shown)


svpernova09:
Will have a look. Not sure if this is something we should delegate to Homestead, or ensure the user performs this themselves.

andrewhood125:
`.bat` 🤢

andrewhood125:
Maybe it's time we fork homestead into homestead-win..

svpernova09:
No reason to fork. But I would like to move `init.bat` to `init.ps1` And clean up the code. <3 PowerShell

svpernova09:
@axelitus Thanks for the PR. Going to hold off on this because I'd rather not have the Windows version acting differently than the Linux version (if at all possible). I'm also not sure this is something Homestead should be responsible for.

Thanks.

axelitus:
Ok, I agree that it should work likewise in both platforms.

I think a win-specific version of the [Accessing Homestead Globally](https://laravel.com/docs/5.3/homestead#accessing-homestead-globally) doc page is missing though as it could be understood that in windows there is no way to access homestead globally, which is not the case.

svpernova09:
@axelitus Feel free to write something up and submit a PR to docs.

axelitus:
Will do ;)

PR #434, opened 2017-01-25
allow for multiple domain names per config

willgopublic:


taylorotwell:
We don't accept PRs without full descriptions. Please resubmit with a full description and example of the feature.

PR #436, opened 2017-01-26
✨ 🎀 🚀 Add statamic serve script

svpernova09:
To use, adjust the following to your `homestead.yaml`

```
sites:
    - map: statamic.app
      to: /home/vagrant/statamic
      type: statamic
```


svpernova09:
Resolves #435

PR #437, opened 2017-01-26
allow for multiple domain names per config

willgopublic:
This patch allows for multiple domain names to be configured for nginx ssl-servers.

In order to do this, spaces are replaced with underscores for filenames that are generated in ```scripts/serve-laravel.sh```. The spaces are however __not__ replaced for the servername.

An example config in ```Homestead.yaml``` would be:

```yaml

sites:
   - map: example.org www.example.org example.com www.example.com
     to: /vagrant/Code/example-site/public

```

### DIFF
--- a/scripts/serve-laravel.sh
+++ b/scripts/serve-laravel.sh
@@ -2,15 +2,17 @@
 
 mkdir /etc/nginx/ssl 2>/dev/null
 
+safe_1=${1//\ /_}
+
 PATH_SSL="/etc/nginx/ssl"
-PATH_KEY="${PATH_SSL}/${1}.key"
-PATH_CSR="${PATH_SSL}/${1}.csr"
-PATH_CRT="${PATH_SSL}/${1}.crt"
+PATH_KEY="${PATH_SSL}/${safe_1}.key"
+PATH_CSR="${PATH_SSL}/${safe_1}.csr"
+PATH_CRT="${PATH_SSL}/${safe_1}.crt"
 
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for the PR.

I'm not wild about the filenames generated (Concatenation of  all the server names). It also fails if you add too many domain names (Probably due to length).

I think the implementation of this would be better as adding an array of names to the site config. Something along the lines of:

```
sites:
   - map: example.org
     to: /vagrant/Code/example-site/public
      names:
        - www.example.org
        - example.com
        - www.example.com
```

scripts/serve-laravel.sh
```diff
@@ -2,15 +2,17 @@
 
 mkdir /etc/nginx/ssl 2>/dev/null
 
+safe_1=${1//\ /_}
```
andrewhood125: If we're naming this can we use the opportunity to give it a more descriptive name?

willgopublic: Would ```filename_safe_1``` be a more descriptive name?

svpernova09: IMHO the filename should be the primary server name. Especially not all of them since that would allow a user to list to many domains and cause the filename to be too long and fail.

Any other server names shouldn't be appended, they should be explicitly declared via config syntax (see previous comment).

willgopublic:
I thought about just using the first domain name as particle for the names of generated files.

That might however cause collisions.

How would you feel about about a domain names check beforehand and then using the first domain name as leading particle for the filename?

svpernova09:
Shouldn't cause collisions unless the user puts the site in twice, which is possible, but with explicit config syntax, should be an obvious way to spot the issue.

PR #443, opened 2017-02-07
Add the host file update + unit test

ezimuel:
This PR adds the update of the host file during the `make` command. It supports Linux/Mac and Windows.
Moreover, the PR adds the unit test to the project, including a first test on the `make` command.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /vendor
 /.vagrant
+Homestead.yaml
+phpunit.xml
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,9 @@
         "symfony/console": "~2.3|~3.0",
         "symfony/process": "~2.3|~3.0"
     },
+    "require-dev": {
+        "phpunit/phpunit": "~5.7"
+    },
     "license": "MIT",
     "authors": [
         {
@@ -16,6 +19,11 @@
     "autoload": {
         "psr-4": {
             "Laravel\\Homestead\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Tests\\": "tests/"
         }
     },
     "extra": {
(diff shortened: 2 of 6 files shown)


taylorotwell:
I don't want Homestead to be touching the hosts file.

ezimuel:
@taylorotwell why? You don't think people can benefit from this feature?

ptondereau:
@ezimuel I think it's the responsability of end developer to add a helper to its project. Homestead's project should concern only required configuration to run. It's not mandatory to add hosts entries to run it.

ezimuel:
@ptondereau actually is mandatory. The Homestead project works out of the box using the host name `homestead.app`. You must add it in your host file to use it (with default IP address `192.168.10.10`). This is also documented in the documentation [here](https://laravel.com/docs/5.4/homestead#configuring-homestead) as requirement. I only provided an automatic way to add the hostname in the host file, if not present.

That said, it's perfectly fine to reject this PR, but I would like to know why from @taylorotwell. I don't see technical cons about that.

svpernova09:
@ezimuel Thanks for taking the time for the PR.

 IMHO It's not vagrant's responsibility to set host names, and since doing so would require sudo access on *nix like systems, Homestead doesn't want to assume the risk that comes along with that. We also would have to rely on sudo-esque admin privileges in Powershell to accomplish the same thing on Windows. It's just MUCH easier and LESS risk for Homestead to not be responsible for that.

There are Vagrant plugins out there that will manage this for you. We've had a number of attempts to merge this functionality into Homestead and it's something we've always adamantly avoided.

I'd love to see a PR with a test for make command in a separate PR though.

taylorotwell:
I basically agree with @svpernova09. Every time we have tried to manage the hosts file within Homestead it has had to be reverted and I find it simpler to just use Gasmask or even edit the host file manually.

ezimuel:
@taylorotwell and @svpernova09 thanks for you reply. Now, I can see your point but still I think that editing a text file is not a big risk. Maybe, the [vagrant-hostmanager](https://github.com/devopsgroup-io/vagrant-hostmanager) it's a better alternative.
Anyway, I will create a new PR to separate the unit test for the make command, included in this PR.

PR #445, opened 2017-02-07
Added unit tests

ezimuel:
This PR adds the unit tests for `MakeCommand`. This is a first attempt, testing only constructor and `execute()` right now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "symfony/process": "~2.3|~3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~5.7"
+        "phpunit/phpunit": "^5.7 || ^6.0"
     },
     "license": "MIT",
     "authors": [
--- a/composer.json
+++ b/composer.json
@@ -6,6 +6,9 @@
         "symfony/console": "~2.3|~3.0",
         "symfony/process": "~2.3|~3.0"
     },
+    "require-dev": {
+        "phpunit/phpunit": "~5.7"
+    },
     "license": "MIT",
     "authors": [
         {
@@ -16,6 +19,11 @@
     "autoload": {
         "psr-4": {
             "Laravel\\Homestead\\": "src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
+            "Tests\\": "tests/"
         }
     },
     "extra": {
--- a/tests/MakeCommandTest.php
+++ b/tests/MakeCommandTest.php
@@ -1,0 +1,41 @@
+<?php
+namespace Tests;
+
+use PHPUnit_Framework_TestCase as TestCase;
+use Laravel\Homestead\MakeCommand;
+use Symfony\Component\Console\Tester\CommandTester;
+
+class MakeCommandTest extends TestCase
+{
+    protected static $testFolder;
+
+    public static function setUpBeforeClass()
+    {
+        self::$testFolder = sys_get_temp_dir() . DIRECTORY_SEPARATOR . uniqid("homestead_", true);
+        mkdir(self::$testFolder);
+        chdir(self::$testFolder);
+    }
+
+    public static function tearDownAfterClass()
+    {
+        array_map('unlink', glob(self::$testFolder . DIRECTORY_SEPARATOR . '*'));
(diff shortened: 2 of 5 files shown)


ezimuel:
@svpernova09 this is the PR containing only the unit tests from #443.

composer.json
```diff
@@ -6,6 +6,9 @@
         "symfony/console": "~2.3|~3.0",
         "symfony/process": "~2.3|~3.0"
     },
+    "require-dev": {
+        "phpunit/phpunit": "~5.7"
```
weierophinney: Maybe `^5.7 || ^6.0`?

tests/MakeCommandTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+namespace Tests;
+
+use PHPUnit_Framework_TestCase as TestCase;
```
weierophinney: Use `PHPUnit\Framework\TestCase` here; 5.7 adds forwards-compat shims to make that work, and 6.0 only has namespaced classes.

tests/MakeCommandTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+    public function testExecuteMake()
+    {
+        $makeCommand = new MakeCommand();
+
+        $tester = new CommandTester($makeCommand);
+        $tester->execute([], []);
+
+        $this->assertContains('Homestead Installed!', $tester->getDisplay());
+        $this->assertEquals(0, $tester->getStatusCode());
```
andrewhood125: It might be worth while to make a `TestCase` class, for every command we can count on having 
```

$tester = new CommandTester(...);
$tester->execute([], []);
$this->assertEquals(0, $tester->getStatusCode());
```

tests/MakeCommandTest.php
```diff
@@ -0,0 +1,41 @@
...
+        rmdir(self::$testFolder);
+    }
+
+    public function testConstructor()
+    {
+        $makeCommand = new MakeCommand();
+        $this->assertInstanceOf(MakeCommand::class, $makeCommand);
+    }
+
+    public function testExecuteMake()
```
andrewhood125: It's totally subjective but I'd prefer using annotations, especially in the case where we are mixing tests with non test functions. 

```
public function setup()

public function teardown()

/** @test */
public function make_command_is_instantiated_correctly

/** @test */
public function make_command_executes_successfully
```

andrewhood125:
Opinionated comment, but awesome work. I'm just glad we may have some coverage.

tests/MakeCommandTest.php
ezimuel: @andrewhood125 actually we have only the `make` command to test here. So the class `MakeCommandTest` is testing already a single command. I will rename the `testExecuteMake` function in `testExecute` to clarify this point.

tests/MakeCommandTest.php
ezimuel: @andrewhood125 updated with `@test` annotation. I read the [contribution guide](https://laravel.com/docs/5.3/contributions) of Laravel but I didn't find any notes about unit testing.

composer.json
ezimuel: Changed.

tests/MakeCommandTest.php
ezimuel: Changed.

tests/MakeCommandTest.php
andrewhood125: Ah yeah you're right. I forget they all got removed.

svpernova09:
@ezimuel 

This looks good, my only request is you make the changes that StyleCI is suggesting: https://styleci.io/analyses/87LPeZ

Good to go after that.

ezimuel:
@svpernova09 I was assuming [this](https://laravel.com/docs/5.3/contributions#styleci) from the contribution guide of Laravel. This is why I didn't fix. Should I do then?

svpernova09:
@ezimuel This is my first time disobeying Style CI, Let's see what happens!

svpernova09:
@ezimuel Thanks for the PR! And thanks @mwop for the feedback :D

PR #461, opened 2017-02-16
added exist checking for folder mapping

bassauer-storms-media:
Added exist checking for the mapping of local folders for the homestead VM.
In special cases (like mine) it may appear that one single folder to map does not exist.
So that one line prevents homestead from crashing in case of a missing folder.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -132,25 +132,27 @@ class Homestead
     # Register All Of The Configured Shared Folders
     if settings.include? 'folders'
       settings["folders"].each do |folder|
-        mount_opts = []
+        if File.exists? File.expand_path(folder["map"])
+          mount_opts = []
 
-        if (folder["type"] == "nfs")
-            mount_opts = folder["mount_options"] ? folder["mount_options"] : ['actimeo=1', 'nolock']
-        elsif (folder["type"] == "smb")
-            mount_opts = folder["mount_options"] ? folder["mount_options"] : ['vers=3.02', 'mfsymlinks']
-        end
+          if (folder["type"] == "nfs")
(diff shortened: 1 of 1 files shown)


svpernova09:
So this is in the case that the folder on the host being mapped doesn't exist? That would indeed cause a hard Vagrant error.

The else of this `if` should be an error message and an exit saying something to the effect of "Mapped folder {$folder} not found on the host, please check your Homestead.yaml folders"

bassauer-storms-media:
well the reason I don't want homestead to crash if the folder does not exist is that we got windows and mac computers here in our company that use one and the same homestead.yaml file (so that not every single person and also designers and managers etc have to fiddle with vhost configs), loaded from a network drive. A mac cannot mount X:\... and a windows OS cannout mount a NFS like /Volumes/...  
So we thought about putting both folders to the yaml and homestead just skips them if the dir does not exist. If you think this is nothing generic useful, we will just modify out own homestead.rb and you can deny that pull request. A error message would be finde but not some sort of kernel panic

svpernova09:
@bassauer-storms-media I'm agreeing with you that a hard crash isn't what we want. But I also don't want a silent fail since that would be confusing to the normal user who may have just mistyped something in their config.

bassauer-storms-media:
a big yelling error notice would be fine as long as it won't interrupt/cancel the vm's start up process. Feel free to add such a message in the else branch :)

svpernova09:
Yeah, I can see where exiting on that could cause issues with VM state, so probably better we continue with the process.

Add that else and I'll get it tested and merged.

bassauer-storms-media:
Okay, but this has to wait a few hours; I don't know how to abandon/cancel a pull request and I dont know how to print error messages within the ruby script. I have to check that first. Thanks for your time :)

svpernova09:
No worries, come back to it when you can.

PR #464, opened 2017-02-16
added checking for none existing local dirs while mapping

bassauer-storms-media:
see https://github.com/laravel/homestead/pull/461

now there is a error message instead of silent skipping


svpernova09:
Change your else to this:

```
          config.vm.provision "shell" do |s|
            s.inline = ">&2 echo \"Unable to mount one of your folders. Please check your folders in Homestead.yaml\""
          end
```

That way it will show up in red in the console.

PR #468, opened 2017-02-16
🔥 Remove HHVM

svpernova09:


runciters:
Why did you remove support for HHVM?

svpernova09:
@runciters 

Primarily this: https://twitter.com/grahamjcampbell/status/760485090212315136

I'd rather drop HHVM support in a known working / known good state than wait until it's broken and someone has to find out the hard way it's broken / not working. We can safely say the last version of Homestead to support HHVM will be version 4.x. This allows those HHVM users to update their systems/projects to lock in at Homestead 4.x so they don't lose any features once Homestead 5.x is released.

runciters:
@svpernova09 Thanks, this actually makes sense!

howdoicomputer:
Can the Homestead documentation be updated to not list HHVM as an option? https://laravel.com/docs/5.2/homestead

svpernova09:
@howdoicomputer This is somewhat tricky because if you use an older version of Homestead, HHVM *will* work. So it's hard to say where the line is in terms of where we should document (or remove documentation in this case) these sort of changes.

howdoicomputer:
@svpernova09 Oh, wait, my apologies! I linked the 5.2 docs. The 5.4 docs are listed in the main documentation page and they don't list HHVM as an option - I was looking at old docs. My fault. Smaller suggestion: maybe a warning flag for older doc pages? Something like this: https://guides.emberjs.com/v2.11.0/

svpernova09:
HHVM isn't listed in the 5.4 docs because Laravel 5.4 doesn't support HHVM.

PR #472, opened 2017-02-17
Refactor/files

DojoGeekRA:
Refactor some files to achieve the following goals:

* All files must have the newline character at the end of the file
* All shell scripts have the shebang at the beginning of the file
* An indentation of 4 spaces is used in all the files

Also it adds an `.editorconfig` that helps to ensure some of that goals.


svpernova09:
Thanks for taking the time to submit a PR.

It's not Ruby convention to have 4 space tabs, they should be 2 space tabs in Ruby files. Any *.rb files and Vagrant files should conform to Ruby syntax/style.

I *love* EditorConfig, but I don't want to add it unless the other Laravel projects are also adding it (to insure compliance/compatibility)

I'm fine with shell scripts having 4 space indentations. Can't say I really care about trailing newlines outside of PHP files.

DojoGeekRA:
@svpernova09 thanks for the feedback!

I didn't knew about the Ruby indentation convention and it's great, today i learned something new 🤘 . I made that change because the Vagrant files already have 4 spaces so that was my reference.

I agree with you about the `.editorconfig`, i knew it was something risky to propose but i wanted to take the shoot.

About the shell scripts i did that because some of them already have 4 spaces, it's great to know that something like that is acceptable.

And finally the newline characters was because in the recent PRs that character has been added and deleted constantly, i though that one convention could help to keep the diffs clean.

P.S. What were your thoughts about the `shebang` in the shell scripts?

Anyways, again, thanks for the feedback and if you're interested i can send another PR with:

* Ruby and Vagrant files with only 2 spaces as indentation
* Shell scripts with 4 spaces as indentation

svpernova09:
@DojoGeekRA 

Yeah, if you want to format the Ruby & Shell scripts to use the right indention that'd be fine.

As for the scripts missing shebang lines, I think these are all being called from other shell scripts, and I have no real opinion about them having shebang lines added.

PR #474, opened 2017-02-21
Fix indentation in ruby files

DojoGeekRA:
It's not Ruby convention to have 4 space tabs, they should be 2 space tabs in Ruby files. Any *.rb files and Vagrant files should conform to Ruby syntax/style.


taylorotwell:
I found 4 more readable.

DojoGeekRA:
@taylorotwell i actually sent a PR before with 4 spaces but was rejected and asked to use 2 spaces in #472, but in any case, these file have both (4 and 2 spaces in the same file), will be great to have just 1 kind of indentation. Should i leave the files as they are or should i send again a PR with 4 spaces?

PR #482, opened 2017-02-22
Add CAcert generation & signing for serve-scripts

mroedel:
Get rid of that annoying SSL Security warnings...

The creation of a CAcert allows you to trust a single certificate and get every host trusted.

### DIFF
--- a/scripts/serve-apache.sh
+++ b/scripts/serve-apache.sh
@@ -4,16 +4,28 @@ apt-get update
 apt-get install -y apache2 libapache2-mod-php7.1
 sed -i "s/www-data/vagrant/" /etc/apache2/envvars
 
+PATH_CA="/vagrant"
+PATH_CA_KEY="${PATH_CA}/ca.key"
+PATH_CA_CRT="${PATH_CA}/ca.crt"
+
+if [ ! -f $PATH_CA_KEY ] || [ ! -f $PATH_CA_CRT ]
+then
+    openssl genrsa -out "$PATH_CA_KEY" 2048 2>/dev/null
+    openssl req -new -x509 -days 365 -key "$PATH_CA_KEY" -out "$PATH_CA_CRT" -subj "/CN=homestead/O=Vagrant/C=UK" 2>/dev/null
+fi
+
 PATH_SSL="/etc/apache2/ssl"
 PATH_KEY="${PATH_SSL}/${1}.key"
 PATH_CSR="${PATH_SSL}/${1}.csr"
 PATH_CRT="${PATH_SSL}/${1}.crt"
+PATH_BUNDLE="${PATH_SSL}/${1}-bundle.crt"
 
-if [ ! -f $PATH_KEY ] || [ ! -f $PATH_CSR ] || [ ! -f $PATH_CRT ]
+if [ ! -f $PATH_KEY ] || [ ! -f $PATH_CSR ] || [ ! -f $PATH_CRT ] || [ ! -f $PATH_BUNDLE ]
 then
-    openssl genrsa -out "$PATH_KEY" 2048 2>/dev/null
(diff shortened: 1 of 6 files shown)


svpernova09:
Testing a laravel project this seems to make no difference.  The output of the `serve-laravel.sh` script looks good, but Chrome, Safari, and Firefox throw `SEC_ERROR_UNKNOWN_ISSUER` errors.

```
root@homestead:/etc/nginx/ssl# ll
total 40
drwxr-xr-x 2 root root 4096 Feb 23 01:25 ./
drwxr-xr-x 9 root root 4096 Feb 23 01:25 ../
-rw-r--r-- 1 root root 2238 Feb 23 01:25 quickstart.app-bundle.crt
-rw-r--r-- 1 root root 1058 Feb 23 01:25 quickstart.app.crt
-rw-r--r-- 1 root root  940 Feb 23 01:25 quickstart.app.csr
-rw-r--r-- 1 root root 1679 Feb 23 01:25 quickstart.app.key
-rw-r--r-- 1 root root 2230 Feb 23 01:25 rfid.app-bundle.crt
-rw-r--r-- 1 root root 1050 Feb 23 01:25 rfid.app.crt
-rw-r--r-- 1 root root  932 Feb 23 01:25 rfid.app.csr
-rw-r--r-- 1 root root 1679 Feb 23 01:25 rfid.app.key
```

Contents of `/etc/nginx/sites-available/rfid.app`

```
server {
    listen 80;
    listen 443 ssl http2;
    server_name rfid.app;
    root "/home/vagrant/rfid/public";
    index index.html index.htm index.php;
    charset utf-8;
    location / {
        try_files $uri $uri/ /index.php?$query_string;
    }
    location = /favicon.ico { access_log off; log_not_found off; }
    location = /robots.txt  { access_log off; log_not_found off; }
    access_log off;
    error_log  /var/log/nginx/rfid.app-error.log error;
    sendfile off;
    client_max_body_size 100m;
    location ~ \.php$ {
        fastcgi_split_path_info ^(.+\.php)(/.+)$;
        fastcgi_pass unix:/var/run/php/php7.1-fpm.sock;
        fastcgi_index index.php;
        include fastcgi_params;
        fastcgi_param SCRIPT_FILENAME $document_root$fastcgi_script_name;
        fastcgi_intercept_errors off;
        fastcgi_buffer_size 16k;
        fastcgi_buffers 4 16k;
        fastcgi_connect_timeout 300;
        fastcgi_send_timeout 300;
        fastcgi_read_timeout 300;
    }
    location ~ /\.ht {
        deny all;
    }
    ssl_certificate     /etc/nginx/ssl/rfid.app-bundle.crt;
    ssl_certificate_key /etc/nginx/ssl/rfid.app.key;
}
```

svpernova09:
If the idea is that we would trust *this* CA in our browsers I would not want to encourage Homestead users to trust *any* CA that they create themselves.

svpernova09:
My other concern would be that every time you add a new site you would have to check for existing cert, use it with the vhost. You'd also have to re-trust the cert every time you destroy and up. (I fully admit I'm likely an edge case as someone who destroys very often)

I've pinged a few security people that I respect and will see what they have to say about the idea.

svpernova09:
Going to hold off on this for now. Thanks for the idea. While I'm still torn on the issue you should be able to easily accomplish this by adding it to your `after.sh` file.

mroedel:
In my opinion it doesn't makes a difference which error clicked away when the CA is not trusted.

Trusting that CA is save, as long users won't re-distribute their CA to others.

You "never" have to re-trust the CA, because the CA is stored in `/vagrant` (synced to your host-folder). When you refresh your Box the "old" CAcert will be used every time as long you don't delete it.

PR #488, opened 2017-02-24
Terminate execution if settings file doesn't exists

DojoGeekRA:
Updates `LocalizedVagrantfile` to behave like the `Vagranfile`.

See #484


svpernova09:
Thanks! Good catch. I totally forgot about `LocalizedVagrantfile`

PR #489, opened 2017-02-25
✨ 💡 ✅ Add --json flag for Make Command. Add test coverage for make command

svpernova09:
Freely admitting that I took the easy way out on supporting JSON flag here.

Generate Yaml as usual -> Did the user ask for JSON? (--json) -> Convert Yaml to Json!

Added test coverage by verifying files are created in the test directory as expected and ensuring that values for name, hostname, and IP are properly set.

/cc @DojoGeekRA

PR #492, opened 2017-02-26
Refactor MakeCommand

DojoGeekRA:
@svpernova09 sorry for the delay, i see that you already merged #489 but like you mentioned, there was some technical debt in there, so instead of just patching the command i heavily refactored it.

In this new version the following goals are reached:

* Improved the readability of the `execute` method extracting implementation details into methods.
* The make command no longer deals with the `json` or `yaml` create or update tasks, it now delegates that to the `JsonSettings` and `YamlSettings` objects through strategy pattern.
* The make command is now 100% test covered 🤘 (thank you very much @ezimuel for your awesome test `setUp` and `tearDown` features, without them achieve this one would have been very hard). The following scenarios are now tested:

  * A `Vagrantfile` is created if not already exists
  * An existing `Vagranfile` is not overwritten
  * An `alises` file is created if requested
  * An existing `aliases` file is not overwritten
  * An `after.sh` file is created if requested
  * An existing `after.sh` is not overwritten
  * An `Homestead.{yaml|json}.example` is created if requested
  * An existing `Homestead.{yaml|json}.example` is not overwritten
  * A `Homestead.{yaml|json}` is created if not already exists
    * It can be created in `yaml` or in `json` format.
    * If the `Homestead.{yaml|json}.example` exists, `the Homestead.{yaml|json}` is created from it
    * If the `Homestead.{yaml|json}.example` doesn't exists, `the Homestead.{yaml|json}` is created from the stub.
  * A `Homestead.{yaml|json}` can be overwritten with the command options: `name`, `hostname` and `ip`

The new features are:
  * The `Homestead.{yaml|json}.example` now can be updated with the same command options as the `Homestead.{yaml|json}` (`name`, `hostname` and `ip`)
  * If an `Homestead.{yaml|json}.example` is requested it is created, but also it will create the `Homestead.{yaml|json}` with exactly the same content.

The missing features (work in progress):
  * Implement the previous `configurePaths` method to update the `folders` and `sites`
  * I'm not 100% sure, but seems [Symfony Yaml component](like http://symfony.com/doc/current/components/yaml.html) doesn't require the php-yaml extension so this can solve the TravisCI problem.
  * Beautify the code. I have been drafting and iterating over this a LOT, i just have to make sure that everything is consistent and properly arranged.

### DIFF
--- a/src/MakeCommand.php
+++ b/src/MakeCommand.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Homestead;
 
+use Laravel\Homestead\Settings\JsonSettings;
+use Laravel\Homestead\Settings\YamlSettings;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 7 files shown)


DojoGeekRA:
@svpernova09 by the way, given that TravisCI is not working right now, this are the test right now 😄 

captura de pantalla 2017-02-26 a la s 14 51 54

svpernova09:
Looks good, resolve the conflicts and we're good to go

DojoGeekRA:
@svpernova09 fixed!

It also has a new test fot the `checkForDuplicateConfigs` that you created.

svpernova09:
Thanks for the work getting this done. Much better than my hack :D

PR #496, opened 2017-02-27
use Symfony yaml component instead of yaml PHP extension

DojoGeekRA:
It now uses `Symfony yaml` component to handle `yaml` parsing.

This have also has the advantage that the developers doesn't need the `php-yaml` extension in their systems.


DojoGeekRA:
@svpernova09 this fixes the TravisCI builds.

PR #498, opened 2017-02-28
Re-implement the nginx sites and shared folders settings

DojoGeekRA:
@svpernova09 this PR re-implement the following features that were missing during the refactor of the make command:

* Configure nginx sites in settings file
* Configure shared folders in settings file
* Improve readability of the code

These features are of course now covered with test and with this, the make command is fully refactored.


svpernova09:
Wow. I must have not checked out a clean copy when I was testing this because mine worked fine. Thanks for catching this.

PR #499, opened 2017-03-01
Allow user to specify their own options for ngrok

vblinden:
Hi guys,

This allows the user to specify the region (-region=eu) or subdomain (-subdomain=test1234). This may come in handy for people from europe.

Thanks.

### DIFF
--- a/src/stubs/aliases
+++ b/src/stubs/aliases
@@ -96,5 +96,7 @@ function share() {
         echo "Error: missing required parameters."
         echo "Usage: "
         echo "  share domain"
+        echo "Invocation with extra params passed directly to ngrok" 
+        echo "  share domain -region=eu -subdomain=test1234"
     fi
 }
--- a/src/stubs/aliases
+++ b/src/stubs/aliases
@@ -91,7 +91,7 @@ function serve-symfony2() {
 function share() {
     if [[ "$1" ]]
     then
-        ngrok http -host-header="$1" 80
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for the PR, will have a look at this later today.

andrewhood125:
Can we add something in the usage for this feature?

svpernova09:
Docs should be supplemented here https://laravel.com/docs/5.4/homestead#sharing-your-environment

vblinden:
Sure, I could add a note in the usage. Should I include every ngrok command or just a note that you can add ngrok commands after the first command?

And also a section in the documentation?

andrewhood125:
I had to look up `${@:2}`. I was thinking something along the lines of:

```
function share() {
    if [[ "$1" ]]
    then
        ngrok http ${@:2} -host-header="$1" 80
    else
        echo "Usage: "
        echo "  share domain "
        echo "Invocation with extra params passed directly to ngrok" 
        echo "  share domain -region=eu -subdomain=test1234"
    fi
}
```

Is that how it would be used @vblinden?

vblinden:
@andrewhood125 That is exactly how this would be used. Shall I add this to the pull request?

andrewhood125:
That would be my 2 cents. Because it wasn't immediately obvious to me what the ${@:2} did but I'll defer to others.

PR #500, opened 2017-03-02
Support of older PHP versions

lachezargrigorov:
I have 100+ project from Laravel 3 to Laravel 5.4  in my homestead local environment  and some of them, the older one needs PHP 5.6 to work because of a lot of deprecated functions, mcrypt and other staffs in PHP 7. Sometimes I need to have the same PHP version on my local environment and the production server that is provided by the firm client and I have no rights to change the PHP version and so on. So I think it is a good idea to have more then one PHP version in your box and to have the ability to change the PHP version for some of the sites in  Homestead.yaml if needed.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -178,7 +178,7 @@ class Homestead
                 config.vm.provision "shell" do |s|
                     s.name = "Creating Site: " + site["map"]
                     s.path = scriptDir + "/serve-#{type}.sh"
-                    s.args = [site["map"], site["to"], site["port"] ||= "80", site["ssl"] ||= "443"]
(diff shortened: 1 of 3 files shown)


jonnywilliamson:
But php5.6 isn't installed in the latest settler scripts. So this could be very confusing if a user added the flag to his config yaml file but php5.6 isn't there!

https://github.com/laravel/settler/blob/master/scripts/provision.sh#L52-L57

svpernova09:
Thanks for taking the time for the PR. Going to pass on this for now because I'd rather not support more than 1 version of PHP in a specific version.

My recommendation is to create a box for each version you need, (5.5 that you can upgrade to 5.6, 7.0, and 7.1). You can follow the instructions here: https://laravel-news.com/using-older-versions-of-homestead to match versions of the Homestead repo to use with versions of the box. (For 5.5/5.6 you'll have to use a different box)

PR #501, opened 2017-03-05
Optimize TravisCI

DojoGeekRA:
* Removes the php-yaml installation

   *It is no longer needed given that the yaml file manipulations is being done with Symfony Yaml component.*

* Disable xdebug during TravisCI execution

    *To improve the builds execution time*


svpernova09:
Thanks for the PR!

PR #502, opened 2017-03-05
add flip script and alias

browner12:
the `flip` command allows the users to easily switch between the Nginx
and Apache web servers on their homestead boxes.

it does this by checking if apache is running, and then starting and
stopping the services depending on which is currently running.

### DIFF
--- a/scripts/flip-webserver.sh
+++ b/scripts/flip-webserver.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+ps auxw | grep apache2 | grep -v grep > /dev/null
+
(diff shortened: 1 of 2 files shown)


svpernova09:
Thanks for the PR!

browner12:
glad to help. Thanks for all the work you do on this project as well @svpernova09 !

I'd like to add a note to the docs on this. Any thoughts on a good place to put it?

svpernova09:
You can skip adding it to the docs since the docs don't mention Apache.

browner12:
sounds good. thanks!

PR #503, opened 2017-03-06
Apache reload

browner12:
if we try to run `service apache2 reload` with apache not running, we get the fatal error ‘apache2.service is not active, cannot reload.’ and it kills the rest of the provisioning script.

this update first checks if apache is running. if it is, we reload like normal.  if it isn’t, we first turn off nginx, then start apache, then perform the reload, and finally turn nginx back on.

not sure if this is the best way to do this, or if we should just ignore the reload if apache isn't started. open to suggestions.

### DIFF
--- a/scripts/serve-apache.sh
+++ b/scripts/serve-apache.sh
@@ -156,4 +156,16 @@ then
 fi
 
 a2dissite 000-default
-service apache2 reload
+
+ps auxw | grep apache2 | grep -v grep > /dev/null
+
+if [ $? == 0 ]
+then
+    service apache2 reload
+else
+    service nginx stop
+    service apache2 start
+    service apache2 reload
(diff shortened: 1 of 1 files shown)


svpernova09:
Installing Apache will start the service, so this should always be a reload. Also there is no reason to do a reload after starting Apache since it wasn't running, it parsed all the config files at start.

browner12:
gotcha, makes sense.

the first time an Apache site is installed, this works fine, because Apache is installed and loaded. However, for future provisions this fails because Apache does not need to be installed, and isn't active when the reload command is called.

I will update the PR to simply skip the reload if Apache isn't active.  Then when the user decides to turn Apache on, the reload will essentially take effect, as you mentioned.

thanks,

PR #504, opened 2017-03-06
only reload apache if it is active

browner12:
the command `service apache2 reload` will cause a fatal error if apache
is not loaded.

this update first checks to see if apache is active, before triggering
the reload


browner12:
I think this should address your points from the last PR.  if nginx is loaded, we will simply skip the apache reload.  then, when the user decides to activate apache, the files will be loaded.

PR #506, opened 2017-03-07
Fix "Please use --connect-expired-password option" issue by resetting DB password first.

kevinsmith:
Fixes this error that causes the `vagrant up` command to exit:
```
==> default: Please use --connect-expired-password option or invoke mysql in interactive mode.
went wrong.
```

The origin of this issue seems to have been uncovered in [this Stack Overflow answer](http://stackoverflow.com/a/40497052/1217620).

### DIFF
--- a/scripts/create-mysql.sh
+++ b/scripts/create-mysql.sh
@@ -2,4 +2,6 @@
 
 DB=$1;
 
+mysql -uhomestead -psecret --connect-expired-password -e "SET password=PASSWORD('secret')";
+
 mysql -uhomestead -psecret -e "CREATE DATABASE IF NOT EXISTS \`$DB\` DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_unicode_ci";


andrewhood125:
What version box are you using? I remember some PRs about this vaguely. I think it was fixed in laravel/settler. _searches_ Specifically here: https://github.com/laravel/settler/pull/60 

Updating your box and re-provisioning should fix the issue. I think there was some discussion about adding what your PR does but was decided against because it was an edge case.

kevinsmith:
Using box 0.3.3 since this project cannot yet move to PHP 7.

svpernova09:
Please see the instructions here for using PHP 5.x with Homestead: https://laravel-news.com/using-older-versions-of-homestead

kevinsmith:
Thanks for the tip, @svpernova09!

I'd humbly suggest these instructions should be in a more prominent place, perhaps even the 5.1 documentation. And the legacy Homestead should be officially maintained alongside the LTS version of Laravel that it pairs with. The project that I'm working with here specifically chose Laravel 5.1 because of the long-term support, but that's increasingly less valuable if the surrounding Laravel ecosystem rushes down the road without it.

/cc @taylorotwell

svpernova09:
@kevinsmith 

Laravel 5.1 supporting 5.5 was exactly why I created the new version of that box. That way you could still have an easily supported Homestead environment from the official source without having to hack it yourself.

Because of how Atlas handles box versioning I can't go back and release a new version of and old version number and since Homestead officially doesn't support it, it's under my name on Atlas rather than the Laravel name.

If you have any issues I'll do my best to help.

kevinsmith:
Appreciate the help here, Joe! Hoping perhaps @taylorotwell sees fit to spin up an official `laravel/homestead-legacy` box on Atlas. That would go a long way in signaling a commitment to LTS support.

kevinsmith:
@svpernova09 One issue I'm running into on the legacy box: When I try to connect to the DB from outside the box (using SequelPro in this case), I get this error

> MySQL said: Lost connection to MySQL server at 'reading initial communication packet', system error: 0

Didn't have a problem connecting to the DB when using the 0.3.3 version of the official box. What's changed between that and your legacy box?

svpernova09:
Are you using SSH to connect to vagrant or connecting via localhost on port 33060? I'd try localhost first.

kevinsmith:
Using localhost on port 33060.

svpernova09:
Check some of the replies here: http://stackoverflow.com/questions/5755819/lost-connection-to-mysql-server-at-reading-initial-communication-packet-syste

If it's still an issue I can look into it later tonight

kevinsmith:
I think I see why. It's picking up it's MySQL config from `/etc/mysql/mysql.conf.d/mysqld.cnf` which has `bind-address	= 127.0.0.1`. Looks like the provisioning script isn't properly setting up remote MySQL access.

I'm happy to do a pull request to fix it, just not sure which repo and branch you're using to build the legacy box.

svpernova09:
Branch and PR to https://github.com/laravel/settler/tree/3.0 Thanks!

PR #518, opened 2017-03-13
Move stubs directory to the project root

DojoGeekRA:
What do you think @svpernova09?

This allows to keep only php files in the src/ directory and actually
it is more friendly for people looking for the stubs files to find
them in the project root.


svpernova09:
I *think* we may want to move `stubs` to the root, and then rename `stubs` -> `resources` considering these are non-php template style files. @pmjones recently did a ton of research about project structure (https://github.com/php-pds/skeleton) and I've pinged him on twitter (and now here) for a second opinion.

pmjones:
From the pds/skeleton research, `resources/` seems a perfectly reasonable location. (You might want `resources/stubs/` instead, so you can keep them separate from other kinds of resources later.)

svpernova09:
@DojoGeekRA I think `/resources` is a fine option for now, if we get into more complex stubs later we can break it down further.

Thanks for the PR!

PR #520, opened 2017-03-14
Hotfix - changed directories in init script from src/stub/ to resources/

Pancrisp:


svpernova09:
Thanks for the PR!

DojoGeekRA:
@svpernova09  I'm sorry for missing that ones, seems that we need tests for the shell scripts 😅

Thanks @Pancrisp!

svpernova09:
@DojoGeekRA No worries at all. I intentionally haven't tagged another release until I can sit down and test these changes.

I've been rolling around the idea of some kind of test suite like that but I just dunno.

DojoGeekRA:
@svpernova09 great to know, thanks for that!

And yup, neither do i know how to test those, buts seems like fun to figure it out.

svpernova09:
I think we should try something like this: https://github.com/laravel/homestead/blob/master/tests/Settings/YamlSettingsTest.php#L18-L20

Create a test folder and run the actual commands and inspect file output. Might be worth looking into.

DojoGeekRA:
@svpernova09 looks good, let's give it a shoot. I will tackle this tonight.

PR #522, opened 2017-03-14
update init.bat

Ken-vdE:
the given 'src/stubs' folder does not exist, the files are present int the 'resources' folder apparently.


svpernova09:
Resolved via #520 Thanks for taking the time for the this

PR #525, opened 2017-03-15
Add tests for shell script

DojoGeekRA:
@svpernova09 as discussed in #520, this PR add tests for the `init.sh` shell script.

It also makes a little refactor to improve code readability.

### DIFF
--- a/src/MakeCommand.php
+++ b/src/MakeCommand.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Homestead;
 
+use Laravel\Homestead\Traits\Slug;
 use Laravel\Homestead\Settings\JsonSettings;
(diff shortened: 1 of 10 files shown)


taylorotwell:
Shouldn't have a trait just called "Slug".. it should be something like `GeneratesSlugs`, etc.

svpernova09:
`GenerateSlug`  would be a good name for it.


Thanks @DojoGeekRA I'll have a look at this later this afternoon.

DojoGeekRA:
@taylorotwell you're right, that name makes more sense. It's fixed now.

And thanks @svpernova09 , let me know if you need anything else.

svpernova09:
Thanks for the PR!

Issue #526, opened 2017-03-22
Browser Certificate Warnings


refringe:
The latest beta version of Chrome (and apparently Firefox) make a change to how self-signed certificates are validated:  
https://www.chromestatus.com/feature/4981025180483584

Open any of your Homestead projects in the beta version of Chrome (58.0.3029.19) right now and (even with the SSL certificate imported into the system keychain and explicitly allowed) you will receive a `net::ERR_CERT_COMMON_NAME_INVALID` certificate error page.

More information about why *that* error is shown here:
https://bugs.chromium.org/p/chromium/issues/detail?id=308330

In short, it's no longer good enough to only use the commonName (CN) field for the domain(s) when generating a certificate. The subjectAltName must be used to list domains that the certificate validates. This would typically involve modifying the OpenSSL configuration file for each domain/certificate that you want to generate. 

How to do that properly is written about here:
http://stackoverflow.com/questions/21488845/how-can-i-generate-a-self-signed-certificate-with-subjectaltname-using-openssl

svpernova09:
I'd rather not add multiple domains to the cert since I'd rather avoid such practices (even in dev) .

Will look into this a bit. I'm out of town until Thursday if someone wants to take a crack at this.

--
- Joe Ferguson
JoeFerguson.me
MemphisPHP.org

refringe:
Understood. I've modified the initial issue and removed the suggestion. Practices aside, it would make detecting previously generated certificates (for regeneration purposes) a pain. More simple to keep them separate. 👍

refringe:
I'll see if I can take a swing at it tonight.

ryancwalsh:
@svpernova09 and @refringe I wonder if you know how to solve this Laravel Homestead local dev HTTPS issue: https://stackoverflow.com/q/48969083/470749 Thanks so much for your help!

browner12:
we now generate a root certificate when the machine is built.  you can find it in `/etc/nginx/ssl`.  Simply copy it onto your host machine, and then make it a trusted certificate.  You can find instructions for your specific OS online.

ryancwalsh:
@browner12 Thanks but I already tried that. (I read about it here: https://stackoverflow.com/a/47755133/470749)

ryancwalsh:
[...]

PR #527, opened 2017-03-23
Updates SSL Certificate Generation

refringe:
This changes a few things regarding SSL certificates and how they’re generated. First, the code to generate a certificate has been removed from each of the `serve-[whatever]` scripts and into it’s own script. The certificate generation code was duplicated 6 times, and in one instance (Symfony2) the checks to see whether or not a certificate was already created weren’t included, like they were in the others. So this change should help prevent things like that from happening again.

Second big change is to the way the certificates are actually generated with OpenSSL. We’re now using an OpenSSL configuration file for each certificate that’s generated. They’re saved along-side the certificates. We’re doing this because the certificates must be generated using the `subjectAltName` property to list domains, and that property can only be added via configuration file. The `subjectAltName` property is the primary property that is used in validating domains. Historically, the CN (commonName) value was used to validate domains (as a fall-back), but apparently this was never to specification and recent browsers are removing that decade long oversight(?)

A small change worth mentioning is the removal of the CSR (Certificate Signing Request). Since these are self-signed certificates, we don’t need to send a request to an authority to have the certificates created. We can just skip that step and create the certificate using the key.

Resolves issue #526

### DIFF
--- a/scripts/create-certificate.sh
+++ b/scripts/create-certificate.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+mkdir /etc/nginx/ssl 2>/dev/null
+
+PATH_SSL="/etc/nginx/ssl"
+PATH_CNF="${PATH_SSL}/${1}.cnf"
+PATH_KEY="${PATH_SSL}/${1}.key"
+PATH_CRT="${PATH_SSL}/${1}.crt"
+
+# Only generate a certificate if there isn't one already there.
+if [ ! -f $PATH_CNF ] || ! -f $PATH_KEY ] || [ ! -f $PATH_CRT ]
+then
+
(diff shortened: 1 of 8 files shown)


refringe:
Should have noted the new results...

After copying the certificate to your local machine and adding it to your system (or login) keychain:

    sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain example.app.cer

Chrome v58.0.3029.33 beta starts giving you that wonderful green feeling.

screen shot 2017-03-22 at 11 22 58 pm

svpernova09:
I'm traveling tomorrow but will have a look during my layover

--
- Joe Ferguson
JoeFerguson.me
MemphisPHP.org

svpernova09:
Thanks for taking the time to share this PR!

refringe:
Happy to help!

Issue #516, opened 2017-03-11
Add ChromeDriver


driesvints:
Can we natively add the ChromeDriver in Homestead so we can run Laravel Dusk tests? 

https://laravel.com/docs/5.4/dusk
https://sites.google.com/a/chromium.org/chromedriver/home

driesvints:
I managed to get it to work with these install commands:

```
wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list'
sudo apt-get update && sudo apt-get install -y google-chrome-stable
sudo apt-get install -y xvfb
```

More info here: https://github.com/laravel/dusk/issues/50#issuecomment-275155974

driesvints:
This is really important: (from the comment above) you **need** to run `Xvfb :0 -screen 0 1280x960x24 &` before attempting to run Laravel dusk or it will timeout. Perhaps it could be nice if we could run it when the machine starts up so the process is always running?

When you're on your local machine you don't need to do this because Chrome probably is already running. But on Homestead you'll need to explicitly tell the chrome driver to start.

svpernova09:
I'm all for it, just haven't had time to get around to doing it. I'd like to add it to settler and then have an alias in Homestead to run the `Xvfb :0 -screen 0 1280x960x24 &`, If you're not using Dusk I wouldn't want that process running, so I think manually starting it is a better way to go.

driesvints:
Yeah, agree. Just wish it wasn't so cumbersome. Currently trying to find a way to combine it in one command but no luck so far.

driesvints:
Made this command real quick:

```bash
#!/bin/sh

pids=$(pidof /usr/bin/Xvfb)
if [ ! -n "$pids" ]; then
    Xvfb :0 -screen 0 1280x960x24 &
fi

php artisan dusk
```

svpernova09:
@driesvints I'm going to get this added to Settler soon™ with a next box version due mid April. (trying not to push those out more than once a month if I can help it)

driesvints:
@svpernova09 ok great! Definitely not urgent.

svpernova09:
[...]

PR #528, opened 2017-03-25
add headless selenium setup for homestead #516

syntacticNaCl:
I stumbled upon #516 when I was trying to get Dusk to work with Homestead and decided to take a whack at it. This pull request adds a headless Selenium setup for Homestead using Chrome and Xvfb. One of the issues brought up was that we don't want the framebuffer process running if the user is not using Dusk. So, I have added a Selenium config to the Homestead.yaml file which allows the user to enable or disable the process on startup.

### DIFF
--- a/resources/Homestead.yaml
+++ b/resources/Homestead.yaml
@@ -32,3 +32,6 @@ databases:
 #     - send: 7777
 #       to: 777
 #       protocol: udp
+
+# selenium: 
+#       - start-driver: 'yes'
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -293,5 +293,24 @@ class Homestead
                 ]
             end
         end
+
+        # Configure Headless Selenium
+        if settings.has_key?("selenium")
+
+            config.vm.provision "shell" do |s|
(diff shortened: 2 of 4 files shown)


svpernova09:
Thanks for this. We should do the heavy lifting of installing and configuring on settler: GitHub.com/laravel/settler

I'll have a look at this later today.

--
- Joe Ferguson
JoeFerguson.me
MemphisPHP.org

syntacticNaCl:
Sounds good @svpernova09. Should we move the package installation to settler and keep the config in the Homestead.yaml file and start-xvfb.sh in homestead so that the user can disable it if they want to run the framebuffer manually?

taylorotwell:
Would like to see support for this for sure. Will play with it as well.

svpernova09:
Going to decline this for now, since we'll likely be splitting up this implementation across settler & Homestead.

Thanks for taking the time to share this.

@taylorotwell Have at this one, I'm still recovering from back to back trips so I may not get back to this until the weekend.

syntacticNaCl:
Sounds good. Thanks for reviewing @svpernova09 @taylorotwell

PR #533, opened 2017-03-30
Fix bash typo in create-certificate.sh

wturrell:
Without this, typical output for each certificate is:
==> homestead: /tmp/vagrant-shell: line 11: -f: command not found
and existing certificates are regenerated unnecessarily.

(NB: I hope I'm on the right branch, `2.0` seems older)


svpernova09:
Thanks for fixing this.

PR #536, opened 2017-04-04
Fix for apache log dir variable and log file per site

padalec:
Before my changes all logs was placed in root folder and all sites was saved to one file for access log and one for error log.


svpernova09:
Thanks for the PR.

Is the leading `\` before `${APACHE_LOG_DIR}` required? If not lets remove.

padalec:
Yes its, because without \ when this setting was saved to apache config file this variable does not exists is replaced by empty string

svpernova09:
Thanks!

padalec:
You welcome :)

PR #538, opened 2017-04-07
For Vagrant 1.9.3

overbid:
https://github.com/mitchellh/vagrant/issues/8395


svpernova09:
Thanks for the PR, I'm not sure we want to do this for *everyone* since it's only affecting Windows Users on Vagrant 1.9.3 and it's already fixed in Vagrant (pending release)

For now my advice would be to have Homestead users affected by this manually change the file as described. 

If this continues to be a problem we can revisit.

PR #540, opened 2017-04-10
Use full path to Homestead config file

driesvints:
This allows us to halt the VM from anywhere on the system.

Otherwise you'll get this:

```bash
$ vagrant halt c67044b
Homestead settings file not found in /Users/driesvints/Sites/laravelio/vendor/laravel/homestead
```


svpernova09:
Thanks for the PR!

PR #542, opened 2017-04-12
Database import/export aliases

thecrypticace:
I've added two commands to handle importing / exporting all databases from ones homestead virtual machine.

### dbexport [file]

This exports a compressed dump of all databases to the given file.
If no file is given then `/vagrant/mysqldump.sql.gz` is assumed.

### dbimport [file]

This imports a compressed dump from the given file.
If no file is given then `/vagrant/mysqldump.sql.gz` is assumed.

## Progress reporting

I've also added support for optional progress reporting if you have the program `pv` installed. If not you can install it via `sudo apt-get install -y pv`. The commands output a note when running them about `pv` support like so:

### Installation Note
```
Want to see export progress?
Install pv with `sudo apt-get install -y pv` then run this command again.

Exporting databases to '/vagrant/mysqldump.sql.gz'
```

### During import
```
Importing databases from '/vagrant/mysqldump.sql.gz'
[=======>                                                      ]  10% ETA 0:05:37
```

### During export
```
Exporting databases to '/vagrant/mysqldump.sql.gz'
Estimated uncompressed size: 2.642GiB
63.6MiB 0:00:03 [24.1MiB/s] [=>                                ]  2% ETA 0:01:41
```
## Notes

- This would be rather useful when destroying the homestead machine if you have databases which have dummy or in progress data which may be non-trivial to recreate.
- The `/vagrant` directory is used as a default as it will persist even after destroying the machine since it is a shared folder.
- It would be kinda awesome to have `pv` installed by default in the homestead box.

### DIFF
--- a/resources/aliases
+++ b/resources/aliases
@@ -133,6 +133,14 @@ function __has_pv() {
     return $?
 }
 
+function __pv_install_message() {
+    if ! __has_pv; then
+        echo $1
+        echo "Install pv with \`sudo apt-get install -y pv\` then run this command again."
+        echo ""
+    fi
+}
+
 function dbexport() {
     FILE=${1:-/vagrant/mysqldump.sql.gz}
 
@@ -140,6 +148,8 @@ function dbexport() {
     # It appears that 80% is a good approximation of
     # the ratio of estimated size to actual size
     SIZE_QUERY="select ceil(sum(data_length) * 0.8) as size from information_schema.TABLES"
+
+    __pv_install_message "Want to see export progress?"
 
     echo "Exporting databases to '$FILE'"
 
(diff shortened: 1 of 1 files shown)


resources/aliases
```diff
@@ -126,3 +126,58 @@ function share() {
 function flip() {
     sudo bash /vagrant/scripts/flip-webserver.sh
 }
+
+function __has_pv() {
```
thecrypticace: I wasn't sure of the best way to handle this. I underscored these helpers because they're only meant to be used by the import/export functions. I wish Bash had classes and access control.

svpernova09:
Thanks for the PR!

(If you want, open a PR to https://github.com/laravel/settler that adds `pv` so Homestead will have it by default.

thecrypticace:
Will do.

paulcervov:
@thecrypticace Thanks for this functionality, very useful!

Issue #544, opened 2017-04-15
xdebug cli alias


tomcastleman:
Hi there,

Would users find value in having a standard alias for running php with xdebug configured and enabled from the cli which works with PHPStorm out the box?:

```bash
function xdebug() {
    (php -m | grep -q xdebug)
    if [[ $? -eq 0 ]]
    then
        XDEBUG_ENABLED=true
    else
        XDEBUG_ENABLED=false
    fi

    if ! $XDEBUG_ENABLED; then xon; fi

    php \
        -dxdebug.remote_enable=1 \
        -dxdebug.remote_host=10.0.2.2 \
        -dxdebug.remote_autostart=1 \
        "$@"

    if ! $XDEBUG_ENABLED; then xoff; fi
}
```
I have included only the xdebug settings we need to override from default.

This could be used as follows:

`$ xdebug artisan foo`

This works with PHPStorm with zero configuration.

svpernova09:
Curious what others think, I don't have a preference. Is this solving the problem where people just aren't aware of the existing aliases to toggle xdebug on the command line? If so we just make mention in the docs to raise awareness rather than adding another alias.

tomcastleman:
Followed up with #545

PR #545, opened 2017-04-16
Adds alias for php-cli with xdebug and auto remote connect

tomcastleman:
Following on from #544:

The problem is that out-the-box homestead 5.2.1 & box 2.1.0 do not configure xdebug on the cli to automatically connect to a remote xdebug server (listening in PHPStorm for example). It took me some time and research to figure out how to get this working. 

In the spirit of Laravel this PR should save users the headache by providing a simple command for running a php script on the cli (artisan for example) with xdebug configured to automatically connect back to a listening xdebug server with the default settings. I have used the bridge gateway IP (192.168.10.1) so hopefully this will work across all providers (I have tested with Virtualbox).

As mentioned this works with PHPStorm 2017.1 with default settings, simply toggling on the _Start listening for PHP Debug connections_ option.

Example use would be:

`xdebug artisan foo:bar`

### DIFF
--- a/resources/aliases
+++ b/resources/aliases
@@ -182,7 +182,7 @@ function dbimport() {
     echo "Done."
 }
 
-function xdebug() {
+function xphp() {
     (php -m | grep -q xdebug)
     if [[ $? -eq 0 ]]
     then
--- a/resources/aliases
+++ b/resources/aliases
@@ -181,3 +181,22 @@ function dbimport() {
 
     echo "Done."
 }
+
+function xdebug() {
+    (php -m | grep -q xdebug)
+    if [[ $? -eq 0 ]]
+    then
+        XDEBUG_ENABLED=true
+    else
+        XDEBUG_ENABLED=false
+    fi
+
+    if ! $XDEBUG_ENABLED; then xon; fi
+
+    php \
+        -dxdebug.remote_host=10.0.2.2 \
+        -dxdebug.remote_autostart=1 \
+        "$@"
+
+    if ! $XDEBUG_ENABLED; then xoff; fi
+}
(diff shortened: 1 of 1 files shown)


tomcastleman:
Or we could name it `xphp` perhaps.

jonnywilliamson:
I'm a bit slow...but I'm still trying to work out what this helps with!

However on a side note, this code: `xdebug.remote_autostart=1` 

Does this run the risk that if a user turns this on, does some debugging and forgets to turn this off, that all their future commands will still try to make a connection back to a debug server?

tomcastleman:
This alias does not leave anything 'turned on'. It simply does the following:

1) if xdebug is not enabled for php-cli, it enables it using the existing `xon` alias.
2) php is executed with the 2 extra xdebug settings to ensure xdebug automatically connects to the vm host, passing through the script and any arguments given to the alias. These extra settings are only valid for this execution of php.
3) if xdebug was not enabled before the alias was run, it is turned off using the existing `xoff` alias.

jonnywilliamson:
Ah ok. I've never needed to do any debugging of command line stuff, so this was under my radar.

tomcastleman:
With an easy alias like this maybe now you can start 😉

svpernova09:
Lets the change the name of the alias to something more descriptive of what's happening. Maybe `enable_xdebug_cli` or something?

tomcastleman:
As it's invoking php with xdebug enabled `php_with_xdebug` would be more appropriate if little long. I think `xphp` sits nicely with the existing `xon` and `xoff` aliases.

svpernova09:
+1 for `xphp`

svpernova09:
Thanks for the PR!

Issue #559, opened 2017-05-02
Aliases should *always* be copied over


svpernova09:
In the per project installation default method the user does *not* get the `aliases` file copied over. This can be confusing when reading the documentation about some of the shortcut commands that have been added.

TODO:
* The Homestead `make` command should always copy `aliases` and `after.sh` if it doesn't exist. It should never overwrite
* The Homestead `make` option for `aliases` should be changed to `--no-aliases` and not copy the file `aliases` if used
* The Homestead `make` option for `after` should be changed to `--no-after` and not copy the file `after.sh` if used
* Update/Add tests to verify new functionality

svpernova09:
We should probably think about making `after.sh` always copy over as well.

sergeyklay:
Yeah, good catch. We did this already https://github.com/phalcon/box/blob/master/src/aliases.rb

ConnorVG:
The regression from this would probably piss me off, honestly.

svpernova09:
@ConnorVG What regression? If you have an existing `aliases` file in the project root, the `make` command would not overwrite it. If the `make` command is overwriting any files (except in some cases `Homestead.yaml`) that's probably a bug.

ConnorVG:
The regression is changed behaviour that alters the environment of the VM. What do you mean "What regression?" – we're literally asking here to implement a _change to behaviour_ that _mutates the vm environment_.

Regression is integral to this 'issue'.

svpernova09:
This change will only affect users who are using the per project installation method. The implementation should not overwrite existing files so the is no way to break a user's current working version on upgrade or subsequent runs of the `make` command.

I don't understand what part will "piss" you off, can you help me understand?

ConnorVG:
Per-project installation is very common (especially in the work places I have been in).

This is a regression issue as every new installation from this change on-wards would add a new `aliases` set to my box – which will add / change functionality based on what I'm used to.

>This **change** will only **affect** users 

See: "Regression".

svpernova09:
So you have a problem that every new install would have these files that the previous versions wouldn't. I can see that being annoying.

The way around that would be to `composer update laravel/homestead` on each of those projects and re-run the `make` command to generate the missing files (if desired). Alternatively you could just delete the files after the make command ran if you didn't want them.

Probably a better scenario would be to reverse the current flags: `--aliases --after` to `--no-aliases --no-after` where the default behavior would be to create the files, but give the user an easy override to not even generate them.

Thanks for the feedback.

[...]

PR #564, opened 2017-05-09
Update aliases and after options

DojoGeekRA:
As requested in #559

* The Homestead make command should always copy aliases and after.sh if it doesn't exist. It should never overwrite already existing files.
* The Homestead make option for aliases should be changed to --no-aliases and not copy the file aliases if used
* The Homestead make option for after should be changed to --no-after and not copy the file after.sh if used


DojoGeekRA:
Done @svpernova09, sorry for the delay.

svpernova09:
Thanks for the PR! I was just about to start on this 👍

PR #572, opened 2017-05-17
✨ 🚑 Create SSL certs when creating sites via aliases shortcut commands

### LINKED ISSUE #570: Serve doesn't work correctly inside the VM
When running `serve` inside the VM (I've not checked any of the other serve scripts but they're probably affected too) it does not create TLS certificates and will cause nginx to fail on restart. This is because when sites are created they have certificates defined by default but they have to be created by calling `/vagrant/scripts/create-certificate.sh` with the domain as an argument.

## Versions
  - Vagrant: v1.9.3
  - Provider: VMWare Fusion v8.5.2 / Vagrant Fusion Plugin v4.0.15
  - Homestead: Repo v5.2.1 / Box v2.1.0

## Host operating system
macOS Sierra v10.12.4

## Expected behavior
Running `serve test.dev /home/vagrant/test/public` should not cause nginx to fail on its next restart

## Actual behavior

Nginx fails to restart because no certificates exist for the new site.

## Steps to reproduce
1. Run `serve test.dev /home/vagrant/test/public` inside the vm
2. Run `sudo service nginx restart`
3. Run `sudo service nginx status` to see the details and the error message about the missing cert/key files.

## Additional note

As an aside this happens every time I run `serve` (though it doesn't prevent it from doing anything):
```
dos2unix: Failed to change the owner and group of temporary output file /vagrant/scripts/d2utmpJWJ0l6: Operation not permitted
dos2unix: converting file /vagrant/scripts/serve-laravel.sh to Unix format ...
dos2unix: problems converting file /vagrant/scripts/serve-laravel.sh
```

It's rather annoying and I figured I'd bring it up. My guess is you can't actually change permissions for anything inside `/vagrant` on VMWare at least so it would need to operate on a copy that is located in some place like `/tmp` or something.

svpernova09:
Resolves / Fixes #570

### DIFF
--- a/resources/aliases
+++ b/resources/aliases
@@ -29,6 +29,7 @@ function dusk() {
 function serve-apache() {
     if [[ "$1" && "$2" ]]
     then
+        sudo bash /vagrant/scripts/create-certificate.sh "$1"
         sudo dos2unix /vagrant/scripts/serve-apache.sh
         sudo bash /vagrant/scripts/serve-apache.sh "$1" "$2" 80
     else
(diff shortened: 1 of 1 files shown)


thecrypticace:
🎉

PR #581, opened 2017-05-24
Feature/localized files

### LINKED ISSUE #574: Serve aliases broken on per project installs
the `serve`* aliases from the `aliases` files point to `scripts/...` files that don't exist in that path on per project installs, they would exist at `vendor/laravel/homestead/scripts/...`.

Maybe the `make` command should do some find & replace on the aliases file that fixes this, or just like we have with a `resources/LocalizedVagrantfile` we localize the aliases file with the static paths already set.

/cc @DojoGeekRA, @thecrypticace

DojoGeekRA:
Fixes #574


DojoGeekRA:
@svpernova09 please don't merge this, is still work in progress but i wanted to push some progress.

DojoGeekRA:
@svpernova09 the `MakeCommand` is updated and it will now copy the files under `resources/localized` if homestead has been installed per project.

It determines if is a per project installation based on [this question](https://github.com/laravel/homestead/pull/581/files#diff-8d2c32ea2624d134ca2f7d2bfd722b3bR106), what do you think about it?

I have also added two tests for it, unfortunately i can't emulate a per project installation so i have marked them as skipped. When we achieve that, the assertions are the right ones.

And finally, this PR allow edit from maintainer so if you can help me to update the `aliases` file with the right paths will be great.

Thanks and let me know if you have any questions 😃

svpernova09:
I think that check on the folder works fine for this use case. I'll have a look at this later tonight. Thanks!

svpernova09:
Just tested this out, looks good. Thanks for the PR!

DojoGeekRA:
Thanks @svpernova09!

I haven't dropped the task yet, i will look for a way to emulate the per project installation and make those tests pass :)

svpernova09:
Thanks @DojoGeekRA You're awesome.

PR #583, opened 2017-05-25
Broken SSL support is fixed

umutkirgoz:
Certificate file locations were wrong, they are fixed. Ports are not hard coded right now.


.gitignore
```diff
@@ -5,3 +5,4 @@ phpunit.xml
 /Homestead.json
 /after.sh
 /aliases
+/.idea
```
svpernova09: This should be in a user's global gitignore file. Projects shouldn't be responsible for ignoring files generated outside of the project.

umutkirgoz: :) sorry for that

svpernova09:
Thanks for the PR!

PR #591, opened 2017-06-08
couchdb support

JulioECastro:


scripts/install-couch.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
svpernova09: We need to tell apt to use noninteractive mode before we run any apt commands.

PR #601, opened 2017-06-22
✨ 💡 Do not use the `Code` subfolder in vagrant, should use top level

svpernova09:
There is no reason to add another folder inside vagrant that we map our projects to. This change will eliminate sites being placed in `/home/vagrant/Code/` and place them in `/home/vagrant/`

/cc @DojoGeekRA

### DIFF
--- a/src/Settings/HomesteadSettings.php
+++ b/src/Settings/HomesteadSettings.php
@@ -102,7 +102,7 @@ abstract class HomesteadSettings
     {
         $site = [
             'map' => "{$projectName}.app",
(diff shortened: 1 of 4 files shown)


browner12:
this seems kind of unnecessary. is there an issue with leaving it as is, or are we gaining some kind benefit from it?

svpernova09:
@browner12 For me it's an annoyance to cd past another folder to reach my code base. It's a small thing, and hopefully more of a quality of life improvement

DojoGeekRA:
@svpernova09 looks good to me 👍

browner12:
sounds good, thanks

PR #613, opened 2017-07-18
The safer, the better

m1guelpf:
Any link that can be HTTPS, should be HTTPS :closed_lock_with_key:


svpernova09:
Thanks for the PR!

PR #624, opened 2017-07-29
❄️ 🚑 ✅ Make command should respect Homestead config example file

svpernova09:
The make command should now correctly copy over all site options and folder options from an existing example file.

@DojoGeekRA Can I get you to have a look at this. A few of the tests changed because we're paying more attention to the example file VS default config.

/cc @dtirer

PR #627, opened 2017-08-02
Contributions to #624

DojoGeekRA:
@svpernova09 i have read all your implementation in #624 and it looks solid. Reading more deeply all the questions that i was to sent you were clearly expressed in the code.

I have contributed with just two things:

- I have restored the docblocks style to maintain consistency with the repositories `laravel/laravel` and `laravel/framework`. Pretty much every class serves as an example of two spaces after the param types.

- I added tests for shared folders configurations following your pattern `from_empty_settings` and `from_existing_settings`.

And basically that's it, please let me know if you have questions. Thanks!

### DIFF
--- a/src/Settings/HomesteadSettings.php
+++ b/src/Settings/HomesteadSettings.php
@@ -14,7 +14,7 @@ abstract class HomesteadSettings
     /**
      * JsonSettings constructor.
      *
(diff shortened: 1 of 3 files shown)


svpernova09:
Thanks @DojoGeekRA !

PR #631, opened 2017-08-03
allow serve functions to specify a php version

browner12:
a little bug introduced by the recent ability to specify PHP verions.

the `serve` functions are currently failing because they do not specify a PHP version, and there is no default.

this PR allows the `serve` functions to specify a specific PHP version (5.6, 7.0, or 7.1 currently) or omit it and be given a default.

2 thoughts for the future:

- not sure if there's a better place to define the default so we're not hardcoding it in everywhere here.
- probably should write a test to avoid this regression again.


svpernova09:
Thanks for the PR, i'll have a look at this over the weekend.

svpernova09:
Thanks for the PR!

PR #633, opened 2017-08-04
Make timezone configurable

maff:
Adds a configuration entry to make timezone configurable from `UTC` to a local timezone. E.g.:

```
timezone: Europe/Vienna
```


svpernova09:
Thanks for taking the time to share a PR.

This is something that should be overridden via `after.sh` since setting server times to UTC/GMT seems to be an industry standard.

maff:
That's what I was doing in the past but thought of being a standard problem which could be handled by a configuration entry. Thanks for reviewing though :)

PR #634, opened 2017-08-04
Add support for Pimcore

maff:
This adds support for Pimcore (https://www.pimcore.org/). As we use apache as standard dev setup and ship a `.htaccess` it was more straightforward to create an apache configuration.


scripts/serve-pimcore5.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
+
+# install apache and bzip PHP extension
+export DEBIAN_FRONTEND=noninteractive
```
svpernova09: I don't like the idea of having the "install apache" code now in 2 places. That's not necessarily a blocker for getting this PR merged, just something I want to think about a better solution for

maff:
Thanks for the review! I updated the type to `pimcore`.

scripts/serve-pimcore5.sh
maff: I see your point - maybe make the apache installation an extra script which can be required from types building on apache?

svpernova09:
Thanks for the PR!

PR #635, opened 2017-08-04
Add Hyper-V support

jm-mwi:
I'm not sure if the issue should be added as a PR or an actual issue, but here is a PR with the files that I believe need to be changed (not sure).

My issue is that Vagrant says it [supports Hyper-V](https://www.vagrantup.com/docs/hyperv/index.html), but I'm not able to use homestead since it doesn't implement any adapter for it.

Please add support for Hyper-V as it works BEAUTIFULLY on a windows box. 

While Virtualbox is free I consistently have bugs with it and the machine complains, on shutdown, about a bug/issue with the Virtualbox Adapter drivers.

### DIFF
--- a/resources/Homestead.json
+++ b/resources/Homestead.json
@@ -2,7 +2,7 @@
     "ip": "192.168.10.10",
     "memory": 2048,
     "cpus": 1,
-    "provider": "virtualbox",
+    "provider": "hyperv",
     "authorize": "~/.ssh/id_rsa.pub",
     "keys": [
         "~/.ssh/id_rsa"
--- a/resources/Homestead.yaml
+++ b/resources/Homestead.yaml
@@ -2,7 +2,7 @@
 ip: "192.168.10.10"
 memory: 2048
 cpus: 1
-provider: virtualbox
+provider: hyperv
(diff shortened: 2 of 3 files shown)


svpernova09:
Thanks for the PR, however this is not how we'd add another provider to Homestead.

I've tried multiple times now to support Hyper-V and I just can't get it to the point where it is as stable on Windows as the other providers due to Microsoft not opening up all of the APIs on Win10 professional and due to Vagrant having a tendency to randomly break things in Vagrant from Windows.

jm-mwi:
Well, that's unfortunate. As an alternate option has there been any thought of moving to Docker and homestead being a container utility instead of a VM utility?

svpernova09:
Homestead by definition is "official, pre-packaged Vagrant box". There are no official Laravel docker container setups although https://github.com/laradock/laradock is the closest thing to that. You're not really solving anything by putting Homestead into a container or creating a group of containers to mimic Homestead because the principle of containers are to only bring what you need, where Homestead is "everything you need and then some". There's just not a practical application where docker makes sense for Homestead.

PR #639, opened 2017-08-09
PHP version aware for prod env

Prometee:


svpernova09:
Good catch! Thanks for the PR

PR #644, opened 2017-08-12
Enable gzip in laravel nginx config

markussorg:
This will enable gzip in laravel projects. It's nice to see gzipped asset file sizes during development. The settings are the ones which forge is using.


svpernova09:
Thanks for taking the time to share a PR. I think this is something people should enable themselves for testing instead of it being on by default.

PR #647, opened 2017-08-15
Add aliases for easy switching between PHP versions

pascal08:
This adds 3 aliases for switching between 3 PHP versions provided by Homestead: php5.6, php7.0 and php7.1.

This is useful when using Homestead for multiple projects with the same Homestead configuration.


pascal08:
Check.

svpernova09:
Thanks for the PR!

PR #648, opened 2017-08-16
Use function instead of alias to swith php version

ricardoseriani:
Remove the alias, since there is already a function to switch php version


svpernova09:
thanks!

PR #649, opened 2017-08-16
Allow pimcore serve function to specify a PHP version

maff:
As #631 and #634 were done at the same time, the Pimcore serve function is missing the changes from #631.


svpernova09:
Thanks for the PR!

Issue #529, opened 2017-03-27
Revert back to Mariadb 10.1, the latest stable release


sisve:
Pull request https://github.com/laravel/homestead/pull/441 (Update Mariadb to 10.2.* to work with Laravel 5.4) introduced Mariadb 10.2 which is not a stable release yet. (It was a beta version when the switch occured,but has since reached a release candidate status.) This change should probably be reverted back to 10.1, the latest stable release of Mariadb.

If this update was done due to the "Specified key was too long; max key length is 767 bytes" users may run into with Laravel 5.4, couldn't we update their my.cnf file instead?

    innodb_file_format=Barracuda
    innodb_large_prefix=1

References:
https://downloads.mariadb.org/mariadb/+releases/
https://github.com/laravel/framework/issues/18201#issuecomment-284131211

svpernova09:
@taylorotwell my thought is we should be using stable versions of things. But not sure about having to patch `my.cnf`. Thoughts?

andrewhood125:
I actually lean towards dealing with this at the application level. I think it's reasonable to have Homestead be the latest _stable_ version with a relatively standard configuration. 

https://laravel-news.com/laravel-5-4-key-too-long-error

andrewhood125:
Of course `after.sh` is always an option for folks.

sisve:
Consider that people are building applications on homestead running mariadb, but they can then not find a compatible host because they require functionality that in the stable release of mariadb yet. No production environment is running the release candidate.

If we still want to provide support for 10.2, couldn't we extend the mariadb configuration?
`mariadb: [ false | true | stable | 10.1 | 10.2 | 10.3 ]` where true==stable. It would allow people to specifically say "I want Mariadb 10.2", while those that with old files stay on the stable version (which will be 10.2 once it is stable).

Perhaps a bit overkill, but it would allow both types of users, default to stable but allow opt-in to the current release candidate (10.2) or the current development version (10.3).

svpernova09:
I'm fine with supporting multiple versions but at the first sign of trouble I'm going to drop support for anything that gives me trouble in the normal Settler -> Homestead -> Release Cycle. I'll always decide to remove (or just not support) something from Homestead (in this case a version of MariaDB) instead of risking shipping the end users something that is broken.

[...]

PR #657, opened 2017-08-30
Upgrade MariaDB version from 10.1 to 10.2 (stable)

cbj4074:
Note: This commit differs slightly from the initial attempt at https://github.com/laravel/homestead/pull/441/files in that this commit addresses the erroneous `default_password_lifetime` directive, too.

For the reasons described from https://github.com/laravel/homestead/issues/529#issuecomment-316290010 to the end of the thread, it seems appropriate to re-introduce Maria DB 10.2.x, now that the 10.2 series is marked GA/stable, per https://mariadb.com/kb/en/the-mariadb-library/changes-improvements-in-mariadb-102/ .

Also, according to https://mariadb.com/kb/en/the-mariadb-library/system-variable-differences-between-mariadb-101-and-mysql-57/ , the `default_password_lifetime` is a `MySQL-only variable determining how long passwords are valid for before expiring.`

With this directive present, Homestead (v6.1.0) provisioning fails with:

    ==> homestead-7: mysql: unknown variable 'default_password_lifetime=0'

Removing this directive resolves the problem.

### DIFF
--- a/scripts/install-maria.sh
+++ b/scripts/install-maria.sh
@@ -30,24 +30,20 @@ rm -rf /etc/mysql
 # Add Maria PPA
 
 sudo apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xF1656F24C74CD1D8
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks, I'll have a look at this later tonight

svpernova09:
Thanks for the PR!

PR #662, opened 2017-09-06
Overhaul and update MongoDB installation script

cbj4074:
The impetus for this rewrite is that from the initial commit, every php-cli invocation caused PHP to emit a warning. This can be tested/verified as follows:

    vagrant@homestead:~$ php -r 'echo "test\n";'
    PHP Warning:  Module 'mongodb' already loaded in Unknown on line 0
    test

The root-cause turned out to be that the PECL installer was adding `extension = mongodb.so` to the very top of `/etc/php/7.1/cli/php.ini` (along with entries for several other extensions, which in itself seems like a bug with the PECL installer).

Once I dug into the source, it became clear that the MongoDB driver was being installed only for PHP 7.1, and none of the other (earlier) PHP versions for which support was added recently. The only means by which to install the driver for all PHP versions is to build it from source for each version. This change addresses that need.

In any case, the nature of the proposed changes is as follows:

1. Upgrade MongoDB version from 3.2 to 3.4, and remove `--allow-unauthenticated` switch from `apt-get install` command (there should be no reason to use this switch).

2. Remove sed command that seems designed to modify /usr/bin/pecl directly, which seems like a "bad idea". Also, as of this writing, `/usr/bin/pecl` is untouched after the command is executed, so perhaps the bug it was designed to address has been fixed upstream.

3. Remove duplicate package names when installing packages required to build MongoDB PECL extension and consolidate into a single `apt-get` command.

4. Remove inline systemd init script for MongoDB, as the official package installs its own, and rename references to match common convention (convention is `mongod`, as in "Mongo Daemon", instead of `mongodb`). Also, remove needless call to `systemctl status mongodb`, as it produces "noise", and fix the order in which systemd calls are executed post-installation.

5. Build and enable mongodb PECL extension for all versions of PHP (not just 7.1). Previously, the extension was built and enabled only for PHP 7.1.

6. Remove needless call to restart nginx, as nginx is uninvolved with this process.

### DIFF
--- a/scripts/install-mongo.sh
+++ b/scripts/install-mongo.sh
@@ -16,9 +16,6 @@ sudo apt-get update
 
 sudo DEBIAN_FRONTEND=noninteractive apt-get -yq -o Dpkg::Options::="--force-confnew" install mongodb-org autoconf g++ make openssl libssl-dev libcurl4-openssl-dev pkg-config libsasl2-dev php-dev
 
-#sudo pecl update-channels
-#sudo pecl install mongodb
-
 sudo rm -rf /tmp/mongo-php-driver /usr/src/mongo-php-driver
 git clone -c advice.detachedHead=false -q -b '1.2.9' --single-branch https://github.com/mongodb/mongo-php-driver.git /tmp/mongo-php-driver
 sudo mv /tmp/mongo-php-driver /usr/src/mongo-php-driver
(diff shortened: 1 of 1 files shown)


svpernova09:
Fantastic work. Thanks for the PR!

cbj4074:
Any time! Thank you for your kind words and for acting on it so quickly! :+1:

DFurnes:
This is amazing! I was so glad when I found out Homestead has hidden MongoDB support. 🙌 I tested out the latest release with this change, and might have noticed two small bugs. I made [a pull request](https://github.com/laravel/homestead/pull/666) with the fixes that worked for me, and would love some eyes on it!

PR #665, opened 2017-09-11
Sign site-certificates with own root Certificate

cored0wn:
Note: Helpful for the development of ServiceWorkers

Problem:
The developer calls a homestead-site via HTTPS. By default, the browser throws an error message like "ERR_CERT_AUTHORITY_INVALID", because the certificate is self-created and therefore not trustworthy.
If you use multiple sites in a homestead-box, each certificate must be imported manually into the local system.

Solution:
You create your own "root"-certificate and load it into your local certificate store. Then copy the certificate and its key into the box using the "copy"-parameter. Next, specify the copied certificate and the certificate key for each site using the "rootCA"-parameter.
The box now signs the certificate of each site with the certificate specified in "rootCA" during the "provision"-process.
If the developer now calls the site via HTTPS, no error message is generated and he does not have to load the corresponding certificate into his memory for each site.

Implementation:
I have added the parameter "rootCA" in homestead.rb. This expects two parameters "key" and "crt".
I also modified the create-certificate.sh so that it checks if a "rootCA"-parameter has been passed. If so, a new certificate is created and signed with the root certificate. Otherwise, the code already implemented will be executed.

Example:
Generate root CA:

`openssl genrsa -out ca.key 2048`
`openssl req -x509 -new -nodes -key ca.key -sha256 -days 3650 -out ca.pem -subj "/C=<COUNTRY>/ST=<STATE>/L=<CITY>/O=<COMPANY NAME>/OU=<ORGANISATION UNIT>/emailAddress=<EMAIL ADDRESS>/CN=<COMMON NAME>"`

After that you have to import your generated CA in your local certificate store.

Now copy your certificate and key in your homestead-box and specify it for every site 
you will use it. (See attached homestead.yaml.txt)

Notice: I've must change the file-type because github doesn't accept *.yaml-files.

[Homestead.yaml.txt](https://github.com/laravel/homestead/files/1292594/Homestead.yaml.txt)

### DIFF
--- a/scripts/create-certificate.sh
+++ b/scripts/create-certificate.sh
@@ -3,19 +3,55 @@
 mkdir /etc/nginx/ssl 2>/dev/null
 
 PATH_SSL="/etc/nginx/ssl"
-PATH_CNF="${PATH_SSL}/${1}.cnf"
+PATH_CSR="${PATH_SSL}/${1}.csr"
+PATH_req_CNF="${PATH_SSL}/${1}_req.cnf"
+PATH_x509_CNF="${PATH_SSL}/${1}_x509.cnf"
 PATH_KEY="${PATH_SSL}/${1}.key"
 PATH_CRT="${PATH_SSL}/${1}.crt"
 
 # Only generate a certificate if there isn't one already there.
-if [ ! -f $PATH_CNF ] || [ ! -f $PATH_KEY ] || [ ! -f $PATH_CRT ]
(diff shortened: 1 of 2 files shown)


svpernova09:
Trusting insecure Certificates is not a practice I want to see Homestead facilitate by default. This is something I'm happy for users to do but I don't want to make it easy. I want them to understand the risks associated with what they're doing.

Thanks for taking the time to put together this pull request.

PR #667, opened 2017-09-16
lowercase 'code' directory

browner12:
not sure why this was uppercase originally, but it's an unnecessary extra keystroke

PR #672, opened 2017-09-22
Remove code directory

DojoGeekRA:
This will allow to share everything inside the `~/code` from the host OS directly into the vagrant shared directory `/home/vagrant`.

This complements the work in the PR #601

### DIFF
--- a/resources/Homestead.json
+++ b/resources/Homestead.json
@@ -10,7 +10,7 @@
     "folders": [
         {
             "map": "~/code",
-            "to": "/home/vagrant/code"
+            "to": "/home/vagrant"
         }
     ],
     "sites": [
--- a/resources/Homestead.yaml
+++ b/resources/Homestead.yaml
@@ -11,7 +11,7 @@ keys:
 
 folders:
     - map: ~/code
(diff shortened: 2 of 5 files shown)


DojoGeekRA:
@svpernova09 this was updated based on your comment in #667. Seems like this was the last occurrence.

@browner12 i think that the uppercase `~/Code` directory is because in Mac OS, every directory inside `~` starts with the first letter uppercased:

captura de pantalla 2017-09-22 a la s 18 09 52

Maybe that's why @taylorotwell  left that default as he uses Mac OS, but it is just a guess 😅

browner12:
yah, i'm sure you're right. MacOS is writing them for your gui users. I'm going to assume a lot of us are on the command line most of the time.

svpernova09:
Thanks for the PR!

antimech:
@svpernova09 Why you reverted 03d8192? It is like before now.

svpernova09:
@antimech it broke paths in the per project installation method during the `make` command.

antimech:
@svpernova09 Oh, thanks!

PR #675, opened 2017-09-23
Revert "Remove code directory"

svpernova09:
Reverts laravel/homestead#672

### DIFF
--- a/resources/Homestead.json
+++ b/resources/Homestead.json
@@ -10,7 +10,7 @@
     "folders": [
         {
             "map": "~/code",
-            "to": "/home/vagrant"
+            "to": "/home/vagrant/code"
         }
     ],
     "sites": [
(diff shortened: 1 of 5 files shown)


svpernova09:
Reverting this for now. it's currently overriding the default home folder of `/home/vagrant`

DojoGeekRA:
😓 i'm so sorry for this, the failure makes sense.

@svpernova09 should we preserve then the `code` directory in this cases or do you have another approach in mind?

svpernova09:
@DojoGeekRA 

No worries at all. I didn't catch it either, and it explains why there is a code directory to begin with. i think we may leave it as is for now. But make sure the per project (make command) doesn't use it. Since we don't know what a user may be doing on a global install we want that code folder to protect them from this kind of bug. With the make command, we know exactly the path they're using so the code folder is only adding another level to traverse and we can easily skip it.

DojoGeekRA:
@svpernova09 understood, i agree with you and will work on that, thanks! 😃

PR #678, opened 2017-09-27
Turn on HTTPS for Apache

kalenjohnson:
Seems like a lot of work went into getting HTTPS support for Apache. But it's not turned on by default.

So, turn it on.


svpernova09:
Geez, I *guess* we can turn it on. :D 

Thanks for the PR!

PR #685, opened 2017-10-09
Create an ngrok config file that allows access to the web inspection …

jonnywilliamson:
…feature.

Ngrok has a brilliant web inspector feature that allows you to view inbound packets to any site on homestead you might have shared.

This is especially useful for monitoring and replaying webhook dataloads etc. However the default setting when you launch ngrok in homestead is for it to bind the this feature to the localhost adapter, making it very hard to make a browser connection to it.

image

This can be easily change/solved by creating a default ngrok config file, and telling ngrok to bind to the normal ip address used by homestead, allowing easy access to the web interface.

This PR attempts to do this. You should end up being able to do this:
image

Pros:
No changes needed to the share command.
Easily changeable
Allows for more options for advanced users

****
Note
****

I've never written a line of ruby code before in my life, the following is purely a guess from looking at other code in the homestead.rb script. Apologies if its terrible.

### DIFF
--- a/scripts/create-ngrok.sh
+++ b/scripts/create-ngrok.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+PATH_NGROK="/home/vagrant/.ngrok2"
+PATH_CONFIG="${PATH_NGROK}/ngrok.yml"
+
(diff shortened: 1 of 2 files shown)


svpernova09:
Thanks! Works well

PR #686, opened 2017-10-10
Fix a few more issues with MongoDB provisioning.

DFurnes:
This pull request fixes a few more issues I've run into with MongoDB on Homestead:

👤 Creates the `homestead` MongoDB user during the "install" script, rather than each time we provision a database (since that would [repeatedly error](https://user-images.githubusercontent.com/583202/31398840-46d4203e-adb8-11e7-8865-1fc00bc2f594.png) when trying to re-create the same user).

🔃 Removes the "restart" step when creating a Mongo database, since it shouldn't be necessary and would cause [a "start request repeated too quickly" error](https://user-images.githubusercontent.com/583202/31398124-e2844cc8-adb5-11e7-8293-45db9224fce9.png) when provisioning a lot of databases.
 
🗝 Finally, fixes a tiny issue where adding the MongoDB public key would display as an error by suppressing output for that particular command from stderr. This just removes [some ugly red text](https://user-images.githubusercontent.com/583202/31397827-002559a8-adb5-11e7-9a7b-bd20f6889932.png) from the provisioning output.

### DIFF
--- a/scripts/create-mongo.sh
+++ b/scripts/create-mongo.sh
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
 
-mongo admin --eval "db.createUser({user:'homestead',pwd:'secret',roles:['root']})"
 mongo $1 --eval "db.test.insert({name:'db creation'})"
 sudo service mongod restart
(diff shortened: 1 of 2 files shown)


svpernova09:
Thanks!

PR #687, opened 2017-10-10
Fix 'vagrant ssh -c' usage.

DFurnes:
This is a potential fix for the issue described in #416 and #514.

According [this comment](https://github.com/hashicorp/vagrant/issues/1673#issuecomment-261041105) in an issue on `hashicorp/vagrant`, the errors are caused by a compatibility issue between `apt-get` and IPv6. From testing on my local, disabling IPv6 while provisioning removes the ugly "Inappropriate ioctl" errors while still allowing the `-c` argument for `vagrant ssh`.

PR #688, opened 2017-10-11
Fix 'vagrant ssh -c' usage.

DFurnes:
After my initial attempt to fix `vagrant ssh -c` with before/after scripts in #687, I noticed it actually seems to work correctly when just removing the line altogether. 🤦‍♂️ It seems like the underlying problem has been fixed in hashicorp/vagrant#8291, which was released in [Vagrant 1.9.2](https://github.com/hashicorp/vagrant/blob/master/CHANGELOG.md#192-february-27-2017).


svpernova09:
Awesome! Thanks! I didn't even realize they finally fixed this. Cheers.

PR #692, opened 2017-10-17
Refactor cron schedule script

DojoGeekRA:
It will now create the `/etc/cron.d` directory only if it does not exists.

Also it creates some variables to improve script readability based on `create-mysql.sh` and `create-postgres.sh` as a precedent for this.


svpernova09:
awesome, thanks!

PR #693, opened 2017-10-17
set netmask on networks setting

raa0121:
my public_network has `172.16.0.0/16` ip segment.
I think that it is desirable to be able to specify netmask with `network["netmask"]`
(However, the default value would be `/24`)

so I write set netmask on networks settings.


svpernova09:
Thanks for the PR! Be careful exposing Vagrant to to the public network :D

PR #696, opened 2017-10-17
Escape ErrorLog and CustomLog

sumityadav:
VirtualHost section of both (SSL/ Non SSL) ErrorLog and CustomLog are not consistent. 

Currently the ${APACHE_LOG_DIR} is parsed as '' and results into: `ErrorLog /error.log`.

Escaping the APACHE_LOG_DIR and adding the site name prefix will write to same file and cause less confusion.


svpernova09:
Thanks for the PR!

PR #699, opened 2017-10-18
Use :: operator instead of . in JSON parse statement

DojoGeekRA:
This allows to maintain consistency with the YAML::load above it.


svpernova09:
👍 nice thanks!

PR #700, opened 2017-10-18
Keep the native colors from the provisioning scripts

DojoGeekRA:
As stated in the vagrant documentation, by default, vagrant automatically colors output in green and red depending on wheter the output is from stdout or stderr. The `keep_color` option allows to maintain the original colors outputted by the provisioning scripts.

Documentation: https://www.vagrantup.com/docs/provisioning/shell.html

Before:

captura de pantalla 2017-10-18 a la s 16 25 17

After:

captura de pantalla 2017-10-18 a la s 16 25 02

As you can see, the red output from the composer command is still red as the original.

If you doubt that the green is bad, take a look at the noise of an `apt update` execution

captura de pantalla 2017-10-18 a la s 16 51 08

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -35,7 +35,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     Homestead.configure(config, settings)
 
     if File.exist? afterScriptPath then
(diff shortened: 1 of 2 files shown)


svpernova09:
Thanks!

PR #701, opened 2017-10-19
🔐 Self-Signed Wildcard Certificates

ompmega:
Automatically generated self-signed certificates will now have a DNS wildcard attached to that domain, and will also point all subdomains to the same server block inside NGINX.

**Screenshots**

### DIFF
--- a/scripts/create-certificate.sh
+++ b/scripts/create-certificate.sh
@@ -39,6 +39,7 @@ then
 
         [ alternate_names ]
         DNS.1 = $1
+        DNS.2 = *.$1
     "
     echo "$block" > $PATH_CNF
 
--- a/scripts/serve-laravel.sh
+++ b/scripts/serve-laravel.sh
@@ -12,7 +12,7 @@ fi
 block="server {
     listen ${3:-80};
     listen ${4:-443} ssl http2;
-    server_name $1;
+    server_name .$1;
     root \"$2\";
 
     index index.html index.htm index.php;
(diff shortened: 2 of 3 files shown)


svpernova09:
Thanks for the PR!

Shkeats:
@ompmega I think this isn't quite right in implementation.

For me with a homestead.yaml sites entry of _subdomain.domain.net_ it generates certificates with two DNS names.

_subdomain.domain.net_
and
_*.subdomain.domain.net_

Should it not do the below?

_subdomain.domain.net_
and
_*.domain.net_

Apologies if I've misunderstood the intention of this feature.

svpernova09:
@Shkeats 

Nice catch. I bet the logic doesn't account for subdomains and just assumes you're only using 1 domain string

ryancwalsh:
@ompmega , @Shkeats , @svpernova09 I have a bounty here since I'm *still* having trouble getting any Laravel Homestead certificates to work: https://stackoverflow.com/q/48969083/470749  Thanks for any help.

ompmega:
Hey @ryancwalsh give this a shot and let me know if it worked for you: https://stackoverflow.com/a/15076602/7782924

So basically, you have to manually trust the self-signed certificates generated by Vagrant.

ryancwalsh:
@ompmega Thanks. I've tried those steps so many times, and even though the certificate exports and imports successfully, I still have the error in Chrome (after restarting): 

```
Certificate - missing
This site is missing a valid, trusted certificate (net::ERR_CERT_AUTHORITY_INVALID).
```

I've opened a new Homestead issue (linked above).

PR #702, opened 2017-10-19
Allow autonetwork plugin to be used.

ryankilf:
Auto network should allow developers to have several homestead boxes running at once without ever having to think about IP addresses.

This solves a particular issue for us, which means we can adopt Homestead as our Vagrant VM across the company.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -16,10 +16,10 @@ class Homestead
         config.vm.hostname = settings["hostname"] ||= "homestead"
 
         # Configure A Private Network IP
-        if settings["ip"] == "autonetwork"
-            config.vm.network :private_network, :ip => "0.0.0.0", :auto_network => true
-        else
+        if settings["ip"] != "autonetwork"
             config.vm.network :private_network, ip: settings["ip"] ||= "192.168.10.10"
+        else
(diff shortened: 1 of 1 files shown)


ryankilf:
Good shout, I have made that change.

svpernova09:
Thanks!

browner12:
@ryankilf  can you ELI5 what this allows us to do? looks interesting, I just don't understand what it's for.

thanks

ryankilf:
Sure, 
It means that if you put 
`ip: "autonetwork"`

in Homestead.yaml your VM will use [Vagrant Auto Network](https://github.com/oscar-stack/vagrant-auto_network) to deal with your IP addresses and your networking and all that.

It's probably less relevant if you're only running one homestead VM, but if you're running lots, and sharing them amongst your team, it's a lifesaver. 

In a previous job we used a page of IP addresses on our wiki. That's not ideal.

browner12:
how do you keep your `/etc/hosts` file up to date then?

ryankilf:
Autonetwork deals with it

ryankilf:
@DojoGeekRA - I just tested this with two sites on one box today, and yes it does.

PR #704, opened 2017-10-19
Modify bind directive before mongod is started

cbj4074:
This fixes a non-obvious regression introduced in https://github.com/laravel/homestead/commit/34e031513766205ce40021e13db05d5b270faa31 .

Please see the comments at the above URL for details.


svpernova09:
Thanks!

PR #708, opened 2017-10-20
Add support for the vagrant-hostmanager plugin

cwt137:
I used a similar approach as the geerlingguy/drupal-vm project to support both `vagrant-hostsupdater` and `vagrant-hostmanager`.


svpernova09:
Thanks for the PR!

ryankilf:
I think this and the autonetwork thing running together means we'll have entries added to the hosts file by both plugins. There may be reasons to use one or the other, but should there be the ability to not use vagrant-hostmanager if you want to use vagrant autonetwork?

PR #720, opened 2017-11-14
Update scripts file permissions

DojoGeekRA:
By default they don't have execution permissions, so if you want to use them in the `after.sh` you need to do something like.

```bash
sudo bash /vagrant/scripts/serve-proxy.sh test.app 8080
```

With these new permissions you can do this directly

```bash
sudo /vagrant/scripts/serve-proxy.sh test.app 8080
```

All of them have a shebang declaration so i assume that indeed they are built to be executed directly.

Also it add new lines at the end of files and after the shebang declaration to maintain consistency across files.


svpernova09:
Nice catch. Thanks!

Issue #721, opened 2017-11-15
Custom (guest) SSH port in Homestead.yaml file


willemo:
## Versions
  - Vagrant: 1.9.8
  - Provider: Virtualbox 5.1.26
  - Homestead: 6.5.0

## Enhancement proposal
Could you add a way to define a different (guest) SSH port through the `Homestead.yaml` file, so it won't conflict with other Vagrant/Homestead boxes on my machine?

To change the port to `22022` manually, I've added the following line to my Vagrantfile:

```
config.vm.network :forwarded_port, guest: 22, host: 22022, auto_correct: false, id: "ssh"
```

After that my `vagrant ssh` command will use port 22022 to connect to the box.

svpernova09:
To override default SSH port for vagrant add default_ssh_port: 64444 to your Homestead.yaml file.

willemo:
@svpernova09 thanks! That works perfectly! Is there anywhere this is documented?

svpernova09:
Not sure if this will make it into the official docs, but if you're using master, the functionality will work. Will tag a new release in a few days / possibly this weekend.

PR #724, opened 2017-11-16
✨ ❄️ Allow Homestead.yaml to override default SSH port for Vagrant

svpernova09:
To override default SSH port for vagrant add `default_ssh_port: 64444` to your Homestead.yaml file.


svpernova09:
Resolves #721

PR #725, opened 2017-11-17
Fix for cron.service: Start request repeated too quickly.

Artistan:
Nov 17 14:02:09 homestead systemd[1]: cron.service: Start request repeated too quickly.

https://laracasts.com/discuss/channels/servers/homestead-create-schedule-fails

This is due to each site configured in the VM restarting the cron service very quickly.
The service does not like this. :) 
I simply added a flag to check if any of the sites have schedule flagged and restart once after the site configurations are done.

### DIFF
--- a/scripts/cron-restart.sh
+++ b/scripts/cron-restart.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+service cron restart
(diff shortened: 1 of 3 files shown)


svpernova09:
thanks for the PR!

PR #729, opened 2017-11-20
missing end after cron-restart.sh fires

brino:
getting error "homestead.rb:377: syntax error, unexpected end-of-input, expecting keyword_end" after `vagrant provision`


DojoGeekRA:
@brino i have refactored this in #730 

An script file for just one line was too much to begin with.

svpernova09:
Thanks for the PR, resolved via #730

Issue #728, opened 2017-11-20
Error when running `vagrant halt`


mikebronner:
I received the following error after pulling the latest this morning.

```~/Homestead/scripts/homestead.rb:377: syntax error, unexpected end-of-input, expecting keyword_end```

(Checking out the latest tag had no issues: `git checkout v6.5.0`)

Thanks :)

svpernova09:
Always fun to find the people brave enough to use `master`

Resolved via #730

mikebronner:
Hehe, glad to help! :) Thanks for the quick fix :)

PR #730, opened 2017-11-20
Refactor cron provisioning

DojoGeekRA:
Removes the flag to check if a site has schedule to trigger the service restart. Now it makes the check directly and its more expressive.

Also, it inlines the script to restart the service.

@svpernova09 there is something that bugs me, here is my scenario:

- A site has schedule and homestead is provisioned.
- A site removes the schedule and homestead is provisioned again.

With the current implementation, homestead will not restart the cron service when the schedule has been removed. It should?

If the answer is yes, then we should restart every time, it will be very cheap because it will happen only once per provisioning. We already have precedent for this with the nginx and php-fpm services that are restarted no matter what in each provisioning cycle.

If the answer is no, then no further action is required.

### DIFF
--- a/scripts/cron-restart.sh
+++ b/scripts/cron-restart.sh
@@ -1,3 +1,0 @@
-#!/usr/bin/env bash
-
-service cron restart
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -184,7 +184,6 @@ class Homestead
         end
 
         if settings.include? 'sites'
-            has_cron = false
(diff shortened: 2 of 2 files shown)


DojoGeekRA:
It fixes #728 too

brino:
I say yes ... it should just restart crond (once) per provision

svpernova09:
@DojoGeekRA 

As always, thanks for the PR.


Yes. I believe cron should be restarted every provision, no matter what.

PR #731, opened 2017-11-20
Restart crond service once per provision

DojoGeekRA:
As discussed in #730 It will now restart the `crond` service once per provision instead of certain situations involving the `schedule` key in the `sites` settings.


DojoGeekRA:
@svpernova09 this applies your feedback in #730, thanks for it 😄

svpernova09:
thanks!

Issue #740, opened 2017-11-29
Error with spaces in DocumentRoot on Apache


mattclements:
I have been setting up Homestead with Apache (for .htaccess reasons), and found what I believe to be a bug. When a DocumentRoot has a space in it (i.e. `/home/vagrant/sites/Test Website/test.co.uk`) the following error is thrown

`Nov 29 20:39:45 homestead apache2[2961]: AH00526: Syntax error on line 7 of /etc/apache2/sites-enabled/test.loc-ssl.conf:
Nov 29 20:39:45 homestead apache2[2961]: DocumentRoot takes one argument, Root directory of the document tree`

I think it's linked to not quoting the DocumentRoot

PR #741, opened 2017-11-29
Fixing #740

mattclements:


svpernova09:
Thanks for the report and PR!

PR #759, opened 2017-12-13
Cron Schedule :: change sites does not remove old

Artistan:
# This will remove all cron.d files (prefixed now) instead of per site.
This is useful for when you reprovision a box with adding some sites and removing others...

Recent Example:  `app.dev` becomes `app.test`

## Possible issues 
 1 - do not remove all
removing ALL cron.d entries if you have custom things in cron.d
## Solution: 
also have the cron-schedule.sh script prefix homestead schedules with something like "homestead-"
Then have `s.inline = "rm -f /etc/cron.d/horizon-*"`

 2 - old crons
anyone reprovisioning with this change, will have duplicate cron.d files
 - `{site_domain}`
 - `horizon-schedule-{site_domain}`
This will not clean up existing boxes now, but it will for future boxes and releases.  I have an idea for this....

### DIFF
--- a/scripts/cron-schedule.sh
+++ b/scripts/cron-schedule.sh
@@ -9,4 +9,4 @@ SITE_PUBLIC_DIRECTORY=$2
 
 cron="* * * * * vagrant /usr/bin/php $SITE_PUBLIC_DIRECTORY/../artisan schedule:run >> /dev/null 2>&1"
 
-echo "$cron" > "/etc/cron.d/$SITE_DOMAIN"
+echo "$cron" > "/etc/cron.d/homestead-schedule-$SITE_DOMAIN"
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -212,6 +212,10 @@ class Homestead
                     s.args = [site["map"], site["to"], site["port"] ||= "80", site["ssl"] ||= "443", site["php"] ||= "7.2", params ||= ""]
                 end
 
+                # Remove all existing Cron Schedules
+                config.vm.provision "shell" do |s|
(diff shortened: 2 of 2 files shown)


svpernova09:
I appreciate the initiative and the time for the PR, but I'd much rather keep the simplicity and instruct users to `vagrant destroy && vagrant up` to ensure there is nothing leftover instead of when a user simply `vagrant reload --provision` which will not clean up after itself.

Virtual Machines (and containers) are cattle, they should not be treated like pets. Don't be afraid to nuke them from orbit (it's the only way to be sure)

svpernova09:
Feel free to hop into #homestead-dev on larachat slack to discuss Homestead feature ideas. I'm always eager to hear what people want. If slack isn't your thing I also have #homestead-dev on Freenode.

Artistan:
there will be a lot of people changing sitenames away from .dev
this is a fix for a provisioning that will break those boxes due to having multiple schedulers running for the same site. -- The system already deletes old schedules for existing sites. I honestly do not see why not to do it cleanly for all cases including sites that are no longer configured.

svpernova09:
I appreciate what you're accomplishing here. But IMHO Homestead shouldn't have to worry about any remnants due to the ease of creating a new VM is ~60 seconds away (depending on your `after.sh`). 

I don't want to introduce complexity into Homestead that solves a problem that Vagrant itself can already solve by a destroy and up.

Artistan:
Looking at Homestead more....
It looks like the `clear-nginx.sh` and `clear-variables.sh` are very much like my proposal here.
I do not want to beat a dead horse, but this could prevent a lot of questions when "changing" a homestead box, just like the clear scripts mentioned and others.

PR #767, opened 2017-12-18
add elasticsearch 6 support

browner12:
- separate scripts for v5 and v6
- add configuration settings to set version


browner12:
interesting note: version 6 has defaulted to allocating 1GB of RAM to the heap.

svpernova09:
Should we drop support for 5?

browner12:
i would prefer to keep it. not everyone upgrades their ES right away.

but we don't need to go crazy and keep every version from here on out, maybe we could do a rolling version where we always keep the last 2 versions.

svpernova09:
Thanks for the PR!

browner12:
glad to help. i'll try and get to the docs this evening.

browner12:
https://github.com/laravel/docs/pull/3923

yifanfu:
Hi browner12,
Thanks for the PR, can I ask how can i use the elasticsearch in the homestead. I tried port forwarding 62000 to 9200 but it doesn't work. thanks in advance.

browner12:
please follow the docs: https://laravel.com/docs/5.6/homestead#installing-elasticsearch

you'll have to figure the port forwarding out on your own.  shouldn't need it by default.

PR #770, opened 2017-12-21
Add `hirak/prestissimo` composer plugin

andrey-helldar:
See more: https://github.com/hirak/prestissimo


browner12:
what is the ELI5 of this package? no where does it mention what it does.

svpernova09:
Thanks for the PR, I'd like to know more about this package, but we wouldn't implement it here, we'd implement upstream in Settler.

Issue #611, opened 2017-07-13
Using npm on windows host will cause symlink errors


jonnywilliamson:
Please note that the Homestead issue tracker is reserved for bug reports and enhancements. We are not always able to debug Vagrant, Provider or Operating System issues but will do our best to help. Thank you!

## Versions
  - Vagrant: Run `vagrant -v` to show the version.
Vagrant 1.9.2

  - Provider: Virtualbox, VMWare or Parallels and it's version.
VirtualBox

  - Homestead: Run `homestead --version` to show the version.
5.3.2

## Host operating system
This is the operating system that you run locally.

Windows 10 64 bit.

## Homestead.yaml
```yaml
---
name: homestead7
ip: "192.168.10.10"
memory: 2048
cpus: 1
provider: virtualbox


authorize: /Users/Jonathan/Dropbox/Documents/SSH-Keys/Homestead/homestead.pub

keys:
    - /Users/Jonathan/<snip>/private.key

folders:
    - map: /Code
      to: /home/vagrant/Code

sites:
    - map: new.app
      to: /home/vagrant/Code/new.app/public


databases:
    - homestead

ports:
     - send: 80
       to: 80
     - send: 443
       to: 443
```

## Expected behavior

Before you begin 

#### Make sure you have started vagrant on windows using a normal user. ie, you have NOT selected to `run as administrator`

My npm verison is 5.0.3 (but has existed in all versions)

1) Create a new folder like /code/npmBug

2) Create a `package.json` file in it with the following

```json
{
  "devDependencies": {
    "babel-cli": "^6.24.1"
  }
}
```

3) Save it then run `npm install`

### What should have happened?

```
21:43:08 ✔ vagrant@homestead $ npm install

> fsevents@1.1.2 install /home/vagrant/Code/npmbug/node_modules/fsevents
> node install

npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN npmbug No description
npm WARN npmbug No repository field.
npm WARN npmbug No license field.

added 182 packages in 35.105s

```

## Actual behavior
What actually happened?

```
[~/Code/npmbug]
22:01:19 ✔ vagrant@homestead $ npm install
npm WARN npmbug No description
npm WARN npmbug No repository field.
npm WARN npmbug No license field.
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: node-pre-gyp@0.6.36 (node_modules/fsevents/node_modules/node-pre-gyp):
[...]

PR #609, opened 2017-07-11
Fix npm problems for windows users

jcricaro:
Fix symlinks problem for windows user so they don't have to use `--no-bin-links` evertime they use `npm`.

PR #783, opened 2018-01-01
Enable symlinks on shared folders with VirtualBox

JamShady:
See https://stackoverflow.com/questions/24200333/symbolic-links-and-synced-folders-in-vagrant

Tried and tested on Win 10 Pro. Without this 'fix', npm install fails for a variety of projects due to not being able to create symlinks

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -45,4 +45,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         config.hostmanager.manage_host = true
         config.hostmanager.aliases = settings['sites'].map { |site| site['map'] }
     end
+
+    config.vm.provider "virtualbox" do |v|
+        v.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate/v-root", "1"]
+    end
 end


svpernova09:
Thanks for the PR, however we won't be merging this fix. You can see #609 and #611

I don't know if running as admin is still a viable fix, but would be something to test.

JamShady:
Running as admin alone is not the full fix, it needs this as well - annoying, yes. But unfortunately that is the case.

And, to a certain extent I agree with the principle of not including fixes that only deal with 'edge' cases (I don't think using Windows qualifies as an edge case, but anyways...) it would at least be helpful to include this in a commented out fashion, as opposed to ignoring it completely? At least if it were commented out and properly commented, other users experiencing the same problem would at least have access to an easy fix, instead of having to perform the same searches and experiments all over again every time.

At the end of the day, the point of using Homestead is to make it easier for the end developer... if we're omitting fixes for common scenarios (yes, I'd argue using Windows is a common scenario), then ultimately what is the point of this project?

svpernova09:
@JamShady Fair enough. I've opened a PR on the docs repo ^ to add a note about Windows + Virtualbox symlinks.

svpernova09:
@JamShady Docs PR was merged and is live on the docs site. If you feel like there needs to be more wording feel free to open a PR against the 5.5 branch of the docs repo. Thanks for the feedback!

martindilling:
S*** seeing this earlier could have saved me a ton of frustration, figuring out why homestead kept failing on windows.
It's good it's added to the docs, but it took quite a while to even figure out it had anything to do with symlinks.
If this is a windows thing, why isn't this just in by default? Is it breaking something else?

PR #784, opened 2018-01-02
allow server with no sites config. (redis, elastic...)

Artistan:
Current Vagrantfile fails if you do not have sites configures. this update moves it into the sites area of the homestead.rb, confirming that sites exist first.  

This is useful for creating additional redis, elastic, database servers to mimic a larger environment.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -37,12 +37,4 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     if File.exist? afterScriptPath then
         config.vm.provision "shell", path: afterScriptPath, privileged: false, keep_color: true
     end
-
-    if Vagrant.has_plugin?('vagrant-hostsupdater')
-        config.hostsupdater.aliases = settings['sites'].map { |site| site['map'] }
-    elsif Vagrant.has_plugin?('vagrant-hostmanager')
-        config.hostmanager.enabled = true
-        config.hostmanager.manage_host = true
-        config.hostmanager.aliases = settings['sites'].map { |site| site['map'] }
-    end
 end
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -243,6 +243,14 @@ class Homestead
                     end
                 end
             end
+
+            if Vagrant.has_plugin?('vagrant-hostsupdater')
+                config.hostsupdater.aliases = settings['sites'].map { |site| site['map'] }
+            elsif Vagrant.has_plugin?('vagrant-hostmanager')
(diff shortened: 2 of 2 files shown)


svpernova09:
Thanks for taking the time. I feel like this may be outside of scope for Homestead. Homestead expects you to have http servers. If you're wanting to use multi machine vagrant configurations you should use specific base boxes for those (Redis, elastic, etc). 

At the moment Homestead doesn't support multi machine vagrant and I'm not sure it should.

Artistan:
I beg to differ about not supporting multiple machines in vagrant. Homestead supports building as many named servers as you want.

## You can simply add another `named` Homestead box 
this is useful when you need to have the servers different for whatever reason....

- after.sh installs differently
- env variables
- keys
- memory / cpu resources

## Example of multiple installs
destroy one, bring up another, or both at the same time...

`~/VMs/Configs/My-Dev/Homestead.yaml`
```yaml
---
name: "My-Dev"
hostname: "my-dev"
ip: "192.168.10.10"
memory: 1028
cpus: 1
provider: virtualbox

variables:
    - key: APP_ENV
      value: local
    - key: PHP_IDE_CONFIG
      value: serverName=chuck-dev

authorize: ~/.ssh/id_rsa.pub

keys:
    - ~/.ssh/id_rsa

folders:
    - map: ~/code
      to: /home/vagrant/code
      type: "nfs"

sites:
    - map: my-dev.test
      to: /home/vagrant/code/my-dev/public
      schedule: false
      php: "7.1"
```

`~/VMs/Configs/Your-Dev/Homestead.yaml`
```yaml
---
name: "Your-Dev"
hostname: "your-dev"
ip: "192.168.10.10"
memory: 1028
cpus: 1
provider: virtualbox

variables:
    - key: APP_ENV
      value: local
    - key: PHP_IDE_CONFIG
      value: serverName=chuck-dev

authorize: ~/.ssh/id_rsa.pub

keys:
    - ~/.ssh/id_rsa

folders:
    - map: ~/code
      to: /home/vagrant/code
      type: "nfs"

sites:
    - map: your-dev.test
      to: /home/vagrant/code/your-dev/public
      schedule: false
      php: "7.1"
```

svpernova09:
While you *can* do this with Homestead it's not something I want to support at the project level. IE if a change ever comes through that breaks this functionality it wouldn't be a priority to fix.

Artistan:
Sure. It is not documented as being supported, so I understand you point.
This is a simple update to allow people to use it that way.  Can we do this fix since it does not break anything. Extending the functionality does not mean you have to support it. :)

PR #787, opened 2018-01-07
Don’t add bindfs configs if the share type is not NFS

thecrypticace:
You’ll get errors when trying to access these folders otherwise:

```
ls: reading directory '.': Software caused connection abort
ls: cannot open directory '.': Transport endpoint is not connected
```

I installed bindfs because I was going to start using NFS shares but I had turned NFS off temporarily to test something and this happened.


svpernova09:
Thanks for the PR!

PR #789, opened 2018-01-08
Unique Hostname CN for Root Certificate

Artistan:
Allow for unique hostname configuration to differentiate between root certs.

### DIFF
--- a/scripts/create-certificate.sh
+++ b/scripts/create-certificate.sh
@@ -4,10 +4,10 @@ mkdir /etc/nginx/ssl 2>/dev/null
 
 PATH_SSL="/etc/nginx/ssl"
 
-# Path to the custom Homestead Root CA certificate.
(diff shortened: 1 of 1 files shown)


browner12:
is this because you have multiple Homestead installs on one machine?

svpernova09:
The root certificates are what people are pulling out of the vagrant box and importing into their browsers?

if so we should make it 1000% absolutely clear it's a HOMESTEAD generated certificate. So I'm not against naming this the host name of the machine, but I'd want to also append `_HOMESTEAD`.

browner12:
good idea.

Artistan:
@browner12 
Yes, this is due to multiple boxes, plus it is more descriptive for looking at your certs.
I have also automated my cert addition and destruction with shell scripts via up and destroy processes using the unique naming conventions.

svpernova09:
Thanks for the PR!

PR #791, opened 2018-01-11
Update the SilverStripe Nginx config

TheBnl:
Updated the SilverStripe Nginx config to be inline with the new version 4 see https://docs.silverstripe.org/en/4/getting_started/installation/how_to/configure_nginx/ 

I can create a new serve-silverstripe3.sh file to keep support for legacy versions.


svpernova09:
Thanks for the PR!

Issue #795, opened 2018-01-17
Projects running under PHP 5.6


andrey-helldar:
## Versions
  - Vagrant: 2.0.1
  - Provider: Virtualbox
  - Homestead: v7.0.1
  - Laravel: 5.4.36
  - PHP Version: 5.6.32

## Host operating system
Windows 10 Pro

## Homestead.yaml
```yaml
---
ip: "192.168.10.10"
memory: 2048
cpus: 1
provider: virtualbox

authorize: ~/.ssh/pub

keys:
    - ~/.ssh/private.ppk

folders:
    - map: c:/OpenServer/domains
      to: /home/vagrant/code

sites:
    - map: mysite.loc
      to: /home/vagrant/code/mysite.dev/public
      subdomain: promo.mysite.loc
      php: "5.6"

databases:
    - helpers
    - main
```

## Actual behavior
The project works under PHP5.6. After updating the dependencies, an error occurred:
![2018-01-17 09-36-05 whoops there was an error - google chrome](https://user-images.githubusercontent.com/10347617/35030547-5fda6b96-fb70-11e7-8130-1c28f8e4e05b.png)

## Steps to reproduce
The error occurs if I access the `apiResource` method in `routes/api.php`:
```php
app('router')
    ->apiResource('orders', 'Api\OrdersController', [
        'only' => ['show', 'store'],
    ]);
```

If you remove the use of this method, the error disappears.

Next, in the file `composer.lock` found the following dependencies:
```
"name": "doctrine/annotations",
"version": "v1.6.0",
"php": "^7.1"

"name": "doctrine/collections",
"version": "v1.5.0",
"php": "^7.1"

"name": "doctrine/inflector",
"version": "v1.3.0",
"php": "^7.1"

"name": "doctrine/instantiator",
"version": "1.1.0",
"php": "^7.1"

"name": "symfony/css-selector",
"version": "v4.0.3",
"php": "^7.1.3"

"name": "symfony/event-dispatcher",
"version": "v4.0.3",
"php": "^7.1.3"

"name": "symfony/yaml",
"version": "v4.0.3",
"php": "^7.1.3"

"name": "phpdocumentor/reflection-docblock",
"version": "4.2.0",
"php": "^7.0",

"name": "phpunit/php-token-stream",
"version": "2.0.2",
"php": "^7.0"
```

If I use routes:
```php
app('router')
    ->get('orders/{orders}', 'Api\OrdersController@show')
    ->name('orders::show');

[...]

PR #797, opened 2018-01-17
Added a `composer56` alias for working with PHP5.6

andrey-helldar:
This solves the problem described in Issue #795

### DIFF
--- a/resources/aliases
+++ b/resources/aliases
@@ -4,7 +4,6 @@ alias ...="cd ../.."
 alias h='cd ~'
 alias c='clear'
 alias art=artisan
-alias composer56=composer56
 
 alias phpspec='vendor/bin/phpspec'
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for taking the time, going to hold off as the normal workflow would be to use the existing php aliases to swap CLI versions as needed described https://github.com/laravel/homestead/issues/795#issuecomment-358302056

thecrypticace:
Something else you can do to execute composer with a given PHP version in a one off manner without needing to deal with the full paths:
```
php7.1 `which composer` update
```

I do this rather often for projects that don't yet support PHP 7.2.

andrey-helldar:
I added myself an alias `composer56` and use it. Personally, it seems to me more convenient to use.

PR #382, opened 2016-12-06
Ensure Virtualbox connect the LAN cable

JulienTant:
Somehow, Some friends and I are unable to create a working homestead with Virtualbox with just a `vagrant up`, because the LAN adapter has it's cable "disconnected" in VB settings. So we have to `vagrant up` to initialize our box, `vagrant halt` to shut it down, then go to VB settings to force the cable connection, and finally `vagrant up` to make it work. 

This simple line ensure that the cable will be connected in VB and solves our problem.

PR #804, opened 2018-01-30
Enable cable connection on the booted vm 

netcode:
A lot of people had a timeout error when booting the vm , because failure of raising the network interfaces.
`A start job is running for raise network interfaces`

fixing it is to enable cable connection in the vm.


svpernova09:
I'm still waiting on a reproducible bug report that shows this is a Homestead issue: #382 

I've encountered issues with host systems using this change that frequently change between Wifi and Ethernet so I would rather not add this change and cause more issues.

PR #810, opened 2018-02-07
🚑 Don't use the Z-Ray Extension unless the user has a site using it.

svpernova09:


santoshachari:
Hi @svpernova09, I am using laravel debug bar, which seems to be calling z-ray. Any way to keep it working on this commit: 8dc198ecb007e3515d3343997c313623eb612ae8? 
screen shot 2018-02-07 at 11 47 02 pm

svpernova09:
@santoshachari 

I added Laravel debug bar to my test application and it worked just fine without Z-Ray. With Z-Ray it might cause some issues. If you're still having problems try upgrading the homestead repo (now @ version 7.1.2) and destroy & up.

PR #813, opened 2018-02-08
Update composer.lock

andrey-helldar:


svpernova09:
Thanks for the PR but we'd need to run this at a minimum of PHP 7.0. 

What was the purpose of updating?

andrey-helldar:
@svpernova09 When updating homestead, the documentation says to execute the commands:
```
vagrant box update
git pull origin master
composer update
```

I use a desktop application `Git Extensions` to work with repositories.

When executing the `composer update` command, the script recreates the `composer.lock` file and does not add it to the local repository. Therefore, when I load new changes, I need to manually confirm the replacement of the `composer.lock` file with the one loaded from the repository. It is not comfortable.

Another situation:

For example, I want to establish 100% working dependencies, tested by the developers of the package. To do this, run the command `composer install`. If the file is old, it will install the old dependencies. This can be considered an error.

Therefore, in my opinion, the best version of the update will be the execution of the command `composer install`.

PS: I have several versions of PHP installed on my computer. I can send a PR with dependencies for the version of PHP 7.0. Send it?

svpernova09:
You only need to composer update if you're using Homestead per project.

andrey-helldar:
I did not know. I'm using globally installed a Homestead for few projects. Thanks!

PR #824, opened 2018-03-06
✨ Allow user to specify explicit version of ES to install

svpernova09:
Must specify a valid major version or valid version or the install process will blow up with errors about not being able to find the specified version.

/cc @browner12


browner12:
LGTM.

thanks for the update!

PR #845, opened 2018-04-22
Certificate generation fix

mejta:
Certificate for a website was generated as SHA-1 instead of SHA-256. This fix ensure, that correct certificate is generated every time.


svpernova09:
Thanks!

PR #846, opened 2018-04-28
Update aliases

matthewtrask:
Added some common aliases for Yarn/NPM utilities.


browner12:
I feel like these should be left up to the user's discretion, and added to the aliases file as needed.

svpernova09:
> I feel like these should be left up to the user's discretion, and added to the aliases file as needed.

The beautiful thing about aliases is we copy them by default but don't version control them. So the user is free to nuke / add whatever they want so adding to the list isn't much of a consequence so long as we're making sure nothing we're adding will collide with other app names, etc.

svpernova09:
Thanks for the PR

PR #858, opened 2018-05-11
Fix the file permission of the install scripts.

ideepblue:
To maintain all script files in consistency, add execute permission to `scripts/install-minio.sh` and `scripts/install-neo4j.sh`.


svpernova09:
Thanks!

PR #865, opened 2018-05-16
Define types for "apigility", "expressive", and "zf" applications

weierophinney:
Adds the script `serve-zf.sh`, which sets up an nginx host for use with any of Apigility, Expressive, or zend-mvc applications, as they all use the same setup.

In each case, developers should do the following in `Homestead.yaml`:

- create a `folders` mapping that maps the application root directory to a directory in the vagrant box.
- add a site that
  - defines a "to" setting that appends "/public" to the "to" mapping from the above step.
  - uses a "type" of either:
    - "apigility"
    - "expressive"
    - "zf"

The patch updates `scripts/homestead.rb` to add a `case` statement that redefines the `type` if either `apigility`, `expressive`, or `symfony` are detected. In the first two instances, they map to `zf`; the last maps to `symfony2` (which it did previously via a conditional).

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -205,7 +205,12 @@ class Homestead
 
                 type = site["type"] ||= "laravel"
 
-                if (type == "symfony")
+                case type
+                when "apigility"
+                    type = "zf"
+                when "expressive"
+                    type = "zf"
+                when "symfony"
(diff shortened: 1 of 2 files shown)


svpernova09:
Thanks for the PR! Great to see more ZF support :D

PR #880, opened 2018-06-13
Update serve-pimcore.sh

alexpozzi:
Update serve-pimcore.sh script in order to use the default Nginx web-server instead of Apache.
The current script fails the installation because of the 'libapache2-mod-fastcgi' which needs extra repository configurations.
The Nginx configuration used is the one suggested on Pimcore's website (https://pimcore.com/docs/5.x/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/Nginx_Configuration.html).

### DIFF
--- a/scripts/serve-pimcore.sh
+++ b/scripts/serve-pimcore.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 export DEBIAN_FRONTEND=noninteractive
-sudo service nginx stop
 apt-get update
 apt-get install -y php"$5"-bz2
 
@@ -152,4 +151,4 @@ server {
 echo "$block" > "/etc/nginx/sites-available/$1"
 ln -fs "/etc/nginx/sites-available/$1" "/etc/nginx/sites-enabled/$1"
 
-sudo service nginx start
+sudo service nginx restart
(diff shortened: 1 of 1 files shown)


scripts/serve-pimcore.sh
```diff
@@ -1,93 +1,155 @@
...
+ln -fs "/etc/nginx/sites-available/$1" "/etc/nginx/sites-enabled/$1"
 
-sudo a2enmod ssl rewrite proxy proxy_fcgi
-service apache2 restart
-
-if [ $? == 0 ]
-then
-    service apache2 reload
-fi
+sudo service nginx start
```
svpernova09: This should be a restart: `sudo service nginx restart`

alexpozzi: I didn't use restart because I'm stopping the service at the beginning of the script (line 4).
Do you still want me to change it to restart?

svpernova09: You can remove the nginx stop at the top of the script. since we're not using apache we can assume nginx is running and will want to restart it at the end.

svpernova09:
Thanks for the PR!

PR #881, opened 2018-06-13
Adding mysqldump.sql.gz to .gitignore

piotr-cz:
After #542, it's possible to dump MySQL database.
By default dump is saved to `vagrant/mysqldump.sql.gz` and so this file should be ignored IMO

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ phpunit.xml
 /Homestead.json
 /after.sh
 /aliases
+/mysqldump.sql.gz


svpernova09:
Thanks for the PR!

thecrypticace:
Heh. I guess I never thought to add it to the ignore. Good idea! 👍👍

piotr-cz:
I'm updating homestead versions using GIT so that's why this file shows up

PR #883, opened 2018-06-15
Add support for database backup before destroying vm

leo108:
Usage:

```
backup: true
```

All databases listed in `databases` will be back up.

Backup files will be placed in `${Homestead}/${db_type}_backup/${db_name}-${timestamp}.sql`

`vagrant-triggers` plugin is required if vagrant version < 2.1.0

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -410,5 +410,29 @@ class Homestead
             s.args = [settings["ip"]]
             s.privileged = false
         end
+
+        if settings.has_key?("backup") && settings["backup"] && (Vagrant::VERSION >= '2.1.0' || Vagrant.has_plugin('vagrant-triggers'))
+            dirPrefix = '/vagrant/'
+            settings["databases"].each do |database|
+                Homestead.backupMysql(database, "#{dirPrefix}/mysql_backup", config)
+                Homestead.backupPostgres(database, "#{dirPrefix}/postgres_backup", config)
+            end
+        end
+    end
+
+    def Homestead.backupMysql(database, dir, config)
+        now = Time.now.strftime("%Y%m%d%H%M")
+        config.trigger.before :destroy do |trigger|
+            trigger.warn = "Backing up mysql database #{database}..."
+            trigger.run_remote = {"inline": "mkdir -p #{dir} && mysqldump #{database} > #{dir}/#{database}-#{now}.sql"}
+        end
+    end
+
+    def Homestead.backupPostgres(database, dir, config)
+        now = Time.now.strftime("%Y%m%d%H%M")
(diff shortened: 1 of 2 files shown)


svpernova09:
I like the idea. Couple of comments:

Backing up should be an all or nothing thing, I shouldn't have to specify database system or database. The `Homestead.yaml` knows about all the databases that exist so we should use that as a template for what DBs to backup.

Paths shouldn't matter, by default we use save the backups to `/vagrant/*` which will be the Homestead (or project) root folder. From here we could build out a folder structure of `/$databaseType/$databaseName-$timestamp.ext` but I think just dropping files into `/vagrant/*` would suffice.

These two things should allow us to simply specify a top level DB backup in `Homestead.yaml`

leo108:
@svpernova09 What I thought is we may support more types e.g. `mongodb` / `elasticsearch` in the future.

The `databases` field can be removed, I'll change this.

leo108:
@svpernova09 I have updated my commit, please take a look, thanks.

scripts/homestead.rb
```diff
@@ -410,5 +410,41 @@ def Homestead.configure(config, settings)
             s.args = [settings["ip"]]
             s.privileged = false
         end
+
+        if settings.has_key?("backup") && (Vagrant::VERSION >= '2.1.0' || Vagrant.has_plugin('vagrant-triggers'))
+            dirPrefix = '/vagrant/'
+            settings["backup"].each do |backup|
```
svpernova09: Instead of `settings["backup"].each` we should do `settings["databases"].each`. We don't want users to have to configure *which* databases to backup. it should back up all specified in settings, and any possible type (Although I'm find limiting to MySQL, MariaDB, and Postgres for the scope of this PR

scripts/homestead.rb
```diff
@@ -410,5 +410,41 @@ def Homestead.configure(config, settings)
...
+                type = backup["type"]
```
svpernova09: Backup type should be irrelevant because we created mysql/maria/postgres DBs for everything in `settings["databases"]`

scripts/homestead.rb
```diff
@@ -410,5 +410,41 @@ def Homestead.configure(config, settings)
...
+                    Homestead.backupPostgres(settings["databases"], "#{dirPrefix}/postgres_backup", config)
+                else
+                    STDERR.puts "Unsupported backup type: #{type}"
+                    next
+                end
+            end
+        end
+    end
+
+    def Homestead.backupMysql(databases, dir, config)
```
svpernova09: Big +1 for breaking these out into methods. we should probably do more of this.

svpernova09:
Fantastic. Thanks!

PR #885, opened 2018-06-22
Support for timeseries database / visualization

khusmann:
InfluxDB: https://docs.influxdata.com/influxdb/v1.5/
Chronograf: https://docs.influxdata.com/chronograf/v1.5/
Grafana: https://grafana.com/

You may be less interested in adding Chronograf / Grafana support, let me know and I can remake the pull request with just InfluxDB. (This is my first time trying pull requests; please let me know if this is the proper approach)

Cheers!

### DIFF
--- a/scripts/install-chronograf.sh
+++ b/scripts/install-chronograf.sh
@@ -10,8 +10,11 @@ fi
 
 touch /home/vagrant/.chronograf
 
-wget -q https://dl.influxdata.com/chronograf/releases/chronograf_1.5.0.1_amd64.deb
-sudo dpkg -i chronograf_1.5.0.1_amd64.deb
+chronourl="https://dl.influxdata.com/chronograf/releases/chronograf_1.5.0.1_amd64.deb"
+
+wget -q -O chronograf.deb $chronourl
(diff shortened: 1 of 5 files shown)


scripts/install-chronograf.sh
```diff
@@ -0,0 +1,18 @@
...
+then
+    echo "Chronograf already installed."
+    exit 0
+fi
+
+touch /home/vagrant/.chronograf
+
+wget -q https://dl.influxdata.com/chronograf/releases/chronograf_1.5.0.1_amd64.deb
+sudo dpkg -i chronograf_1.5.0.1_amd64.deb
+
```
svpernova09: Do we need to `rm chronograf_1.5.0.1_amd64.deb`? We wouldn't want the script to leave the artifact.

khusmann:
You're totally right -- here's a fix!

svpernova09:
Thanks for the PR!

scripts/install-chronograf.sh
Quix0r: And maybe move to it `/tmp/ instead of current directory.

PR #889, opened 2018-07-04
Switch php tools' versions also while switching

cangelis:
Switch aliases currently only change the php binary's version. Tools also need to be switched to work as expected with the selected php version.


svpernova09:
Thanks!

PR #908, opened 2018-07-23
Add Z-Ray token support

zsuraski:
Add support for storing the token in the .yaml file, and also make it easier for people to discover and enable Z-Ray for the default site...

(will follow up with more detailed info in a separate email).

### DIFF
--- a/resources/Homestead.yaml
+++ b/resources/Homestead.yaml
@@ -16,6 +16,7 @@ folders:
 sites:
     - map: homestead.test
       to: /home/vagrant/code/public
+#     zray: "true"
 
 databases:
     - homestead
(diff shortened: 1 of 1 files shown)


zsuraski:
I imagined that you'd be unlikely to accept the comments part as I indeed noticed there weren't any comments anywhere else but gave it a shot anyway :)

Currently there's no independent URL where you can register/preregister, you can only do that from within Z-Ray under Homestead.  Do you think it's important to have one?

svpernova09:
> I imagined that you'd be unlikely to accept the comments part as I indeed noticed there weren't any comments anywhere else but gave it a shot anyway :)

:D

> Currently there's no independent URL where you can register/preregister, you can only do that from within Z-Ray under Homestead. Do you think it's important to have one?

If they can do it from the Z-Ray screen I think that would be fine.

svpernova09:
Thanks for the PR!

PR #914, opened 2018-07-28
Add nah alias

nikolaynizruhin:
This PR adds `nah` alias which means `git reset --hard;git clean -df`


svpernova09:
Thanks for the PR but we're going to hold off on this for now. I'd prefer to keep the Homestead default aliases tailored to shortcuts around around tooling we provide and while this does add a git alias I don't want to encourage a PR war adding specific aliases.

You're free to add this to your own `aliases` file and commit it since Homestead will never overwrite this file if it exists unless your run `init.sh` or `init.bat` again.

nikolaynizruhin:
Ok, no problem 🙂

PR #915, opened 2018-07-30
Add support for Apache Multiple PHP version with proxy_fcgi

jn-jairo:
This PR adds support for Apache Multiple PHP version using `proxy_fcgi` when `fastcgi` is not available.

### DIFF
--- a/scripts/serve-apache.sh
+++ b/scripts/serve-apache.sh
@@ -33,6 +33,13 @@ block="<VirtualHost *:$3>
         Alias /php"$5"-fcgi /usr/lib/cgi-bin/php"$5"
         FastCgiExternalServer /usr/lib/cgi-bin/php"$5" -socket /var/run/php/php"$5"-fpm.sock -pass-header Authorization
     </IfModule>
+    <IfModule !mod_fastcgi.c>
+        <IfModule mod_proxy_fcgi.c>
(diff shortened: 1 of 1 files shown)


svpernova09:
`fastcgi` should always be available since it's included in the base box. How are you running into an issue where it's not?

jn-jairo:
The steps I made:

```bash
vagrant box add laravel/homestead

git clone https://github.com/laravel/homestead.git ~/Homestead

cd ~/Homestead

git checkout v7.14.0

bash init.sh
```

In the `Homestead.yaml` I used `virtualbox` and I add my site:

```yaml
folders:
    - map: ~/code
      to: /home/vagrant/code

sites:
    - map: mysite.test
      to: /home/vagrant/code/mysite/public
```

The `nginx` was causing some headers errors in the turbolinks so I changed to `apache` and it worked, but the code was running with php 7.2 so I set the config:

```yaml
sites:
    - map: mysite.test
      to: /home/vagrant/code/mysite/public
      type: "apache"
      php: "7.1"
```

And run:

```bash
vagrant reload --provision
```

But I still with php 7.2, after some debugging I found that `fastcgi` module wasn't in the server, so I use `proxy_fcgi` as in the commit above.

svpernova09:
I see, thanks for the PR!

PR #925, opened 2018-08-16
update MariaDB to 10.3 stable

it-can:


scripts/install-maria.sh
```diff
@@ -31,20 +31,20 @@ rm -rf /etc/mysql
 # Add Maria PPA
 
 sudo apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xF1656F24C74CD1D8
-sudo add-apt-repository 'deb [arch=amd64,i386,ppc64el] http://ftp.osuosl.org/pub/mariadb/repo/10.2/ubuntu xenial main'
+sudo add-apt-repository 'deb [arch=amd64,i386,ppc64el] http://ftp.osuosl.org/pub/mariadb/repo/10.3/ubuntu xenial main'
```
svpernova09: We should be using `bionic` instead of `xenial`.

it-can: Ok wil change asap

it-can:
@svpernova09 ok changed it

svpernova09:
Thanks!

PR #932, opened 2018-08-31
Proposal: user-specific after.sh files

stevegrunwell:
I'm a huge fan of the `after.sh` convention, which really lets me streamline the setup of an application (I tend to use per-project Homestead installations); picking up one of my Laravel apps usually looks like:

1. Clone the git repo
2. Install Composer dependencies
3. `$ vendor/bin/homestead make`
4. `$ vagrant up` (with `after.sh` handling migrations, seeding, etc.)
5. `$ vagrant ssh`

— then start building awesome stuff!

The `after.sh` script is great to ensure that new VMs automatically get system requirements and let me save developers a step in running migrations/seeders. Of course, developers are an opinionated bunch, so I find myself running this anytime I rebuild the VM:

```sh
# Replace nano with a much better editor 💥
$ sudo update-alternatives --set editor /usr/bin/vim.basic
```

For developers more dependent on their own, more-unique aliases, it's nice to have a way to set user-specific settings automatically in a script that's not tracked under version control.

This PR introduces just that, in the form of an `after.local.sh` file: when present, the `after.local.sh` file will be executed immediately following the `after.sh` file (though the two are not tied to one another). Nothing is generated automatically, but developers who want to be able to fully customize the VM without losing their work with each rebuild, `after.local.sh` is 👌 

## Alternate approach

If the maintainers/community find that this doesn't add enough value, I'll also leave the less-canonical implementation for teams that would benefit from this concept:

```sh
# after.sh

# Load developer-specific after.sh commands.
if [ -e after.local.sh ]; then
    ./after.local.sh
fi
```

### DIFF
--- a/resources/localized/Vagrantfile
+++ b/resources/localized/Vagrantfile
@@ -10,7 +10,7 @@ confDir = $confDir ||= File.expand_path("vendor/laravel/homestead", File.dirname
 homesteadYamlPath = File.expand_path("Homestead.yaml", File.dirname(__FILE__))
 homesteadJsonPath = File.expand_path("Homestead.json", File.dirname(__FILE__))
 afterScriptPath = "after.sh"
-afterLocalScriptPath = "after.local.sh"
+customizationScriptPath = "user-customizations.sh"
 aliasesPath = "aliases"
 
 require File.expand_path(confDir + '/scripts/homestead.rb')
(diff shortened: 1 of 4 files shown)


resources/localized/Vagrantfile
```diff
@@ -38,6 +39,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         config.vm.provision "shell", path: afterScriptPath, privileged: false, keep_color: true
     end
 
+    if File.exist? afterLocalScriptPath then
+        config.vm.provision "shell", path: afterScriptPath, privileged: false, keep_color: true
```
svpernova09: This should be `path: afterLocalScriptPath`

stevegrunwell:
Phew, I wasn't particularly attached to `after.local.sh`, but was having trouble coming up with a better name; `user-customizations.sh` gives it a much clearer meaning 👍 

PR has been updated to use the new name and address the missed copy/paste error in `resources/localized/Vagrantfile`.

svpernova09:
Great! I’m still on coder cruise until Monday afternoon so it may. It be until then I have enough bandwidth to test and merge.

--
- Joe Ferguson
JoeFerguson.me
osmihelp.org

svpernova09:
Thanks for the PR and changes. Looking forward to implementing this in my own projects.

PR #933, opened 2018-08-31
Install the latest version of neo4j rather than a specific version.

ideepblue:
Update the install neo4j script.


svpernova09:
Thanks!

PR #939, opened 2018-09-09
Create minio buckets via config

fppz:
This adds support for provisioning minio/s3 buckets through `buckets:` inside of the config file.

Each bucket requires a `name` and a `policy` inside of the config file.
Acceptable policies include:  `none, download, upload, public` however will default to `none` if not provided

Example inside of my `Homestead.yaml`

```
buckets:
    - name: public-bucket
      policy: public
    - name: none-bucket
```

This is added by using the [minio client](https://docs.minio.io/docs/minio-client-complete-guide), accessible through the `mc` command


svpernova09:
Thanks for the PR!

sebastiaanluca:
Can't really deduct it from the file changes; does this support SSL/HTTPS?

If not, maybe we can add it? For example, we're currently manually setting up Minio:

```bash
…

mkdir -p /home/vagrant/Downloads
wget -O /home/vagrant/Downloads/generate_cert.go https://golang.org/src/crypto/tls/generate_cert.go?m=text

cd /home/vagrant/Downloads
go run /home/vagrant/Downloads/generate_cert.go -ca --host "192.168.116.20"
mv /home/vagrant/Downloads/cert.pem /home/vagrant/.minio/certs/public.crt
mv /home/vagrant/Downloads/key.pem /home/vagrant/.minio/certs/private.key

# Restart Supervisor job, etc
```

PR #945, opened 2018-09-17
Backup mysql functions and procedures

jn-jairo:
This PR makes the backup of mysql database contains functions and procedures.

I don't know if there is a reason for homestead isn't already backuping functions and procedures, but it will be really nice to have this backuped.


svpernova09:
Thanks for the PR!

PR #946, opened 2018-09-23
hyperv : use linked_clone instead of deprecated differencing_disk option

nicolus:
Vagrant would throw a lot of warnings because differencing_disk is a deprecated option. According to the docs linked_clone is the correct way now and does exactly the same thing : https://www.vagrantup.com/docs/hyperv/configuration.html


svpernova09:
Thanks for the update!

PR #949, opened 2018-09-26
Allow config overrides for per-project installs

fitztrev:
In per-project installs, it's [common for the `Homestead.yaml` to be under version control](https://github.com/search?utf8=%E2%9C%93&q=filename%3AHomestead.yaml+path%3A%2F&type=).

This allows each environment to change some of the settings via a `Homestead.override.yaml` file which can be gitignored.

Reason I wanted this was so I can have 2+ boxes of the same project going at once on my machine, but since each needed a different IP address, I would have had to edit the `Homestead.yaml` and have uncommitted changes.

### DIFF
--- a/resources/localized/Vagrantfile
+++ b/resources/localized/Vagrantfile
@@ -9,6 +9,8 @@ confDir = $confDir ||= File.expand_path("vendor/laravel/homestead", File.dirname
 
 homesteadYamlPath = File.expand_path("Homestead.yaml", File.dirname(__FILE__))
 homesteadJsonPath = File.expand_path("Homestead.json", File.dirname(__FILE__))
+homesteadOverrideYamlPath = File.expand_path("Homestead.override.yaml", File.dirname(__FILE__))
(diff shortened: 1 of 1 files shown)


svpernova09:
Going to hold off on this because the already existing solution of copying your `Homestead.yaml` to `Homestead.yaml.example` is how this should be accomplished. The `make` command parses `Homestead.yaml.example` and uses all values found there *except* folder mapping. Folder mapping is unique to the individual user's machine since the `make` command figures out where it's being run from so that it can place a full local path to the current project.

If I recall from our chat in Discord you're mapping `.` which is not recommended by Vagrant because when you map `.` Vagrant assumes you know what you're doing and doesn't map `.` to `/vagrant` which is how the initial provisioning tooling is executed in the VM (From `/vagrant` path).

The better solution is to add documentation around `Homestead.yaml.example` and the `make` command and encourage users to not map `.` and to not version control `Homestead.yaml`, but instead version control `Homestead.yaml.example`.

Thanks for the discussion and taking the time to build the PR.

PR #950, opened 2018-09-28
Fix error message for missing Homestead settings

fitztrev:
For localized Homestead installs, the path to the `Homestead.yaml/.json` is expected to be the same as the `Vagrantfile`, as seen in `homesteadYamlPath` and `homesteadJsonPath`. This fixes the error message.


svpernova09:
Thanks!

Issue #548, opened 2017-04-21
How to specify port on proxy pass of nginx


dqlopez:
Trying to figure out on documentation but cannot seem to find it. How can I specify the port that will be serve here `proxy_pass http://127.0.0.1:${2};`?

## Homestead.yaml
```
---
ip: "192.168.10.10"
memory: 512
cpus: 1
provider: virtualbox

authorize: ~/.ssh/id_rsa.pub

keys:
    - ~/.ssh/id_rsa

folders:
    - map: ~/Web
      to: /var/www/html
      type: "nfs"

sites:
    - map: test.local
      to: /var/www/html/test
      type: proxy
```
With the YAML above, it will generate this error on nginx during provision:
```
Job for nginx.service failed because the control process exited with error code. See "systemctl status nginx.service" and "journalctl -xe" for details.
```
And looking at the generated sites-available, the port on **proxy_pass** has no value which cause the error:
```
location / {
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection upgrade;
        proxy_set_header Host $host;
        proxy_http_version 1.1;
        proxy_pass http://127.0.0.1:/var/www/html/test;
    }
```

svpernova09:
From the `aliases` file:

```
function serve-proxy() {
    if [[ "$1" && "$2" ]]
    then
        sudo dos2unix /vagrant/scripts/serve-proxy.sh
        sudo bash /vagrant/scripts/serve-proxy.sh "$1" "$2" 80
    else
        echo "Error: missing required parameters."
        echo "Usage: "
        echo "  serve-proxy domain port"
    fi
}
```

You should be using `serve-proxy domain port`, where the port is the `${2}` value.

If you are using Homestead per project, you may be missing your `aliases` file. If so just run`php vendor/bin/homestead make --aliases`.

You can verify your aliases are working by `vagrant ssh` and inspecting the `.bash_aliases` file in `/home/vagrant`

dqlopez:
I am not using Homestead per project. Where is the value of `${2}` being fetched?

svpernova09:
We pass the value here: https://github.com/laravel/homestead/blob/master/scripts/homestead.rb#L188-L189 You need to ensure you're specifying a port on your site:

```
sites:
    - map: test.local
      to: /var/www/html/test
      type: proxy
[...]

PR #951, opened 2018-09-28
Add serve-ngrok.sh

cwilby:
See https://github.com/laravel/homestead/issues/548.

### DIFF
--- a/resources/localized/aliases
+++ b/resources/localized/aliases
@@ -94,6 +94,18 @@ function serve-proxy() {
     fi
 }
 
+function serve-ngrok() {
+    if [[ "$1" && "$2" && "$3" ]]
+    then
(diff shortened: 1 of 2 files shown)


svpernova09:
`server-*.sh` scripts are for generating virtualhosts based on different site types. There shouldn't be a site type for ngrox since the share command uses ngrok. What are you trying to solve here?

From #548 it sounds like you're saying there's something wrong with the proxy script?

cwilby:
The proxy script works as intended (forwards ngrok requests using `proxy_pass=127.0.0.1:${2}`), I was looking for ways to replace `proxy_pass=127.0.0.1` with the URL used under the sites section in Homestead.yaml.

However, the share command solves the issue. Closing the PR.

PR #953, opened 2018-09-30
Fix CPU soft lockup when audio is enabled

Swanty:
For some weird reason on my laptop I am getting CPU "soft lockup" errors during boot and boot time becomes extremely long.

Once I disabled audio homestead booted up really fast.

I can't explain why this is happening, but I'm sure no one needs the audio enabled either way right? =)


svpernova09:
I'm not sure how I feel disabling this to default off. It's not like I've ever had the need to play audio from my Homestead box, but feels odd. Going to play with the audio settings a bit.

Thanks for sharing.

PR #955, opened 2018-10-02
Add support for custom headers in sites

jordanramstad:
First my use case here was that I had a separate domain for images, so in homestead I set up a separate site to handle images specifically `images.homestead.test` for example. The site I was using was an api/spa setup as well, so the client was separated from the homestead box (a `nuxt` app).

However I also needed to use JavaScript in some cases to convert the images to canvas for cropping, but this will trigger CORS issues. To get around this I needed to add headers to the images to allow it, but passing the images through PHP in order to add the headers is a little overkill so doing it in nginx made sense.

That is what this update brings in, right now its more of a feature request rather then fixing anything. I also have only applied it to the default site type (laravel). It will add headers with some settings in the homestead.yaml like so:

```
sites:
    -
        map: image.homestead.test
        to: /home/vagrant/code/public/storage
        headers: 
            -
                key: "Access-Control-Allow-Origin"
                value: "*"
```

Let me know what you think and I will apply it to the other site types, otherwise I am open to any ideas.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -230,8 +230,15 @@ class Homestead
             end
             params += ' )'
           end
+          if site.include? 'headers'
+            headers = '('
+            site['headers'].each do |header|
+                headers += ' [' + header['key'] + ']=' + header['value']
+            end
+            headers += ' )'
+          end
           s.path = script_dir + "/serve-#{type}.sh"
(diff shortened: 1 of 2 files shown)


svpernova09:
Thanks! Going to merge and poke around.

jordanramstad:
Another thought is that adjusting headers usually might have a specific purpose and it is local (so security stuff is less important, focusing on ease of development)

So might be able to simplify to this
```
sites:
    -
        map: image.homestead.test
        to: /home/vagrant/code/public/storage
        cors: true
```

This would effectively just ensure cors headers are added to requests, since I cannot think of other things that might be needed to adjust headers for from the nginx/apache level. This option would just lower flexibility but make it easier to configure.

PR #957, opened 2018-10-05
Add Load Balancer

fppz:
## How to enable (Homestead.yaml)
```
load_balancer: true
```

I'm not sure if this would be overkill or not.
This would essentially add a load balancer in front of all request through nginx assuming you don't set a port inside one of the `sites`.
Downside to this is that you can't use the `apache` type.

I've been doing a lot of work that gets put behind a load balancer and was looking to have my local dev environment match as close to production.


svpernova09:
Thanks!

PR #964, opened 2018-10-22
Add verify_host_key config

dolcy:
The following deprecation error message may appear in some Vagrant installs:
```
verify_host_key: false is deprecated, use :never
```
Adding a config for verify_host_key addresses this issue.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -10,7 +10,7 @@ class Homestead
     # Allow SSH Agent Forward from The Box
     config.ssh.forward_agent = true
       
-    # Configure Verfiy Host Key
+    # Configure Verify Host Key
     config.ssh.verify_host_key = :never
 
     # Configure The Box
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -9,6 +9,9 @@ class Homestead
 
     # Allow SSH Agent Forward from The Box
     config.ssh.forward_agent = true
+      
+    # Configure Verfiy Host Key
+    config.ssh.verify_host_key = :never
 
     # Configure The Box
     config.vm.define settings['name'] ||= 'homestead-7'
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -11,7 +11,13 @@ class Homestead
     config.ssh.forward_agent = true
       
     # Configure Verify Host Key
-    config.ssh.verify_host_key = :never
+    config.ssh.verify_host_key = true
+
+    if config.ssh.verify_host_key == nil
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for taking the time.

Instead of just disabling this for everyone, it should be a configurable setting. See the `backup` usage in `scripts/homestead.rb` for implementation information. It should default to ON, if set it should change to use this setting.

dolcy:
Thanks @svpernova09 - agreed; will change.

scripts/homestead.rb
```diff
@@ -9,6 +9,9 @@ def self.configure(config, settings)
 
     # Allow SSH Agent Forward from The Box
     config.ssh.forward_agent = true
+      
+    # Configure Verfiy Host Key
```
browner12: spelling: 'Verfiy' --> 'Verify'

scripts/homestead.rb
```diff
@@ -9,6 +9,15 @@ def self.configure(config, settings)
 
     # Allow SSH Agent Forward from The Box
     config.ssh.forward_agent = true
+
```
browner12: can we simplify this to use our default in the `else` conditional? also, we need to use input from the settings.  not sure about the naming though....

```suggestion
    # Configure Verify Host Key
    if settings.has_key?('verify_host_key')
      config.ssh.verify_host_key = settings['verify_host_key']
    else
      config.ssh.verify_host_key = true
    end
```

dolcy:
Thanks @browner12

scripts/homestead.rb
```diff
@@ -9,6 +9,14 @@ def self.configure(config, settings)
...
     # Allow SSH Agent Forward from The Box
     config.ssh.forward_agent = true
+    
+    # Configure Verify Host Key
+    if settings.has_key?('verify_host_key')
+      config.ssh.verify_host_key = settings['verify_host_key']
+    else
+      config.ssh.verify_host_key = true || :accept_new_or_local_tunnel
+    end    
+    config.ssh.verify_host_key = :accept_new_or_local_tunnel
```
svpernova09: i believe we want to remove this line or else it will always set the value

scripts/homestead.rb
```diff
@@ -9,6 +9,11 @@ def self.configure(config, settings)
 
     # Allow SSH Agent Forward from The Box
     config.ssh.forward_agent = true
+    
+   # Configure Verify Host Key
```
browner12: 4 spaces

```suggestion
    # Configure Verify Host Key
```

svpernova09:
Thanks!

PR #979, opened 2018-11-14
Fix mysql 8 authentication plugin

raguilar:
The previous implementation was just echoing the string to stdout, this ensures that it reaches `/etc/mysql/conf.d/mysql.cnf`

Also while testing this i found that this requires a restart before creating the users.


svpernova09:
Thanks!

Issue #978, opened 2018-11-14
.editorconfig should set indent_size = 2 for Ruby files


hagabaka:
The project's .editorconfig file sets a default `indent_size` of 4 for all files, and adds an exception for *.yml files. It should also add an exception for *.rb files, since scripts/homestead.rb uses 2-space indentation.

svpernova09:
Thanks for opening an issue!

While Ruby syntax *does* specify 2 spaces instead of 4 I have no interest in changing this. (I'm not a fan at all of 2 spaces).

hagabaka:
However scripts/homestead.rb already uses indentation of 2 spaces.

It wouldn't involve changing indentation of other files, just adding overrides for .rb files in .editorconfig.

Issue #910, opened 2018-07-24
"no implicit conversion of Integer into String" error when configuring a proxy site type


stevegriffiths:
## Versions
  - Vagrant: 2.1.1
  - Provider: Virtualbox 5.2.16
  - Homestead: 7.8.0

## Host operating system
Windows 10

## Homestead.yaml
---
```yaml
ip: "192.168.10.10"
memory: 2048
cpus: 1
provider: virtualbox

authorize: ~/.ssh/id_rsa.pub

keys:
    - ~/.ssh/id_rsa

folders:
    - map: C:\Users\xx\Documents\Websites
      to: /home/vagrant/code

sites:
    - map: homestead.phpmyadmin
      to: /home/vagrant/code/phpmyadmin
    - map: homestead.app1
      to: /home/vagrant/code/app1/public
    - map: homestead.nodejs1
      to: 8080
      type: proxy

databases:
    - homestead

# blackfire:
#     - id: foo
#       token: bar
#       client-id: foo
#       client-token: bar

#ports:
#    - send: 8081
#      to: 8080
#      protocol: udp
```

## Vagrant destroy & up output
Provide a link to a [GitHub Gist](https://gist.github.com/) containing the complete output of `vagrant destroy && vagrant up`.
Do **NOT** paste the debug output in the issue, just paste the link to the Gist.

## Expected behavior
Expect vagrant up to run without error, and for one of the sites in the resulting VM to act as a proxy to port 8080 (running a Node.js app) within the machine.

## Actual behavior
Vagrant up fails with the following error:

`homestead.rb:233:in '+': no implicit conversion of Integer into String (TypeError).`

I have narrowed this error down to being the result of this line in the Homestead.yaml file, despite various sources suggesting that this is the correct format for defining a proxy type site in homestead:

`to: 8080`

## Steps to reproduce
1. Create Homestead.yaml file as displayed above.
2. Run vagrant up

svpernova09:
Have you tried putting quotes around 8080?

Make sure ports is uncommented, you have a # in front of it.

--
- Joe Ferguson
JoeFerguson.me
MemphisPHP.org

stevegriffiths:
Hi,

I am not using the ports section - I had experimented with it previously but really need to get the application working as a proxy - it's the config in the sites section I am focused on, specifically the block:
`
    - map: homestead.nodejs1
      to: 8080
      type: proxy
`

[...]

PR #982, opened 2018-11-16
Fix two issues related to homestead.rb #910 and #978

hagabaka:
#910 and #978

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,3 +14,7 @@ trim_trailing_whitespace = false
 [*.yml]
 indent_style = space
 indent_size = 2
+
+[*.rb]
+indent_style = space
+indent_size = 2
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -265,8 +265,11 @@ class Homestead
               s.inline = 'ln -sf /opt/zray/zray.ini /etc/php/7.2/fpm/conf.d/zray.ini'
             end
           else
-            config.vm.provision 'shell' do |s|
+            # For proxy type sites, "to" is a port number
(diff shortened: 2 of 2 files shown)


scripts/homestead.rb
```diff
@@ -265,8 +265,11 @@ def self.configure(config, settings)
               s.inline = 'ln -sf /opt/zray/zray.ini /etc/php/7.2/fpm/conf.d/zray.ini'
             end
           else
-            config.vm.provision 'shell' do |s|
+            # For proxy type sites, "to" is a port number
+            if site['to'].is_a? String
```
roberto-aguilar: Do nothing when the port is not a string is incorrect, it is going to silently fail and lead to unexpected behavior. What if instead you cast always to string?

I have tried this with `8080` and `"8080"` and work as expected:

```ruby
s.inline = 'rm -rf ' + site['to'].to_s + '/ZendServer'
```

.editorconfig
```diff
@@ -14,3 +14,7 @@ trim_trailing_whitespace = false
 [*.yml]
 indent_style = space
 indent_size = 2
+
+[*.rb]
+indent_style = space
```
roberto-aguilar: The `indent_style = space` is already defined by `[*]` so there is no need to declare it again.

But given that this was previously rejected, why don't you keep this PR more scoped by just fixing #910 and sending this in an independent PR to help with the discussion?

svpernova09:
As commented, a PR to solve the issue of #910 would be welcome, but it looks like we don't need any changes to `.editorconfig`

PR #986, opened 2018-11-23
Test against PHP 7.3

driesvints:
PHP 7.3 is right around the corner so let's make sure Homestead works on it.


driesvints:
Added `|| true` to the xdebug removal line to prevent removing it when it's not installed in the first place. This is also consistent with the other Laravel libraries.

svpernova09:
Thanks!

PR #989, opened 2018-11-25
Fixed .user.ini subdirectory issue

paolooo:
.user.ini does not work on subdirectories due to the extra leading slash. See related stackoverflow:

https://stackoverflow.com/a/41879334/959060

I've removed the trailing slash.


svpernova09:
Thanks

PR #990, opened 2018-11-27
"da" alias for "composer dumpautoload'

hivokas:


browner12:
I think these are better served in your personal `aliases` file.

svpernova09:
Agreed with @browner12, better suited for your `aliases`. Thanks for taking the time to open a PR.

PR #991, opened 2018-11-27
"nah" alias for "git reset --hard; git clean -df;"

hivokas:


svpernova09:
Thanks for sharing, but going to pass on these. You can easily add them to your own `aliases` file.

PR #992, opened 2018-11-27
"t" alias for "artisan tinker"

hivokas:


browner12:
I think these are better served in your personal `aliases` file.

svpernova09:
Thanks for taking the time, going to pass on single character aliases. You can easily add it to your own `aliases` file.

PR #995, opened 2018-11-30
Support SMB credentials in Homestead.yaml

cbj4074:
When using SMB mounts, having to provide the SMB credentials every time the VM boots grows tiresome. This simple change adds the ability to specify those credentials in Homestead.yaml.

It goes without saying that storing these credentials in plaintext poses a security risk, but most developers are using an unprivileged, local account exclusively for this purpose, in which case any associated risk may be acceptable.

I have tested `vagrant up` with and without specifying the credentials and it responds as expected: Vagrant prompts for them if not provided, and it uses them without prompting if they are.

Note that I added support for `smb_host`, too (for details, see: https://www.vagrantup.com/docs/synced-folders/smb.html ).

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -178,6 +178,8 @@ class Homestead
             mount_opts = folder['mount_options'] ? folder['mount_options'] : ['actimeo=1', 'nolock']
           elsif folder['type'] == 'smb'
             mount_opts = folder['mount_options'] ? folder['mount_options'] : ['vers=3.02', 'mfsymlinks']
+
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for sharing this change!

cbj4074:
No problem!

I should add that the usage is straightforward, and the key names match their Vagrant counterparts, e.g.:

```
folders:
    - map: C:\Users\me\project
      to: /home/vagrant/code/project
      type: smb
      smb_username: homestead
      smb_password: secret
```

Thanks for the quick merge; much appreciated!

PR #996, opened 2018-12-04
Merge SMB creds from Homestead.yaml into config

cbj4074:
While SMB credentials were being parsed, they weren't being merged into the effective configuration; this change implements the intended behavior.

@svpernova09 This was simply an oversight on my part; I had a bunch of WIP code all over the place and had meant to include this in my previous PR. Thanks!


svpernova09:
Thanks for the update!

PR #1000, opened 2018-12-11
Websockets

svpernova09:


lcharette:
FYI, adding the web socket domain create a breaking change. When single domain is defined in `Homestead.yaml`, it used to be reachable using the VM IP address. Adding the `socket-wrench.test` make it so browsing to the VM IP address errors out with the "No Input File" message.

svpernova09:
> FYI, adding the web socket domain create a breaking change

Yep! the next version will be a major version bump.

PR #1005, opened 2018-12-14
Script para instalar OCI8

zemisdev:


svpernova09:
We're not able to support downloading from a random DropBox URL. This is something you should have in your own `aliases` file. The default `after.sh` scripts should not do anything since they are specifically for Homestead users (as opposed to Homestead providing it itself).

We can potentially add this to the base box, but would need to ensure the source download is coming from a verified source.

PR #1007, opened 2018-12-18
Add Sites from Homestead.yaml to /etc/hosts

cbj4074:
While VirtualBox VMs have a means by which to resolve hostnames from the *host* machine's hosts file, the same is not true of other providers, such as Hyper-V (probably due to greater isolation between host and guest).

For example, it's possible to reach (e.g., ping) a hostname defined in the *host* machine's hosts file immediately after provisioning with VirtualBox (on Windows, at least). But with an otherwise identical configuration in Hyper-V, the hostname cannot be resolved.

The primary reason for which I propose this change is that hostname resolution within the VM, for sites defined in `Homestead.yaml`, is required for Dusk testing. Absent the proposed change, Dusk testing does not work on Hyper-V because `homestead.site`, for example, cannot be resolved unless `127.0.0.1 homestead.site` is added to `/etc/hosts`.

Ref: [Laravel Dusk returns empty html document when dumping browser, assertions therefore fail
](https://stackoverflow.com/a/50686795/1772379)

### DIFF
--- a/scripts/serve-laravel.sh
+++ b/scripts/serve-laravel.sh
@@ -82,3 +82,4 @@ block="server {
 
 echo "$block" > "/etc/nginx/sites-available/$1"
 ln -fs "/etc/nginx/sites-available/$1" "/etc/nginx/sites-enabled/$1"
+echo "127.0.0.1 $1" >> /etc/hosts


svpernova09:
We can't merge this as is due to the fact we would always be polluting `/etc/hosts` if a user runs `vagrant provision` which based on many of the issues we see is something our users really like to do.

We need some way to also remove these same entries possibly when we clear the sites such as in https://github.com/laravel/homestead/blob/master/scripts/clear-nginx.sh

svpernova09:
Merged this and have it disabled until we can get a mechanism to reset the file in place. Thanks!

mikeybinns:
Hi, not sure if this helps, but this is how I update my hosts file automatically:
Background: I'm on windows 10, and it requires admin access for the terminal/powershell as admin access is required to write to hosts file. Also requires the vagrant-hostsupdater plugin.

In Homestead.yaml 
```
sites:
    - map: homestead.local
      to: /home/vagrant/websites/homestead
      hosts: 192.168.10.20

hosts_file_additions: true
```

In /scripts/homestead.rb
```
class Homestead
  def self.configure(config, settings)
    # Updating the hosts file with all the sites that are defined in Homestead.yaml
    if Vagrant.has_plugin?("vagrant-hostsupdater") && settings["hosts_file_additions"] == true
      hosts = []
      settings["sites"].each do |site|
      hosts.push(site["map"])
    end
      config.hostsupdater.aliases = hosts
    end
  end
end
```
This method adds the hosts on VM up and removes on suspend. It marks each listing with a machine ID to detect which ones it has added, so it doesn't clear any previously added sites.

Hope this helps

cbj4074:
@mikeybinns Apologies for the late reply.

If I understand correctly, the technique you describe modifies the hosts file on the *host machine* (that is, the Windows 10 machine), correct?

The problem that this PR intends to address is hostname resolution from *within the guest* (VM).

It seems that this problem manifests only on Hyper-V (and not VirtualBox -- haven't tried VMware), so most people probably haven't encountered it, due to relatively low Hyper-V adoption.

In any case, we would need to do the same *type* of thing you describe, but inside the VM, although I'm not sure if the approach would be basically the same -- that is, use Vagrant + Ruby -- or if it would make more sense to do it via provisioning (shell) scripts.

@svpernova09 Would it be as simple as emptying the hosts file (the one inside the VM) when provisioning, just as it's done in the [nginx script you referenced](https://github.com/laravel/homestead/pull/1007#pullrequestreview-186079442)? Or can you think of some unwanted side-effect that would result?

I suppose the real question is: does any other script read from or write to the hosts file, currently?

svpernova09:
> @svpernova09 Would it be as simple as emptying the hosts file (the one inside the VM) when provisioning, just as it's done in the nginx script you referenced? Or can you think of some unwanted side-effect that would result?

You wouldn't want to just nuke the existing hosts file, you'd want to append to it, or possibly even put a unique block in there that you can parse and replace anything after it, that way you don't interfere with the OS dependent bits in the hosts file.

browner12:
not sure if this help, but Local by Flywheel uses this block in the host machine `/etc/hosts` file:

```sh
## Local by Flywheel - Start ##
192.168.75.100 site1.test #Local Site
192.168.75.100 www.site1.test #Local Site
## Local by Flywheel - End ##

When you add and remove sites through their application, they will add and remove these entries.
```

cbj4074:
@svpernova09 True enough! I forgot about the default loopback entries. *smacks forehead*

@browner12 It does... while I doubt we could readily reuse whatever code Flywheel uses to parse the string (even if it's FOSS and licensing permits), I think we'd need to do something similar. The format looks useful, with the start/end lines, and the comment denoting each domain.

Ultimately, this will probably require some fiddling with `sed` or similar. I'll propose something more detailed soon.

Thanks!

Fuxy22:
I personally use the below shell functions in my aliases file to add an remove hosts
```sh
# remove specified host from /etc/hosts
function removehost() {
    if [[ "$1" ]]
    then
        HOSTNAME=$1

        if [ -n "$(grep $HOSTNAME /etc/hosts)" ]
        then
            echo "$HOSTNAME Found in your /etc/hosts, Removing now...";
            sudo sed -i".bak" "/$HOSTNAME/d" /etc/hosts
        else
            echo "$HOSTNAME was not found in your /etc/hosts";
        fi
    else
        echo "Usage: "
        echo "  removehost domain"
    fi
}

#add new ip host pair to /etc/hosts
function addhost() {
    if [[ "$1" && "$2" ]]
    then
        IP=$1
        HOSTNAME=$2

        if [ -n "$(grep $HOSTNAME /etc/hosts)" ]
            then
                echo "$HOSTNAME already exists:";
                echo $(grep $HOSTNAME /etc/hosts);
            else
                echo "Adding $HOSTNAME to your /etc/hosts";
                printf "%s\t%s\n" "$IP" "$HOSTNAME" | sudo tee -a /etc/hosts > /dev/null;

                if [ -n "$(grep $HOSTNAME /etc/hosts)" ]
                    then
                        echo "$HOSTNAME was added succesfully:";
                        echo $(grep $HOSTNAME /etc/hosts);
                    else
                        echo "Failed to Add $HOSTNAME, Try again!";
                fi
        fi
    else
        echo "Usage: "
        echo "  addhost ip domain"
    fi
}
```
and at the end of the `scripts/serve-laravel.sh` i have something like
```sh
if [ -n "$(grep $1 /etc/hosts)" ] # Check is site already exists in hosts
    then
        sudo sed -i".bak" "/$1/d" /etc/hosts # Remove site entry if it does
fi
printf "%s\t%s\n" 127.0.0.1 "$1" | sudo tee -a /etc/hosts > /dev/null;    # Add site to hosts file
```

cbj4074:
@Fuxy22 That works brilliantly! Thank you! Do you mind if I weave it into a revised PR? Or would you prefer to add it so you're credited with the commit (which is perfectly reasonable)?

@svpernova09 I've tested this and it seems to work excellently, both for `vagrant up` and `vagrant reload --provision`. Are you opposed to adding those functions to the `aliases` file and calling them at the bottom of `serve-laravel.sh`, in place of the commented line from my first PR? If so, I'll go ahead and submit a revised PR.

PR #1017, opened 2019-01-01
Fix bug

shafan:


roberto-aguilar:
There is no bug to fix, the code is looking for both settings files and warn the user if that is the case.

The `&&` is the right operator to use, @shafan

svpernova09:
Thanks for the PR, but as @roberto-aguilar mentioned it's currently the correct operator.

PR #1018, opened 2019-01-01
Patch 2

shafan:


svpernova09:
Thanks for taking the time to submit the PR. The current value is a default safe shebang for unattended non interactive work. We don't want to change the environment we run `after.sh` in. This is something the user can do on their end as the `after.sh` that the user would be editing is not in our version control.

PR #1019, opened 2019-01-06
Add rewrite support to nginx config

martijn94:
This PR adds rewrite support to the nginx site config. This makes you redirect/rewrite unfound resources to another destination. 

This is especially usefull for media heavy sites where you don't want all the uploads/media on you local machine. This way you can load then from the production/staging environment.

No apache support needed as this can also be achieved with .htaccess.

### Homestead.yaml syntax

```
sites:
    - map: homestead.test
      to: /home/vagrant/code/public
      rewrites:
        - map: /wp-content/uploads(/.*)
          to: https://my-production-site.com/wp-content/uploads$1
        - map: /storage/images(/.*)
          to: https://my-production-site.com/storage/images$1
```

### Result in nginx config file

```
location ~ /wp-content/uploads(/.*) { if (!-f $request_filename) { return 301 https://my-production-site.com/wp-content/uploads$1; } }
location ~ /storage/images(/.*) { if (!-f $request_filename) { return 301 https://my-production-site.com/storage/images$1; } }
```

Any improvements or comments are very welcome, an alternative way to achieve this without modifying the source when PR is not merged would also be much appreciated.

Thanks

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -256,8 +256,18 @@ class Homestead
             end
             headers += ' )'
           end
+          if site.include? 'rewrites'
+            rewrites = '('
+            site['rewrites'].each do |rewrite|
+                rewrites += ' [' + rewrite['map'] + ']=' + "'" + rewrite['to'] + "'"
+            end
+            rewrites += ' )'
+            # Escape variables for bash
+            rewrites.gsub! '$', '\$'
+          end
+
           s.path = script_dir + "/serve-#{type}.sh"
-          s.args = [site['map'], site['to'], site['port'] ||= http_port, site['ssl'] ||= https_port, site['php'] ||= '7.2', params ||= '', site['zray'] ||= 'false', site['exec'] ||= 'false', headers ||= '']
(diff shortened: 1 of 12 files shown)


svpernova09:
Thanks for the PR.

To accept it, we'll have to make the same changes to all of the other `serve-` scripts that are using nginx.

martijn94:
Just added support to all other site types (Except proxies).

Ran nginx config tests and redirect tests for all types individually.

Also notices some other issues while going through all sites types:

- Elgg site type is missing ssl certificate
- spa site type sites fast cgi parameters outside location block
- All other site types miss the additional headers support
- Some site types miss the additional fastcgi_params support.

Will make some pull request for those issues if needed!

svpernova09:
Thanks for the PRs! Much appreciated.

martijn94:
@svpernova09 Thanks for the merge, much appreciated. 

Would you say that a docs update would be a good idea? Maybe also for the additional header support (when implemented in every `serve-` script)?

PR #1022, opened 2019-01-08
Remove params support from spa site type

martijn94:
Fastcgi params are only supported in PHP location blocks. spa site type does not use PHP


svpernova09:
Thanks!

PR #1026, opened 2019-01-12
Add header support to missing site types

martijn94:
This adds the additional header support to all `serve-` scripts where they were missing


svpernova09:
Thanks for the PRs!

PR #1027, opened 2019-01-12
Add header support to apache site type

martijn94:
This adds the additional header support to the apache site type


svpernova09:
Thanks!

PR #1030, opened 2019-01-15
Version 8

svpernova09:


paulgeneres:
I noticed two functions at the end of the aliases file: update-socket-wrench and seed-socket-wrench. What are they for?

svpernova09:
> I noticed two functions at the end of the aliases file: update-socket-wrench and seed-socket-wrench. What are they for?

I once had delusions of grandeur about including a project IN Homestead that would serve as a websocket server. I was quickly out of my league in terms of getting it working and just didn't have the time. I'd be interested in someone taking over the feature and seeing it through, the main blocker I ran into was making sure the site would be functional if the Homestead user was using nginx or apache.

paulgeneres:
Thanks. I knew there had to be an interesting story behind those intriguing names. I would help if I had a better idea of what you are looking for.

PR #1032, opened 2019-01-15
Update localized Vagrantfile to use same double colon method syntax

fitztrev:
This makes it consistent with `Yaml::load` in the line above it and the other Vagrantfile.

https://github.com/laravel/homestead/blob/237387c99ca2b549945ec4aec0d9c101f9ccf3bc/resources/localized/Vagrantfile#L31

https://github.com/laravel/homestead/blob/237387c99ca2b549945ec4aec0d9c101f9ccf3bc/Vagrantfile#L31


svpernova09:
Thanks!

PR #1040, opened 2019-01-22
Update aliases

davidjr82:
Set alias nrwp="npm run watch-poll" in the same way yarn alias is alias yrwp="yarn run watch-poll"


svpernova09:
Thanks!

PR #1048, opened 2019-01-29
Add possibility to dynamically set folders->map (host path)

Perf:
Add possibility to dynamically set `folders -> map` (host path) in Homestead.yaml and Homestead.json files, when Homestead installed per project. Also `HOST_PATH` variable is passed to `after.sh` and `user-customizations.sh` allowing to perform more advanced setup.

This will allow easier default configuration, and more advanced things like:
put these lines in `after.sh`
```
phpenmod -s ALL xdebug
XDEBUG_INI=/etc/php/7.2/mods-available/xdebug.ini
XDEBUG_INI_ORIG=${XDEBUG_INI}.orig
test -f ${XDEBUG_INI_ORIG} || cp ${XDEBUG_INI} ${XDEBUG_INI_ORIG}
echo "xdebug.file_link_format = \"phpstorm://open?file=%f&line=%l&/home/vagrant/code>${HOST_PATH}\"" >> ${XDEBUG_INI}
echo "xdebug.remote_autostart = 1" >> ${XDEBUG_INI}

service php7.2-fpm restart
```
after that, all links generated by XDebug will be able to open PHPStorm with target file and target line.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -12,6 +12,7 @@ homesteadJsonPath = confDir + "/Homestead.json"
 afterScriptPath = confDir + "/after.sh"
 customizationScriptPath = confDir + "/user-customizations.sh"
 aliasesPath = confDir + "/aliases"
+hostPath = '~/code'
 
 require File.expand_path(File.dirname(__FILE__) + '/scripts/homestead.rb')
 
@@ -26,9 +27,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     end
 
     if File.exist? homesteadYamlPath then
-        settings = YAML::load(File.read(homesteadYamlPath))
+        settings = YAML::load(File.read(homesteadYamlPath) % { HOST_PATH: hostPath })
     elsif File.exist? homesteadJsonPath then
-        settings = JSON::parse(File.read(homesteadJsonPath))
+        settings = JSON::parse(File.read(homesteadJsonPath) % { HOST_PATH: hostPath })
     else
         abort "Homestead settings file not found in #{confDir}"
     end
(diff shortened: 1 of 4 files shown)


Perf:
Hi @svpernova09 ,
something wrong with PR?
I could fix/update it

svpernova09:
Thanks for taking the time to submit this PR but we're not going to implement it at this time.

I'm not seeing how the `HOST_PATH` you're referring to brings any value to the user when this path is completely arbitrary and optional? You can accomplish the same goal by sticking to a known folder structure and utilizing it in your customization scripts.

The other thing I'm not a fan of is the idea that there has to be some top level "root" folder of all your code *in* the VM. Maybe I'm just lazy (I know I am) but I refuse to map any folder past the `/home/vagrant` path because I don't want to have to `cd ~/code` **every** **single** **time** I `vagrant ssh`

I *really* do not like the fact a user would have to open the `Vagrantfile` to adjust this setting. 


I see what you're trying to accomplish but I don't feel like the complexity trade off is worthwhile when you can already accomplish the same tasks via `after.sh`. You could even accomplish this by setting the value of your `HOST_PATH` at the top of your `after.sh`.

Perf:
Hey @svpernova09,
thanks for reply!

> I'm not seeing how the HOST_PATH you're referring to brings any value to the user when this path is completely arbitrary and optional? You can accomplish the same goal by sticking to a known folder structure and utilizing it in your customization scripts.

According to official documentation in paragraph "**Configuring Shared Folders**" it's mandatory.
With this PR everything stays backwards compatible.
But, you don't need to adjust `folders -> map` value in `Homestead.yaml` (or json) file from `~/code` to exact place where your codebase is, as it will be dynamically configured via `HOST_PATH` variable, that is pointing to `__dir__` for a project based installations of Homestead, or to `~/code`, according to default behavior for a single installation.

> The other thing I'm not a fan of is the idea that there has to be some top level "root" folder of all your code in the VM. Maybe I'm just lazy (I know I am) but I refuse to map any folder past the /home/vagrant path because I don't want to have to cd ~/code every single time I vagrant ssh

I see your point, but this is another setting `folder -> to`, which I didn't touched, and it's completely not related to this PR.

> I really do not like the fact a user would have to open the Vagrantfile to adjust this setting.

With this PR user shouldn't do anything with Vagrantfile.

> I see what you're trying to accomplish but I don't feel like the complexity trade off is worthwhile when you can already accomplish the same tasks via after.sh. You could even accomplish this by setting the value of your HOST_PATH at the top of your after.sh.

I didn't saw any complexity tradeoff, as just one variable was added `HOST_PATH`.
Example with `after.sh` is just an example on how you could easily configure PHPStorm and XDebug feature of making PHPStorm specific URLs **without** any specific actions.
And I also didn't included `after.sh` into PR.
PR contains only generic "things" that could be applicable in many cases.

Well, in any case it's up to you to accept or decline.
I just wanted to make Homestead more friendly and easier, this is my intention.

svpernova09:
@Perf, I'll get a reply to you later today. Don't want you to think I'm ignoring your reply :D

PR #1064, opened 2019-02-18
Update install-mysql8.sh to use latest deb package

cosmicvibes:
Resolves apt key signing error which was preventing mysql 8 install.


svpernova09:
Thanks!

PR #1070, opened 2019-03-06
Support Symfony 3.x components

deviantintegral:
I'd like to use Homestead within a Drupal 8 project, but Drupal does not yet support Symfony 4 components. Judging from https://github.com/laravel/homestead/pull/1030 it looks like nothing here actually depends on new 4.x APIs. Note I haven't done any tests beyond phpunit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ php:
 
 before_script:
   - phpenv config-rm xdebug.ini || true
-  - composer install --no-interaction
+  - composer install --no-interaction $PREFER_LOWEST
 
 script:
   - vendor/bin/phpunit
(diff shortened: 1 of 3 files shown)


svpernova09:
We dropped support for Symfony 3.x when we dropped PHP 5.x support. I don't think we want to allow those versions to be used anymore?

deviantintegral:
Symfony 3.4 is an LTS supported until 2021. Drupal 8 [may end up supporting Symfony 4.x](https://www.drupal.org/project/drupal/issues/2937984) before then, but it's not a sure thing. A broader range of supported components would at least open the possibility of more non-Laravel projects using Homestead.

It's possible to use homestead without a site-local config, but there's advantages in keeping it site-local.

Not supporting PHP 5 seems much clearer, and is 💯 from me given it's EOL schedule. This PR isn't touching that requirement.

svpernova09:
Sounds good. Thanks for the PR!

Issue #1071, opened 2019-03-07
Homestead as a replacement for lullabot/trusty32-vm


deviantintegral:
I'm considering abandoning the https://github.com/Lullabot/trusty32-lamp vagrant box I maintain in favour of Homestead, as it's built on Ubuntu 14.04 which EOL's soon. For details, see https://github.com/Lullabot/trusty32-lamp/issues/50

To reach feature parity, there's a few items we'd need to add to Homestead.

# Database Snapshots

One of the core features the the box includes is fast copy-on-write branching and snapshots of MySQL databases. We often deal with large databases in local development, and being able to quickly roll back without a full table or database import / export is a huge time saver. It works with a combination of a thinly-provisioned LVM volume and [the Logical MySQL Manager](https://github.com/Lullabot/lmm) set of scripts.

The good news is that Homestead already uses LVM, but the volumes use all of the PV space. Since Vagrant uses VMDKs and not VDIs, Virtualbox can't expand the volume, and reducing the root logical volume would require booting into single-user mode and remounting `/` readonly.

# Performance profiling

Blackfire.io is included with Homestead, but it's very expensive ($29/USD a month) if you want to trigger a profile from your browser. You can profile from the CLI, but that's tricky when submitting form POSTs that may have many fields, requires escaping, and so on. [xhgui](https://github.com/perftools/xhgui) is the tool I've used before, and Homestead already includes Mongo support.

# mDNS support

Supporting mDNS means you don't have to deal with `/etc/hosts` hacking in most cases. It's also very useful for on-device mobile testing where you'd have to edit DNS records since you can't edit a hosts file. I found https://github.com/laravel/homestead/pull/103, but I don't think we need that much code. Simply installing `avahi-daemon` exposes the hostname as `hostname.local` automatically. I was looking through the repo for where base packages are managed, but all I could find were scripts for optional packages.

# Docs for xdebug

xdebug is included, but not documented on how to use it. I know this isn't homestead specific, but I think documenting how to debug both browser and CLI requests would be helpful to end users.

Before I start on any of these, could I get a 👍 or 👎 ? I think xhgui and mdns could be optional, but I don't see a way to support database snapshots without altering the base box itself.

svpernova09:
I'm 👍 on all these but somewhat concerned about modifying the volumes. I don't want to change the build process for Homestead [https://github.com/laravel/settler](https://github.com/laravel/settler) in regards to running 1 packer command to build a base box to verify.

[...]

PR #1073, opened 2019-03-08
Add support for profiling with tideways and xhgui

deviantintegral:
Following up from https://github.com/laravel/homestead/issues/1071

To enable profiling:

1. In a site, set `xhgui: 'true'`:

```
sites:
    -
        map: drupal8.test
        to: /home/vagrant/code/web
        type: "apache"
        xhgui: 'true'
```

2. Build or provision the vagrant box.
3. To profile a web request, add `xhgui=on` as a query parameter. A cookie is set so future requests will be profiled (like form submits or ajax requests), and will expire 1 hour after the last request (as profiling is a significant performance hit in itself).
4. To profile a CLI request, set an `XHGUI=on` environment variable with `export` or for a single command with `XHGUI=on php ...`.

TODOs:

- [x] Add support for nginx too.
- [ ] Documentation somewhere?

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -267,7 +267,7 @@ class Homestead
           end
 
           s.path = script_dir + "/serve-#{type}.sh"
-          s.args = [site['map'], site['to'], site['port'] ||= http_port, site['ssl'] ||= https_port, site['php'] ||= '7.3', params ||= '', site['zray'] ||= 'false', site['exec'] ||= 'false', headers ||= '', rewrites ||= '']
+          s.args = [site['map'], site['to'], site['port'] ||= http_port, site['ssl'] ||= https_port, site['php'] ||= '7.3', params ||= '', site['zray'] ||= 'false', site['xhgui'] ||= '', site['exec'] ||= 'false', headers ||= '', rewrites ||= '']
 
           if site['zray'] == 'true'
             config.vm.provision 'shell' do |s|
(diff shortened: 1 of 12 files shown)


scripts/homestead.rb
```diff
@@ -267,7 +267,7 @@ def self.configure(config, settings)
           end
 
           s.path = script_dir + "/serve-#{type}.sh"
-          s.args = [site['map'], site['to'], site['port'] ||= http_port, site['ssl'] ||= https_port, site['php'] ||= '7.3', params ||= '', site['zray'] ||= 'false', site['exec'] ||= 'false', headers ||= '', rewrites ||= '']
+          s.args = [site['map'], site['to'], site['port'] ||= http_port, site['ssl'] ||= https_port, site['php'] ||= '7.3', params ||= '', site['zray'] ||= 'false', site['xhgui'] ||= '', site['exec'] ||= 'false', headers ||= '', rewrites ||= '']
```
deviantintegral: This is starting to get pretty tricky, and was a lot of copy-pasta for the shell scripts. I think this could be refactored (the nginx templates are substantially similar), but the arguments are not 100% consistent between each one. I've left this as is, thinking a followup PR can clean this and the shell scripts up either by extracting some functions in bash, or perhaps going all in and converting it to be a RoboFile to run with https://robo.li.

svpernova09:
Thanks!

PR #1080, opened 2019-03-25
Correct variable index in reaction to new xhgui option being added

Liopia:
Corrects variable index shift introduced in #1073


svpernova09:
Thanks!

PR #1084, opened 2019-04-08
Support for i386 architecture removed from the MariaDB installation

afgloeden:
The MariaDB repository does not support i386 architecture, when the `sudo apt update` command is used, the error is reported:

```
N: Skipping acquire of configured file 'main/binary-i386/Packages' as repository 'http://ftp.osuosl.org/pub/mariadb/repo/10.3/ubuntu bionic InRelease' doesn't support architecture 'i386'
```


svpernova09:
Thanks!

PR #1105, opened 2019-04-23
Set default shell to zsh when installing oh-my-zsh

### LINKED ISSUE #1099: When installing oh-my-zsh, it should set the default terminal to zsh too ?
When installing oh-my-zsh, it should set the default terminal to zsh too, or it should remain the bash ?
If it should be set to zsh, I'm happy to provide a PR.

ricardoseriani:
Closes #1099 


svpernova09:
Thanks!

PR #1110, opened 2019-04-24
Fix owner of /home/vagrant/.webdriverutils

ricardoseriani:
Closes #1109 

### DIFF
--- a/scripts/install-webdriver.sh
+++ b/scripts/install-webdriver.sh
@@ -7,6 +7,7 @@ then
 fi
 
 touch /home/vagrant/.webdriverutils
+chown -Rf vagrant:vagrant /home/vagrant/.webdriverutils
 
 # Install The Chrome Web Driver & Dusk Utilities
 


svpernova09:
Thanks!

browner12:
Sorry to jump on this PR. I seem to running into the same issue with `/home/vagrant/.npm`, and ends up with me running npm with `sudo`.  Is the fix the same thing?

ricardoseriani:
> Sorry to jump on this PR. I seem to running into the same issue with `/home/vagrant/.npm`, and ends up with me running npm with `sudo`. Is the fix the same thing?

I'm no maintainer, but I think that you should copy the commands that you are using and log/error, then open an issue about this, so anyone could verify it for you.

deviantintegral:
Since the box doesn't do significant provisioning after the first boot, running `sudo chown -Rv vagrant:vagrant ~/.npm` should fix things and save you from having to `sudo npm` until this is fixed.

svpernova09:
Sounds like maybe we should add a specific inline `chown -R vagrant:vagrant /home/vagrant`?

deviantintegral:
I was thinking the same thing. Another option would be to add a check with something like `find /home/vagrant -not -user vagrant -and -not -group vagrant` that errors out during builds if any results are found. That would alert us to new problems without blindly chown'ing the directory.

svpernova09:
> I was thinking the same thing. Another option would be to add a check with something like find /home/vagrant -not -user vagrant -and -not -group vagrant that errors out during builds if any results are found. That would alert us to new problems without blindly chown'ing the directory.

fwiw Settler does a blind chown -R on the home folder as one of the last things during cleanup lol. Not opposed to having Homestead do it.

ricardoseriani:
> Sounds like maybe we should add a specific inline `chown -R vagrant:vagrant /home/vagrant` that runs every time the box is provisioned?
> 
> btw tons of thanks to @ricardoseriani for cleaning up the messes I left <3

Wow, thanks a lot for the compliment 😊.
About the `chown -R vagrant:vagrant /home/vagrant`, anyone know what he does with the mapped folders ?
I'll try to discover this, because I'm afraid that it messes up with these folders.

ricardoseriani:
Apparently, the 'chown -R vagrant:vagrant /home/vagrant` should not cause any problem, as it is stated in [https://github.com/hashicorp/vagrant/issues/897](https://github.com/hashicorp/vagrant/issues/897), on first and second comments.

I can make a PR with this, but I'm not certain what is the best way to ensure that this script will be the last running.

svpernova09:
The more I think about it, Homestead doing a blind chown sounds gross. The only newly discovered issues here should be the features I "shifted", if we've fixed the chown on those installers, I'm happy to leave it at that.

ricardoseriani:
For sure, I think it's best to doing this selectively than doing it in general /home/vagrant folder.

PR #1115, opened 2019-04-27
Add option to install docker-ce

### LINKED ISSUE #1114: [feature request] Add opt-in option to install docker-ce
## Versions
  - Vagrant: 2.2.4
  - Provider: Virtualbox 6.0.6
  - Homestead: 7.2.1

Add a opt-in feature to install docker-ce, like crystal, dotnetcore, golang, etc.

If this is useful, I'm happy to provide a PR, but I have a couple of questions:

When installing docker-ce, what method should be used ?
- deb
- repository (manual)
- script provided by docker

The vagrant user should be able to run docker without sudo ?
(this can have security issues, as quoted in  [https://docs.docker.com/install/linux/linux-postinstall/](https://docs.docker.com/install/linux/linux-postinstall/), `the docker group grants privileges equivalent to the root user. For details on how this impacts security in your system, see Docker Daemon Attack Surface`?

ricardoseriani:
Closes #1114 

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -395,6 +395,14 @@ class Homestead
       end
     end
 
+    # Install Docker-CE If Necessary
+    if settings.has_key?("docker") && settings["docker"]
+        config.vm.provision "shell" do |s|
+            s.name = "Installing Docker-CE"
+            s.path = script_dir + "/install-docker-ce.sh"
+        end
+    end
+
     # Install DotNetCore If Necessary
     if settings.has_key?("dotnetcore") && settings["dotnetcore"]
         config.vm.provision "shell" do |s|
--- a/scripts/install-docker-ce.sh
+++ b/scripts/install-docker-ce.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+# Install docker-ce
+curl -fsSL https://get.docker.com | bash -s
+
+# Enable vagrant user to run docker commands
+usermod -aG docker vagrant


ricardoseriani:
The script provided by docker output some messages to the stdout when provisioning. Is this a problem ? If so, I can update this to use the deb or the manual installation procedure.

And I noticed that some scripts create a .[script-name] file on the home directory to avoid reinstalling.In this case, this PR should created this too, or avoid it so when doing the provision it will auto-update the docker installed ?

svpernova09:
> And I noticed that some scripts create a .[script-name] file on the home directory to avoid reinstalling.In this case, this PR should created this too, or avoid it so when doing the provision it will auto-update the docker installed ?

Lets run with it for now and see how it goes. Thanks!

browner12:
can you guys ELI5 why we'd want docker within our VM? wouldn't you just install Docker Desktop onto your host?

deviantintegral:
I've never had great experiences with Docker Desktop on macOS. In particular, the hyperkit process will sometimes spin at 100% CPU, killing my battery. `osxfs` is pretty slow for me as well. Finally, the memory use of the Docker VM can be very high, even when running just a single container. I've had none of these issues with Docker on Linux.

I'm using / have used Docker inside the VM for two cases:

1. On the project I'm working on right now, we are distributing databases for development as docker images. Inside the VM, I disable mysql entirely and run it from docker, but keep everything else running directly on the VM image.
1. For stateless services like Varnish, it's the easiest way to run the latest versions if backports  / PPAs don't exist.

If I was using a Linux laptop, I agree there isn't much point. But for Windows and macOS users, it's pretty useful.

browner12:
interesting, thanks for the info.

PR #1122, opened 2019-05-03
Update homestead.rb to include 9600 port forward

jmannau:
The docs indicate that 9600 is included in the default port forwards for MinIo. Updated homestead.rb to reflect this.


svpernova09:
thanks.

PR #1129, opened 2019-05-07
Add Sites from Homestead.yaml to /etc/hosts in guest

cbj4074:
This is a follow-up to #1007 and implements the feature.

### DIFF
--- a/scripts/hosts-add.sh
+++ b/scripts/hosts-add.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+# Add new IP-host pair to /etc/hosts.
+
+if [[ "$1" && "$2" ]]
+then
+    IP=$1
+    HOSTNAME=$2
+
+    if [ -n "$(grep [^\.]$HOSTNAME /etc/hosts)" ]
+        then
+            echo "$HOSTNAME already exists:";
+            echo $(grep [^\.]$HOSTNAME /etc/hosts);
+        else
+            sudo sed -i "/#### HOMESTEAD-SITES-BEGIN/c\#### HOMESTEAD-SITES-BEGIN\\n$IP\t$HOSTNAME\twww.$HOSTNAME" /etc/hosts
+
+            if ! [ -n "$(grep [^\.]$HOSTNAME /etc/hosts)" ]
+                then
+                    echo "Failed to Add $HOSTNAME, Try again!";
+            fi
+    fi
+else
+    echo "Error: missing required parameters."
+    echo "Usage: "
+    echo "  addhost ip domain"
+fi
--- a/scripts/serve-laravel.sh
+++ b/scripts/serve-laravel.sh
@@ -103,4 +103,5 @@ block="server {
 
 echo "$block" > "/etc/nginx/sites-available/$1"
 ln -fs "/etc/nginx/sites-available/$1" "/etc/nginx/sites-enabled/$1"
-#echo "127.0.0.1 $1" >> /etc/hosts
+
+sudo /vagrant/scripts/hosts-add.sh '127.0.0.1' "$1"
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -223,6 +223,11 @@ class Homestead
     config.vm.provision 'shell' do |s|
       s.path = script_dir + '/clear-nginx.sh'
     end
+    
+    # Clear any Homestead sites and insert markers in /etc/hosts
+    config.vm.provision 'shell' do |s|
+      s.path = script_dir + '/hosts-reset.sh'
+    end
 
     if settings.include? 'sites'
       # socket = { 'map' => 'socket-wrench.test', 'to' => '/var/www/socket-wrench/public' }
--- a/scripts/hosts-reset.sh
+++ b/scripts/hosts-reset.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+# Remove any Homestead entries from /etc/hosts and prepare for adding new ones.
+
+sudo sed -i '/#### HOMESTEAD-SITES-BEGIN/,/#### HOMESTEAD-SITES-END/d' /etc/hosts
+
+printf "#### HOMESTEAD-SITES-BEGIN\n#### HOMESTEAD-SITES-END" | sudo tee -a /etc/hosts > /dev/null


browner12:
is this for the Host or the Guest?

cbj4074:
@browner12 The guest.

The idea is explained in detail in #1007 .

browner12:
Yup, thanks. I thought we had fixed that part, and were trying the Host side of things, but I guess I misread.

cbj4074:
No problem; I should have mentioned that it's for the guest in the PR title! I'll edit it to that effect.

cbj4074:
Actually, hold-off @svpernova09 , I'm going to rework this slightly.

Adding new sites to `Homestead.yaml` and running `vagrant reload --provision` causes any new site to be added to `/etc/hosts` in the guest, *but*, removing a site from `Homestead.yaml` and doing the same does not cause the removed site to be removed from `/etc/hosts`, which when I say it out loud, is not at all surprising.

And in fact, you and @browner12 did already mention this and provide suggestions for how to implement it [here](https://github.com/laravel/homestead/pull/1007#issuecomment-488045003) and [here](https://github.com/laravel/homestead/pull/1007#issuecomment-488050430), respectively.
 
I'm going to do as you both suggested and add a unique block that contains these entries so we can empty only that block upon re-provisioning, and then re-add each hostname.

I'll force-push an update to my branch shortly.

cbj4074:
Okay, I think this is finally doing what we had intended.

@svpernova09 If you'd prefer the code to be organized differently, just let me know (or feel free to make changes as needed).

One thing I noticed while testing -- and I have no idea how it could possibly be related to these changes -- is that when I `vagrant ssh`, sometimes I see duplicate Homestead banners.

Has anybody else seen this happen? I can't tell if it's "real", or some glitch from my shell (I'm using Cmder wrapped around PowerShell 6).

```
λ  vagrant ssh
Welcome to Ubuntu 18.04.2 LTS (GNU/Linux 4.15.0-47-generic x86_64)

Thanks for using
 _                               _                 _
| |                             | |               | |
| |__   ___  _ __ ___   ___  ___| |_ ___  __ _  __| |
| '_ \ / _ \| '_ ` _ \ / _ \/ __| __/ _ \/ _` |/ _` |
| | | | (_) | | | | | |  __/\__ \ ||  __/ (_| | (_| |
|_| |_|\___/|_| |_| |_|\___||___/\__\___|\__,_|\__,_|
Welcome to Ubuntu 18.04.2 LTS (GNU/Linux 4.15.0-47-generic x86_64)

Thanks for using
 _                               _                 _
| |                             | |               | |
| |__   ___  _ __ ___   ___  ___| |_ ___  __ _  __| |
| '_ \ / _ \| '_ ` _ \ / _ \/ __| __/ _ \/ _` |/ _` |
| | | | (_) | | | | | |  __/\__ \ ||  __/ (_| | (_| |
|_| |_|\___/|_| |_| |_|\___||___/\__\___|\__,_|\__,_|

* Homestead 8.4.0 released!
* Settler v7.2.1 released! Make sure you update

153 packages can be updated.
17 updates are security updates.


Last login: Tue May  7 22:02:52 2019 from 192.168.96.1
```

browner12:
forgive me if this has been addressed already, I haven't had time to read all of the comments on this feature.

It seems very strange to me to rely on order of operations for this to all work. If someone were to add a new Laravel site outside the context of provisioning, it doesn't seem like it would get added in the appropriate place in `/etc/hosts`.

I believe Laravel Forge runs into this same issue when updating Nginx config files. What they do is have markers (comments), and then edit based on looking for those markers. Couldn't we do something similar?  On initial provision, we add a marker comment in `/etc/hosts`, and then when provisioning we search (`sed` or whatever) and add the entry below the marker line?

cbj4074:
@browner12 No problem.

Firstly, we *are* using markers.

Secondly, I'm not sure I understand what you mean by

> If someone were to add a new Laravel site outside the context of provisioning, it doesn't seem like it would get added in the appropriate place in /etc/hosts.

Isn't the *only* way to add a new Laravel site to Homestead to add it to `Homestead.yaml` and run `vagrant reload --provision`?

Is your concern that somebody is going to manually edit `/etc/hosts` in the *guest*? I can't conceive of any reason for which somebody would do that. Is that a valid and supported use-case for Homestead?

To explain what, exactly, this PR does:

1. Upon provisioning (whether initial or re-provision), look for the begin/end markers in `/etc/hosts` and remove them and everything between them. In effect, "reset" the `hosts` file to its pre-provisioning state.
2. Add a beginning marker to `/etc/hosts`, so we know where the Homestead sites start (so as not to interfere with the default, OS-level entries, or any that have been added outside of `Homestead.yaml`).
3. Iterate over the sites in `Homestead.yaml` and add an entry for each one after the marker.
4. Add an ending marker, so we can target the entries between this and the beginning marker in step 1.

*Much* of the provisioning process relies on order-of-operations. Certain assumptions must be made, and in this case, the only assumption we're making is that the user doesn't jam entries between the start and end markers that we put into `/etc/hosts` manually. Again, why somebody would do that eludes me.

Even if somebody is using Homestead in the manner you describe, and makes entries in the guest VM's `/etc/hosts` file, as long as they are not between our custom markers, there is no issue (we won't touch them).

browner12:
These may not be as used as much as they were back in the day, but there are actually function in your `~/.bash_aliases` file called things like `serve-laravel` and `serve-apache`, which call little bash scripts. for example, here's the `serve-laravel()` one

```sh
function serve-laravel() {
    if [[ "$1" && "$2" ]]
    then
        sudo bash /vagrant/scripts/create-certificate.sh "$1"
        sudo dos2unix /vagrant/scripts/serve-laravel.sh
        sudo bash /vagrant/scripts/serve-laravel.sh "$1" "$2" 80 443 "${3:-7.1}"
    else
        echo "Usage: "
        echo "  serve domain path"
    fi
}
```

If someone were to add a new site to Homestead this way, they would not get their host file updated correctly.

I would propose 3 new scripts:

- `hosts-clear.sh`
- `hosts-add.sh`
- `hosts-remove.sh`

The provisioning script would *only* setup the markers.

Then the serve scripts (`serve-laravel.sh`) would *only* call the `hosts-add.sh`.  The 'add' script would search `/etc/hosts` for the opening marker, and then insert a new line below it.

cbj4074:
Ahh, fair enough.

Now I understand what you meant by `If someone were to add a new Laravel site outside the context of provisioning...`. Makes perfect sense, given your explanation. I didn't realize that people used those functions outside of the provisioning context.

I force-pushed over it, but that's how I had done this originally. Thankfully, I still have all the files, so it'll be easy to go back to that approach.

I'll take another look at this in the morning and try to implement in the way you describe. I agree that it's cleaner and more robust.

Thank you for taking the time to digest what's happening here and provide valuable feedback!

browner12:
Absolutely, glad to help! thanks for taking the initiative here to get this working, it's been a PITA for awhile

cbj4074:
One other thought: if somebody were to remove a site from `Homestead.yaml` and re-provision, how would we remove it from `/etc/hosts`?

The site is gone from the configuration, so there's no way to "know which site to remove" unless we remove them all.

When I was pondering how to make this work earlier, this is why I approach it as I did.

browner12:
That's what I was thinking about with the `hosts-clear.sh` script. Maybe a more appropriate name would be `hosts-reset.sh`.

I think you'll have to blow away any of our customization and re-add them, since like you said, there's no way for us to know which ones are removed without a lot of unnecessary processing. So the reset script should:

1. remove everything between and including our markers
1. add the markers

then we can add sites to the hosts file like new

svpernova09:
> Has anybody else seen this happen? I can't tell if it's "real", or some glitch from my shell 

I updated the motd site yesterday and I'm intermittently seeing this as well. It shouldn't affect VM operation. I'll fix it as soon as I can pinpoint the cause.

svpernova09:
I'll play around with this tonight / tomorrow.

cbj4074:
@browner12 Please take another look when you have a chance.

I believe I've implemented this exactly as you describe in your previous comments.

I should note, in case anybody wonders about it, that [calling Bash explicitly seems to be necessary here](https://github.com/laravel/homestead/pull/1129/files#diff-e4c97683023b78def58da15578c8d737R107), as using just `sh` seems to use Dash, and the script fails to be parsed with a syntax error.

browner12:
I think your explicit `sh` issue may be due to the fact you are creating them as functions, and then calling the function within the script. It doesn't seem like any of the other scripts behave this way, did you have a reason for doing it that way?

Also, would we want to make both a non-www and www entry for the site? So it would end up as 

```
127.0.0.1   mysite.test    www.mysite.test
```

cbj4074:
> It doesn't seem like any of the other scripts behave this way, did you have a reason for doing it that way?

No, not a good reason. Initially, I had considered adding those functions as Bash aliases, but decided against it, and copy-pasted them as they were into individual scripts. I've modified the scripts to reflect the pattern used elsewhere, but it doesn't eliminate the need to call bash explicitly.

As it turns-out, the proper fix is not to specify the shell in the command at all, and let the exec process determine which shell to use based on the shebang at the top of the script.

And yes, it probably makes sense to include a www-version of the hostname, which I've added.

How's it all look to you now?

scripts/hosts-add.sh
```diff
@@ -0,0 +1,26 @@
...
+    HOSTNAME=$2
+
+    if [ -n "$(grep $HOSTNAME /etc/hosts)" ]
+        then
+            echo "$HOSTNAME already exists:";
+            echo $(grep $HOSTNAME /etc/hosts);
+        else
+            sudo sed -i "/#### HOMESTEAD-SITES-BEGIN/c\#### HOMESTEAD-SITES-BEGIN\\n$IP\t$HOSTNAME\twww.$HOSTNAME" /etc/hosts
+
+            if ! [ -n "$(grep $HOSTNAME /etc/hosts)" ]
```
browner12: why would it fail to add? if the markers don't exist?

scripts/hosts-add.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+# Add new IP-host pair to /etc/hosts.
+
+if [[ "$1" && "$2" ]]
+then
+    IP=$1
+    HOSTNAME=$2
+
+    if [ -n "$(grep $HOSTNAME /etc/hosts)" ]
```
browner12: this is probably an edge case, and probably not worth holding up this merge to figure out, but would subdomains cause issues here? for example, if `admin.mysite.test` is already setup, and we try to add `mysite.test`, I believe your grep would find it.

scripts/hosts-add.sh
```diff
@@ -0,0 +1,26 @@
...
+        else
+            sudo sed -i "/#### HOMESTEAD-SITES-BEGIN/c\#### HOMESTEAD-SITES-BEGIN\\n$IP\t$HOSTNAME\twww.$HOSTNAME" /etc/hosts
+
+            if ! [ -n "$(grep $HOSTNAME /etc/hosts)" ]
+                then
+                    echo "Failed to Add $HOSTNAME, Try again!";
+            fi
+    fi
+else
+    echo "Error: missing required parameters."
```
browner12: can these all go in one `echo` statement?

scripts/hosts-add.sh
cbj4074: I suppose it could fail for any number of reasons, but the check is basically saying, "If the markers are not found, complain about it".

FWIW, I grabbed that, verbatim, from https://github.com/laravel/homestead/pull/1007#issuecomment-489484013 and didn't clean it up beyond reducing verbosity for the success cases. We can certainly eliminate any error-handling in this use-case; it looks like the code was originally written to be a bit more modular and used within a function.

scripts/hosts-add.sh
cbj4074: Same deal here; I didn't write this code, but yes, I believe you are correct. And while I agree that it shouldn't hold-up the merge, it would be trivial to fix. Good catch.

scripts/hosts-add.sh
cbj4074: They could, but I believe the original author (whomever it was) did this so as to avoid having to insert newlines. Using `echo` in this way has the same effect as using a single call with newlines in each place that a linebreak is desired.

browner12: ahh, good to know.

cbj4074: Indeed. This is probably most legible as-is, so I'll mark this Resolved.

scripts/hosts-add.sh
cbj4074: Upon further consideration, I think it makes sense to leave the output as-is for failure scenarios, only because it will make problems more obvious; it seems less than ideal to ignore failures silently.

scripts/hosts-add.sh
cbj4074: I just fixed this by requiring that the match not begin with a `.`, which would prevent the scenario you describe.

You can see what I mean by examining the demo at https://regex101.com/r/V7lYpc/2 . For the expression to match, it would need to be changed from `[^.]homestead.site` to `[^.]docs.homestead.site`, which achieves what we want, if I'm not mistaken.

There's still a flaw with it, which is that the `.` character in `$HOSTNAME` is not matched literally, but rather, represents "any character". By no means a show-stopper, but worth mentioning.

The way to fix it would be to call the equivalent of https://php.net/preg_quote on `$HOSTNAME`, but I don't know enough Bash to do that without guessing.

svpernova09:
> Also, would we want to make both a non-www and www entry for the site? So it would end up as

IMHO we should always trust the end user is specifying domains as they expect to use them. So I would not make any entries for hosts not explicitly specified. (`homestead.test` gets put in as `homestead.test` alone, no entry for `www.homestead.test`

scripts/serve-laravel.sh
```diff
@@ -103,4 +103,5 @@ block="server {
 
 echo "$block" > "/etc/nginx/sites-available/$1"
 ln -fs "/etc/nginx/sites-available/$1" "/etc/nginx/sites-enabled/$1"
-#echo "127.0.0.1 $1" >> /etc/hosts
+
+sudo /vagrant/scripts/hosts-add.sh '127.0.0.1' "$1"
```
svpernova09: We'll want to add this line to all the `scripts/serve-*.sh` files.

browner12:
> IMHO we should always trust the end user is specifying domains as they expect to use them.

In our Nginx config we specify the `server_name` as `.mysite.test`, which means we're responding to all subdomains.  I feel like we need to make some attempt here, since www is so common.

svpernova09:
> In our Nginx config we specify the server_name as .mysite.test, which means we're responding to all subdomains. I feel like we need to make some attempt here, since www is so common.

Right, which means if a user wants to use `www.mysite.test` it would be added as `.www.mysite.test`, but we wouldn't want to add `www.www.mysite.test` to the hosts file.

browner12:
works for me. I think we've got a good base for now, so let's get this in, and then see if there are enhancements we can make later.

svpernova09:
We're good to merge this once the scripts are flagged `+x`. My local testing looks good from here.

svpernova09:
Thanks for all the work on this everyone!

cbj4074:
Sorry for the late reply; I had stepped away for a couple hours.

Good catch re: the `+x` flag @svpernova09 . Thanks for fixing it!

It seems like we're all in agreement that this can be merged with confidence and we can fine-tune in a subsequent PR.

As it stands, what changes do we need to make? Just these two?

1. Remove the automatic `www` inclusion.
2. Replace (or add, as applicable) `echo "127.0.0.1 $1" >> /etc/hosts` by `sudo /vagrant/scripts/hosts-add.sh '127.0.0.1' "$1"` in `scripts/serve-*.sh`.

PR #1130, opened 2019-05-07
Fix ruby install script

### LINKED ISSUE #1125: Enabling ruby doesn't install 2.6.1 as intended
## Versions
  - Vagrant: 2.2.4
  - Provider: Virtualbox 6.0.6
  - Homestead: 8.4.0

## Host operating system
Arch Linux

## Homestead.yaml
```yaml
---
ip: "192.168.22.2"
memory: 2048
cpus: 1
provider: virtualbox
mariadb: true
ruby: true

authorize: ~/.ssh/id_rsa.pub

keys:
    - ~/.ssh/id_rsa

copy:
   - from: ~/.gitconfig
     to: /home/vagrant
   - from: ~/.homestead/.bashrc
     to: /home/vagrant

folders:
    - map: ~/Sites
      to: /home/vagrant/Sites

sites:
    - map: laravel.io
      to: /home/vagrant/Sites/laravel/public
      schedule: true

databases:
    - homestead
```

## Vagrant destroy & up output
```
    homestead-7: Processing triggers for libc-bin (2.27-3ubuntu1) ...
    homestead-7: Cloning into '/home/vagrant/.rbenv'...
    homestead-7: Cloning into '/home/vagrant/.rbenv/plugins/ruby-build'...
    homestead-7: ruby-build: definition not found: 2.6.1
    homestead-7: 
    homestead-7: See all available versions with `rbenv install --list'.
    homestead-7: 
    homestead-7: If the version you need is missing, try upgrading ruby-build
    homestead-7: .
    homestead-7: rbenv: version `2.6.1' not installed
    homestead-7: Reading package lists...
```

## Expected behavior
`rbenv install 2.6.1` should work as intended and install version 2.6.1 of ruby

## Actual behavior
It fails to install and falls back instead to the the system ruby installed previously when the below command was run
```sh
apt-get -y install libssl-dev libyaml-dev libxml2-dev libxslt1-dev libcurl4-openssl-dev software-properties-common \
libffi-dev rbenv
```
This command installed ruby 2.5.1 as a dependency.

## Steps to reproduce
1. Enable ruby in your `Homestead.yaml`

## Solution/Work around
What appears to be happening is that the `rbenv install 2.6.1` command seems to be using the apt-get installed old outdated version which doesn't accept newer versions as valid.

It only supports up to `2.5.*` so this is invalid.

As a very hackie and dangerous work around it can be tricked to use `/home/vagrant` as a home directory and that seems to fix it like below.
```sh
echo 'export PATH="$HOME/.rbenv/plugins/ruby-build/bin:$PATH"' >> /home/vagrant/.bashrc

export HOME="/home/vagrant"

rbenv install 2.6.1
```

Although this fixes the issue ans seems to be working just fine from what i can tell I'm a bit uncomfortable with this solution.

I leave it up to the maintainer if he wants to go with it.

ricardoseriani:
Point PATH config to ~/.profile, so it's loaded in sudo
Run rbenv and gem as vagrant user

Closes #1125

### DIFF
--- a/scripts/install-ruby.sh
+++ b/scripts/install-ruby.sh
@@ -13,9 +13,10 @@ echo 'eval "$(rbenv init -)"' >> /home/vagrant/.profile
 git clone https://github.com/rbenv/ruby-build.git /home/vagrant/.rbenv/plugins/ruby-build
 echo 'export PATH="$HOME/.rbenv/plugins/ruby-build/bin:$PATH"' >> /home/vagrant/.profile
 
-rbenv install 2.6.1
-rbenv global 2.6.1
-apt-get -y install ruby`ruby -e 'puts RUBY_VERSION[/\d+\.\d+/]'`-dev
-gem install rails -v 5.2.2
-rbenv rehash
 chown -Rf vagrant:vagrant /home/vagrant/.rbenv
+
+# Run as vagrant user
+sudo -i -u vagrant -- rbenv install 2.6.1
+sudo -i -u vagrant -- rbenv global 2.6.1
+sudo -i -u vagrant -- rbenv rehash
+sudo -i -u vagrant -- gem install rails -v 5.2.2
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks!

PR #1133, opened 2019-05-08
Port 80 in the import links of the GPG keys removed from the scripts

afgloeden:
When using Homestead with proxy, port 80 on GPG key bindings causes problems. Details:

* Windows 7 Professional
* Vagrant 2.2.4
* VirtualBox 6.0.6
* Box laravel/homestead 7.2.1
* Homestead in master branch (Latest release v8.4.0)

Vagrant plugins:

```shell
$ vagrant plugin list
vagrant-proxyconf (2.0.1, global)
  - Version Constraint: > 0
```

Logs:

* MariaDB:

```shell
vagrant@homestead:~$ sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0xF1656F24C74CD1D8
Executing: /tmp/apt-key-gpghome.TRZRc3GO4x/gpg.1.sh --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0xF1656F24C74CD1D8
gpg: keyserver receive failed: No keyserver available

vagrant@homestead:~$ sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com --recv-keys 0xF1656F24C74CD1D8
Executing: /tmp/apt-key-gpghome.bpQLD7KW90/gpg.1.sh --keyserver hkp://keyserver.ubuntu.com --recv-keys 0xF1656F24C74CD1D8
gpg: key F1656F24C74CD1D8: 6 signatures not checked due to missing keys
gpg: key F1656F24C74CD1D8: public key "MariaDB Signing Key <signing-key@mariadb.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
```

* MongoDB:

```shell
vagrant@homestead:~$ sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 9DA31620334BD75D9DCB49F368818C72E52529D4 2>&1
Executing: /tmp/apt-key-gpghome.fp4SRSGqRm/gpg.1.sh --keyserver hkp://keyserver.ubuntu.com:80 --recv 9DA31620334BD75D9DCB49F368818C72E52529D4
gpg: keyserver receive failed: No keyserver available

vagrant@homestead:~$ sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com --recv 9DA31620334BD75D9DCB49F368818C72E52529D4 2>&1
Executing: /tmp/apt-key-gpghome.RGzBmoAM73/gpg.1.sh --keyserver hkp://keyserver.ubuntu.com --recv 9DA31620334BD75D9DCB49F368818C72E52529D4
gpg: key 68818C72E52529D4: public key "MongoDB 4.0 Release Signing Key <packaging@mongodb.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1
```

### DIFF
--- a/scripts/install-crystal.sh
+++ b/scripts/install-crystal.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Install Crystal Programming Language Support
-apt-key adv --keyserver hkp://keys.gnupg.net:80 --recv-keys 09617FD37CC06B54
+apt-key adv --keyserver hkp://keys.gnupg.net --recv-keys 09617FD37CC06B54
 echo "deb https://dist.crystal-lang.org/apt crystal main" | tee /etc/apt/sources.list.d/crystal.list
 apt-get update
 apt-get install -y crystal
--- a/scripts/install-maria.sh
+++ b/scripts/install-maria.sh
@@ -31,7 +31,7 @@ rm -rf /etc/mysql
 
 # Add Maria PPA
 
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 0xF1656F24C74CD1D8
+sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com --recv-keys 0xF1656F24C74CD1D8
 sudo add-apt-repository 'deb [arch=amd64,ppc64el] http://ftp.osuosl.org/pub/mariadb/repo/10.4/ubuntu bionic main'
 apt-get update
 
--- a/scripts/install-mongo.sh
+++ b/scripts/install-mongo.sh
@@ -11,7 +11,7 @@ fi
 touch /home/vagrant/.mongo
 chown -Rf vagrant:vagrant /home/vagrant/.mongo
 
-sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 9DA31620334BD75D9DCB49F368818C72E52529D4 2>&1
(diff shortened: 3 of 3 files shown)


cbj4074:
These changes may warrant further discussion.

That `hkp` protocol uses a port that is closed outbound for a significant number of users. Have a look at https://superuser.com/a/222083 .

I know that in my particular case, I'm unable to provision when port 80 is not specified explicitly, and I suspect I'm not alone.

I'm not saying that my use-case is any more important than that of somebody using a proxy, but I think we need to dig into this a bit deeper before we merge this change.

svpernova09:
Thanks for taking the time to open a pull request.

One of the foundations of Homestead is it shouldn't *require* any extra plugins for Vagrant (outside of Providers) to work. Also Homestead shouldn't be concerned with your host's networking (proxies in this case). The responsibility of proxy configuration and set up has to be on you, the end user because it's not feasible for me to even try to maintain support for proxies.  I'm not at all opposed to improvements that help facilitate these types of things but it shouldn't be the default (IMHO).

I feel like this can be solved via existing `after.sh` and `user-customization.sh` on an as needed basis. Simply copy the `scripts/install-maria.sh` contents and move it to your `after.sh` and make any changes there. I admit it's not ideal, I'm just not willing to risk this change on everything without fully understanding the impact.

I also will echo everything @cbj4074 has mentioned above.

afgloeden:
I did not know about it @cbj4074, it is better to continue using port 80 anyway and I will base myself on your explanation for the next pull requests @svpernova09, thanks.

PR #1140, opened 2019-05-13
Update default num cpus to 2

### LINKED ISSUE #1136: Default to 2 CPUs in Homestead.yml
This reminded me:

https://github.com/laravel/homestead/issues/1134#issuecomment-490996336

For myself, I always run with 4 CPUs, even when running multiple VMs. There's a few reasons:

1. Even for a simple site, you'll have a web server, php-fpm, and mysql all with multiple processes.
1. Browsers spawn multiple requests, so you'll have multiple threads or processes of all of the above running.
1. On battery, "race to idle" is generally a better strategy for battery life. If you have a bunch of requests hitting a single CPU core, the CPU is more likely to scale up it's frequency to meet the load.

While hyperthreading means the host will see 2x the number of cores, the kernel scheduler needs to be hyperthreading aware and know which cores are "virtual". Unfortunately, Virtualbox (and VMWare a long time ago) didn't expose this to VMs, so going to 8 CPUs would actually be a significant performance hit.

Someone I work with recently commented how slow the initial boot was and thought it was a general Homestead problem. After putting CPUs to 4, initial boots were perfectly fast. So, I think this could affect perception of Homestead's performance compared to MAMP and others that will use all CPU cores.

My suggestion is that we default to 2 CPUs, since I think it's been years since single-core systems were sold. If we default to 4, we could have VMs failing to boot or poor performance for 2-core i5s with hyperthreading. We could put a note in the docs summarizing the above details and suggesting that it be customized.

ricardoseriani:
Closes #1136 

### DIFF
--- a/Homestead.yaml.example
+++ b/Homestead.yaml.example
@@ -1,7 +1,7 @@
 ---
 ip: "192.168.10.10"
 memory: 2048
-cpus: 1
+cpus: 2
 provider: virtualbox
 
 authorize: ~/.ssh/id_rsa.pub
--- a/resources/Homestead.json
+++ b/resources/Homestead.json
@@ -1,7 +1,7 @@
 {
     "ip": "192.168.10.10",
     "memory": 2048,
-    "cpus": 1,
+    "cpus": 2,
     "provider": "virtualbox",
     "authorize": "~/.ssh/id_rsa.pub",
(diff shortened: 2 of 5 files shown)


svpernova09:
Thanks!

PR #1141, opened 2019-05-15
Remove automatic www prefix in /etc/hosts entries

cbj4074:
Also, replace last line in serve-modx.sh to match that in serve-laravel.

This is a follow-up to the as yet unresolved comments in #1129 .

Are there any other scripts in `scripts/serve-*` that need this change? Upon cursory analysis, I could find only one, `serve-modx.sh`.


svpernova09:
Thanks!

Issue #1142, opened 2019-05-16
Enhancement to use node projects with pm2 with easy config


MerrySean:
## Versions
Vagrant 2.2.4
Virtualbox 6.0.4
homstead 8.3.1

## Host operating system
MAC OS MOVAJE (10.14.4)

## Problem to solve - Use of node project with watch enabled and hot-reloading specifically (vue) with webpack mix

1. First Step is to install pm2 on your machine
1.1 `vagrant up`
1.2 `vagrant ssh`
1.3 `npm install pm2 -g`
1.4 make sure pm2 is installed properly it should show the version of the install pm2
2. open homestead.rb file ( can be found on scripts folder ) 
2.1 paste the code below AFTER line 287
```
     # generate pm2 JSON config file
          if site['pm2']
            config.vm.provision "shell" do |s2|
              s2.name = 'Creating Site Ecosystem for pm2: ' + site['map']
              s2.path = script_dir + "/create-ecosystem.sh"
              s2.args = Array.new
              s2.args << site['pm2'][0]['name']
              s2.args << site['pm2'][0]['script'] ||= "npm"
              s2.args << site['pm2'][0]['args'] ||= "run serve"
              s2.args << site['pm2'][0]['cwd']
            end
          end
```

3. create a file on scripts folder named create-ecosystem.sh
3.1 Paste the code below 
```
#!/usr/bin/env bash

mkdir /home/vagrant/ecosystem 2>/dev/null

PATH_ECOSYSTEM="/home/vagrant/ecosystem"

rm -rf /home/vagrant/ecosystem/*

PATH_JSON="${PATH_ECOSYSTEM}/${1}.json"

# variables
# 1 = name
# 2 = script
# 3 = args
# 4 = path

BASE_ECO='{
  "name": "'$1'",
  "script": "'$2'",
  "args": "'$3'",
  "cwd": "'$4'"
}'

# Only generate an ecosystem if there isn't one already there.
if [ ! -f $PATH_JSON ]
then
    echo "$BASE_ECO" > $PATH_JSON
fi
```
3.1 **IMPORTANT**: as you could see above we are creating a file on `~/ecosystem` folder
4. Now on your homestead.yaml file, you can now use the kind of syntax seen below, as you can see you can find a pm2.
```
- map: projectName.box
      to: 8080
      pm2:
        - name: "passport"
          script: "npm"
          args: "run serve"
          cwd: "/home/vagrant/projects/vue/projectName"
      type: proxy
```
[...]

PR #1143, opened 2019-05-17
Ability to use pm2 and run node apps on homestead with ease

MerrySean:
this would be a small introduction to the feature

https://github.com/laravel/homestead/issues/1142


svpernova09:
Looks good from here, just need to resolve conflicts.

PR #1152, opened 2019-05-17
remove zray

luketlancaster:
fixes [#1147](https://github.com/laravel/homestead/issues/1147)
(or at least tries to)

I just went for it. Found these blocks being added in #777, so I went ahead an yanked the matching ones. Then all other zray-related things that I found. Feel free to 🔨 this down if it's way off though 🙃

### DIFF
--- a/Homestead.yaml.example
+++ b/Homestead.yaml.example
@@ -37,8 +37,3 @@ databases:
 #       client-id: foo
 #       client-token: bar
 
-# zray:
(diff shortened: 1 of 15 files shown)


svpernova09:
Thanks! Will test drive this later tonight. Just want to double check that since we're removing a param we're passing to the `serve-` scripts that we don't drop anything or assign them incorrectly.

PR #1154, opened 2019-05-18
Add cakephp3 to supported site types

Fuxy22:
It's very similar to the laravel one but this one follows more closely the setting on https://book.cakephp.org/3.0/en/installation.html


svpernova09:
Thanks!

PR #1157, opened 2019-05-21
use PHP 7.3 as default

browner12:
these bash aliases have fallen a little behind with 7.1 as their default.  updating to use 7.3


svpernova09:
Thanks!

PR #1158, opened 2019-05-23
Add support for custom headers in proxy sites

almas1992:
## Add support for custom headers in proxy sites

```yaml
sites:
    - map: exemple.local
      to : 8808
      headers:
          - key: "Access-Control-Allow-Origin"
            value: "*"
          - key: "Access-Control-Allow-Headers"
            value: "*"
      params:
          - key: "proxy_read_timeout"
            value: "600s"
          - key: "proxy_connect_timeout"
            value: "60s"

```

### DIFF
--- a/scripts/serve-proxy.sh
+++ b/scripts/serve-proxy.sh
@@ -1,5 +1,24 @@
 #!/usr/bin/env bash
 
+declare -A params=$6       # Create an associative array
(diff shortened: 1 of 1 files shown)


svpernova09:
This should be expanded to all `serve-*.sh` scripts so that we're always sending the same thing to each site type before being merged.

PR #1160, opened 2019-05-25
Refactor opt in features to use yaml collections

### LINKED ISSUE #1153: [proposal] Utilize collections to do the installs of "optional packages/features"
What you guys think about using collections to do the installs of "optional packages/features".
This could be something like:

```yaml
features: 
  - ohmyzsh
  - ruby
```

Pros:
A new group could make things more organized.
Since AFAIK any of this "optional features" it's only a installation and not a setup, this could facilitate the `homestead.rb`, because if it's follows a pattern, the `homestead.rb` could iterate over the items and call the scripts without checking for each option explicitly.
Make the creation of additional "optional packages/features" more easy, as far as it follows the convention.

Cons:
It's a breaking change, but since homestead 9.0 will be a major version, this could be more "forgiving"
Needs development of a "pattern" to follow for easy integration.

ricardoseriani:
Close  #1153


svpernova09:
image

Thanks!

PR #1162, opened 2019-05-26
Fix nginx unable to start on VM restart

Fuxy22:
Because this config relies on the site path `/vagrant` already being mounted when nginx start it will fail when restarting the VM since the server will attempt to start way sooner.

This will just restart it 40 seconds later.

Ideally i would like to start nginx after `vagrant.mount` but haven't found anything that works yet.

### DIFF
--- a/scripts/serve-cakephp3.sh
+++ b/scripts/serve-cakephp3.sh
@@ -105,6 +105,21 @@ server {
 }
 "
 
+# Restart nginx if it started before vagrant path was mounted
+# (happens after stopping and restarting the VM)
+timer="[Unit]
+Description=Nginx delay
+
+[Timer]
+OnStartupSec=40sec
+
+[Install]
+WantedBy=timers.target
+"
+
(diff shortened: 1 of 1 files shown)


svpernova09:
Why is this needed? 

We manually restart nginx after we've configured all the virtual hosts, which happens after folders have been mounted.

Fuxy22:
When you provision yes but once you do `vagrant halt` and then `vagrant up` the nginx server will start before the folders have been mounted and it fails.

On second thought maybe relying on the `/vagrant` folder to be mounted not a good idea maybe I should just create some symbolic links to the logs instead.

What do you think?

svpernova09:
If nginx doesn't start up after a vagrant halt, that's a different issue. We need to see what is causing that and resolve it. Going to hold off on this PR.

Fuxy22:
I know what the issue is. Its nginx trying to start before `/vagrant` is mounted which the config for the cakephp site requires.

Since its not provisioning the order by default is reversed.

PR #1166, opened 2019-05-27
Group opt-in lockfiles

### LINKED ISSUE #1165: [proposal] Create a script to do setup before installation of optional features
I noted that some opt-in features set a file in `/home/vagrant/<.feature_name>` to mark that feature as `already installed`.
I don't know if the correct is to all feature to create this kind of file, but if so, what you guys think about creating a script that do this creation/check and only then, call the feature script to continue with the installation.

Pros:
The installation scripts will only handle the installation, avoiding that any script should do the creation/check of that file.
Could group this file in some folder, to avoid cluttering the home folder (and avoid conflicts too).
The development of scripts will be simpler.

If you guys think that this is OK, I can make a PR.

ricardoseriani:
Close #1165

### DIFF
--- a/scripts/features/chronograf.sh
+++ b/scripts/features/chronograf.sh
@@ -2,14 +2,14 @@
 
 # Check If Chronograf Has Been Installed
 
-if [ -f /home/vagrant/.chronograf ]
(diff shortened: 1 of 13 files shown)


svpernova09:
Thanks!

PR #1163, opened 2019-05-27
Implement support for Apache Solr Search Database

stefnats:
To activate Solr just use in Homestead.yaml:

    solr: true

And open the port on: 

    ports:
        - send: 18983
          to: 8983

This installs Solr 7.7.1 per default, current version is 8.1.0 (as of today). I have not much experience with the 8.* branch, so i implemented the latest 7.* version. It also installs a default "homestead" search core.
This is one of my first PR, if you have any improvements or suggestions, let me know.

-Stefan

PR #1169, opened 2019-05-28
Implement Apache Solr as optional feature

stefnats:
As discussed yesterday in PR #1163 this is now for Homestead 9.

To activate Solr just use in Homestead.yaml:

    features:
        - solr: true

And open the port on:

    ports:
        - send: 18983
          to: 8983

This installs Solr 7.7.1 per default, current version is 8.1.0 (as of today). I have not much experience with the 8.* branch of Solr, so i implemented the latest 7.* version. It also installs a default "homestead" search core.

-Stefan


scripts/features/solr.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+# Install Java Runtime Enviroment
+sudo apt-get update
```
browner12: @svpernova09 can correct me if I'm wrong, but I believe we decided on switching to `apt` instead of `apt-get` for v9.

stefnats:
Didn't know that! Corrected.

scripts/features/solr.sh
svpernova09: I'm somewhat torn, while I would like to switch to apt, and we currently have in V9, even apt gives CLI warnings that the CLI interface isn't stable, which has me rethinking it.

svpernova09:
Thanks!

scripts/features/solr.sh
browner12: interesting answer here. https://askubuntu.com/questions/990823/apt-gives-unstable-cli-interface-warning

maybe we were wrong on this.  sounds like we should use things like `apt-get` and `apt-cache` for scripts, but users running straight from the command line should use `apt`.

PR #1170, opened 2019-05-28
Implement Apache Cassandra as a feature

stefnats:
Installation of a single node Cassandra cluster & compilation of PHP extension to work with all PHP 7.* versions.

Enable via:

    features:
        - cassandra: true

And open ports:

    ports:
        - send: 19042
          to: 9042
        - send: 19142
          to: 9142
        - send: 19160
          to: 9160

The PHP Extension needs to be compiled, because current release does not support PHP 7.3, but git master does. 
Looking forward to some feedback.


svpernova09:
Thanks

PR #1173, opened 2019-06-02
Remove the hosts add section since there is a script for it now

Fuxy22:
Reove hosts add section from serve scrypt


svpernova09:
Thanks

PR #1175, opened 2019-06-03
Added script to support site type "wordpress"

coolamit:
Added script to support site type `wordpress` which:

1. Sets up Nginx rules for WordPress
2. Downloads latest stable version of WordPress
3. Installs WP_Cli
4. Generates `wp-config.php` for the WordPress site

### DIFF
--- a/scripts/serve-wordpress.sh
+++ b/scripts/serve-wordpress.sh
@@ -1,0 +1,153 @@
+#!/usr/bin/env bash
+
+declare -A params=$6       # Create an associative array
+declare -A headers=${10}   # Create an associative array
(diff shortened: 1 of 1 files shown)


svpernova09:
Wp cli is already installed in the base box and doesn’t need to be reinstalled. We could run a self update command here though.

--
- Joe Ferguson
JoeFerguson.me
osmihelp.org

coolamit:
@svpernova09 Good idea. I've removed the code for WP_Cli installation & added self-update so that WP_Cli would update to latest stable release if one is available.

svpernova09:
Thanks!

PR #1177, opened 2019-06-03
Fix webdriver installation

staudenmeir:
The `webdriver` feature doesn't work for me. When I add `webdriver: true` to my `Homestead.yaml`, I get 404 errors for multiple packages.

Running `apt-get update` beforehand fixes it.


svpernova09:
thanks!

El-Mehdi-Abdi:
great man thanks

PR #1180, opened 2019-06-05
Implement Gearman Job Server as optional feature

stefnats:
Installation of a [gearman job server](http://gearman.info/) instance which listens on Port 4730 and installation of the [PHP Extension (php-gearman)](https://pecl.php.net/package/gearman)

Enable via:

    features:
        - gearman: true

And open ports:

    ports:
        - send: 4730
          to: 14730

I was not sure which `apt` or `apt-get` is preferred now. Is there a clear direction?


svpernova09:
We should continue to use apt-get for all packages.

Can you have a look at how we implemented features (scripts/features) in branch version_9 and update this PR to that format?

--
- Joe Ferguson
JoeFerguson.me
osmihelp.org

stefnats:
Hey Joe, is this what you mean or what do you mean exactly?
When this is merged, i can go ahead and make sure every scripts/features/*.sh script in version_9 uses `apt-get`

svpernova09:
Thanks!

PR #1182, opened 2019-06-06
Implement RabbitMQ Message Broker as optional Feature

stefnats:
This implements the [RabbitMQ Message Broker](https://www.rabbitmq.com/) as optional feature.

Default credentials are:
- Username: `homestead`
- Password: `homestead`

There is a HTTP Admin Interface at http://localhost:15673 if enabled with the following settings in Homestead.yaml
The actual AMQP Port is opened at `5673`:

```yaml
features:
    - rabbitmq: true

ports:
    - send: 5673
      to: 5672
    - send: 15673
      to: 15672
```


stefnats:
Should the php extensions, if possible, installed for 7.1, 7.2 and 7.3? Currently only the `php-amqp` and `php-bcmath` are installed, not for a specific PHP version, but for the current (from ondrej/php launchpad repo)

svpernova09:
Yes if possible. But there may not be specific versions as in the case of `php-amqp`.

Can we set the password to `secret` so it lines up with the Homestead defaults?

stefnats:
Whoops, didn't see that there are no specific versions for `php-amqp` in the `ondrej/php` repository.
Changed default password to `secret`. User is still `homestead`

scripts/features/rabbitmq.sh
```diff
@@ -0,0 +1,33 @@
...
+echo "deb https://packages.erlang-solutions.com/ubuntu bionic contrib" | sudo tee /etc/apt/sources.list.d/rabbitmq.list
+sudo apt-get update
+sudo apt-get -y install erlang
+
+# Install RabbitMQ
+
+wget -q -O- https://dl.bintray.com/rabbitmq/Keys/rabbitmq-release-signing-key.asc | sudo apt-key add -
+wget -q -O- https://www.rabbitmq.com/rabbitmq-release-signing-key.asc | sudo apt-key add -
+echo "deb https://dl.bintray.com/rabbitmq/debian $(lsb_release -sc) main" | sudo tee /etc/apt/sources.list.d/rabbitmq.list
+sudo apt-get update
```
svpernova09: We should group the statements we need to do *before* `apt-get update` so we only do the update once. This will save time during provisioning.

stefnats: You are right. That's now corrected.

svpernova09:
Thanks!

PR #1185, opened 2019-06-14
add install-openresty.sh

amorZhu:
openresty: true


amorZhu:
```shell

vagrant@homestead:~$ ps -ef | grep nginx
root      1055     1  0 06:16 ?        00:00:00 nginx: master process /usr/local/openresty/nginx/sbin/nginx -g daemon on; master_process on;
root      1141     1  0 06:16 ?        00:00:00 nginx: master process /usr/sbin/nginx -g daemon on; master_process on;
vagrant   1143  1141  0 06:16 ?        00:00:00 nginx: worker process
nobody    1155  1055  0 06:16 ?        00:00:00 nginx: worker process
vagrant  22003 18751  0 06:25 pts/0    00:00:00 grep --color=auto nginx
```

amorZhu:
openresty port 8888.

amorZhu:
address 127.0.0.1:8888

browner12:
Can you give a brief explanation of what OpenResty is, and why we either need it, or why it would be nice to have?

The easier you can make it on maintainers to understand the purpose and benefit of the PR, the better.

svpernova09:
I'm torn, I have no idea what this is, I don't want to have to support it, but at the same time, there's not a "good" reason to reject it.

browner12:
I personally think that's a good reason to hold up on this. People PR features in, but Joe, you have to maintain them.

Let's wait to see the explanation.  Maybe we want to add a PR template requiring people to add the ***why***, not just the ***what*** to their PRs.

While I'm not sure we can define a specific objective cutoff for features to add in, we do need to hit some kind of subjective level of cost/benefit, otherwise what's stopping everything plus the kitchen sink getting added to Homestead.

svpernova09:
> I personally think that's a good reason to hold up on this. People PR features in, but Joe, you have to maintain them.

\o/

> Let's wait to see the explanation. Maybe we want to add a PR template requiring people to add the why, not just the what to their PRs.

I'll look into PR templates

> While I'm not sure we can define a specific objective cutoff for features to add in, we do need to hit some kind of subjective level of cost/benefit, otherwise what's stopping everything plus the kitchen sink getting added to Homestead.

I'm much more accepting of these kinds of things because they're opt-in and we have the foundation and workflow established to facilitate these kinds of easy features. My main concern is if someone has an issue I won't be able to help debug anything past verifying the system was provisioned properly.

amorZhu:
https://openresty.org/en/

amorZhu:
> I'm torn, I have no idea what this is, I don't want to have to support it, but at the same time, there's not a "good" reason to reject it.

Users can choose whether to use or not by configuration.
In the homestead box only nginx has no openresty
You may not use much, but there are many people in China.

browner12:
Providing a link to the docs isn't the same as giving a summarized explanation. Most maintainers don't have the time to read through all the documentation to understand what something does, so it's your job to give a simplified explanation, and a reason why it would be beneficial to the project.

svpernova09:
Thanks for the PR.

PR #1192, opened 2019-06-20
Issue #1191 Avoid running ohmyzsh install more than once and correct the git URL

### LINKED ISSUE #1191: oh my zsh support broken and doesn't handle repeated provisioning
## Versions
- Vagrant: 2.2.4
- Provider: Virtualbox 5.2.26
- Homestead: 8.5.3

## Host operating system
macOS High Sierra 10.13.6

## Homestead.example.yaml
```yaml
ip: 192.168.10.10
memory: 2048
cpus: 4
provider: virtualbox
authorize: ~/.ssh/id_rsa.pub
keys:
    - ~/.ssh/id_rsa
folders:
    -
        map: .
        to: /home/vagrant/code
        type: rsync
        options:
            rsync__exclude: [".git/", ".idea", "web/sites/*/files", "web/sites/*/files-private", "web/artifacts", "web/sites/simpletest"]
            rsync__args: ["--verbose", "--archive", "--delete", "-z", "--chmod=g+rwX"]
sites:
    -
        map: mysite.local
        to: /home/vagrant/code/web
        type: apache
        php: '7.3'
        xhgui: 'true'
databases:
    - mysite
name: mysite
hostname: mysite.local
mongodb: true
ohmyzsh: true
variables:
    -
        key: ZSH_THEME
        value: robbyrussell
```

## Expected behavior

Should install oh my zsh as the shell with the robbyrussell theme.

## Actual behavior

I get errors on provision:

```
==> mysite: Running provisioner: shell...
    mysite: Running: script: Installing Oh-My-Zsh
    mysite: Cloning into '/home/vagrant/.oh-my-zsh'...
    mysite: fatal: unable to connect to github.com:
    mysite: github.com[0: 192.30.253.113]: errno=Connection timed out
    mysite: cp: cannot stat '/home/vagrant/.oh-my-zsh/templates/zshrc.zsh-template': No such file or directory
```
Also as I debugged and ran `vagrant provision` multiple times, I noticed that the script for installing oh my zsh doesn't check for a prior install and not all the commands are idempotent.

m4olivei:
Resolves #1191.

### DIFF
--- a/scripts/install-ohmyzsh.sh
+++ b/scripts/install-ohmyzsh.sh
@@ -9,6 +9,7 @@ then
 fi
 
 touch /home/vagrant/.ohmyzsh
+chown -Rf vagrant:vagrant /home/vagrant/.ohmyzsh
 
 git clone https://github.com/robbyrussell/oh-my-zsh.git /home/vagrant/.oh-my-zsh
 cp /home/vagrant/.oh-my-zsh/templates/zshrc.zsh-template /home/vagrant/.zshrc
--- a/scripts/install-ohmyzsh.sh
+++ b/scripts/install-ohmyzsh.sh
@@ -2,7 +2,13 @@
 
 # Install oh-my-zsh
 
-git clone git://github.com/robbyrussell/oh-my-zsh.git /home/vagrant/.oh-my-zsh
+if [ -f /home/vagrant/.oh-my-zsh ]
+then
+    echo "oh my zsh already installed."
+    exit 0
+fi
+
+git clone https://github.com/robbyrussell/oh-my-zsh.git /home/vagrant/.oh-my-zsh
 cp /home/vagrant/.oh-my-zsh/templates/zshrc.zsh-template /home/vagrant/.zshrc
 printf "\nemulate sh -c 'source ~/.bash_aliases'\n" | tee -a /home/vagrant/.zprofile
 printf "\nemulate sh -c 'source ~/.profile'\n" | tee -a /home/vagrant/.zprofile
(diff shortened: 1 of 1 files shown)


scripts/install-ohmyzsh.sh
```diff
@@ -2,7 +2,13 @@
 
 # Install oh-my-zsh
 
-git clone git://github.com/robbyrussell/oh-my-zsh.git /home/vagrant/.oh-my-zsh
+if [ -f /home/vagrant/.oh-my-zsh ]
+then
+    echo "oh my zsh already installed."
+    exit 0
+fi
+
```
svpernova09: Need to add a line here that touches the `/home/vagrant/.oh-my-zsh` like we do in other cases.

svpernova09: Oh wait, no, we need to NOT rely on the `.oh-my-zsh` folder existing and instead touch another file. This is the paradigm we're moving towards for v8

m4olivei: Ok cool. You mean like this for example right?

https://github.com/laravel/homestead/blob/master/scripts/install-mysql8.sh#L12

svpernova09: Exactly!

m4olivei:
There we go. Tested and working for me. I used the name `.ohmyzsh` for the flag file, which is slightly awkward b/c it sits beside `.oh-my-zsh` which is the source directory for oh my zsh. I think it's OK, but open to suggestions.

svpernova09:
>  I used the name .ohmyzsh for the flag file, which is slightly awkward b/c it sits beside .oh-my-zsh which is the source directory for oh my zsh. I think it's OK, but open to suggestions.

Yeah, in v8 we shove these into another folder so it's less awkward.

svpernova09:
Thanks!

PR #1193, opened 2019-06-21
Allow configuration of the default PHP version used for the CLI

mikelerch:
The PHP CLI version can be specified by adding the following key in your Homestead config:
```
php: "7.2" # Or any other supported version
```
I often find myself adding a custom provisioning step in `after.sh` to switch the CLI version during provisioning, which should instead be easily configurable.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -203,6 +203,14 @@ class Homestead
       end
     end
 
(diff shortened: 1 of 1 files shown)


svpernova09:
Would you please make another PR against `version_9` branch with this change.

PR #1198, opened 2019-06-24
Attempt to restart the web server if it's not started

Fuxy22:
If a site config relies on vagrant being mounted and it isn't when the server is starting the server won't start.

This is just as small script that attempts to restart the server if it detects it isn't running otherwise it won't do anything.

By the time this is running everything will be mounted.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -612,6 +612,11 @@ class Homestead
       s.inline = 'sudo service motd-news restart'
     end
 
+    config.vm.provision 'shell', run: "always" do |s|
+        s.name = 'Restart Nginx'
+        s.path = script_dir + '/server-reset.sh'
+    end
+
     if settings.has_key?('backup') && settings['backup'] && (Vagrant::VERSION >= '2.1.0' || Vagrant.has_plugin?('vagrant-triggers'))
       dir_prefix = '/vagrant/'
       settings['databases'].each do |database|
--- a/scripts/server-reset.sh
+++ b/scripts/server-reset.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+#
+# Check if web server is running and if not restart it.
+# In case some site setting required mount directories to be present and made it unable to start
+
+systemctl is-active --quiet nginx.service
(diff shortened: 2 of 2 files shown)


svpernova09:
> If a site config relies on vagrant being mounted and it isn't when the server is starting the server won't start.

Web servers are only started *after* folder shares have been mounted. It shouldn't be possible to restart services *before* all of the folders are mounted. If you are seeing this, you should open a bug report. I'd guess it's something weird with your system resuming from sleep/hibernate.

We also can't take this change because it would cause issues with users who aren't using Nginx (Those who are using site type `apache`)

Fuxy22:
This isn't on the first provision when all script are run. This is once the VM has been halted once and is beig brought up again without re provisioning.

Although you have a point with non nginx users i'l fix that

svpernova09:
> This isn't on the first provision when all script are run. This is once the VM has been halted once and is beig brought up again without re provisioning.

I've never been able to reproduce this. I've tested on every provider we support and nearly every platform.

I get this can be a *really* annoying problem. The best solution I can offer would use `after.sh` to add a startup script that is run later in the boot process to restart nginx.

Fuxy22:
The key to reproducing it if having a site config that requires the `/vagrant` mount to be present when the web server is booting up like https://github.com/laravel/homestead/blob/01227bf1fdcddf61ad1238940fbfb538fc6d953b/scripts/serve-cakephp3.sh#L82

If it is not mounted yet the server won't start from my experience.

deviantintegral:
Interesting, so with nginx the logs are saved to `/vagrant`? Is that all site types or just cake? I know with apache they're stored in the normal `/var/log/apache2/...` location. Given AFAIK all other logs are in `/var/log`, perhaps the simplest solution would be to store nginx logs there too.

svpernova09:
> The key to reproducing it if having a site config that requires the /vagrant mount to be present when the web server is booting up like 

User shouldn't be writing anything to `/vagrant`. Your logs and any other data you want to keep from the VM should be stored in a user specified folder share.

PR #1203, opened 2019-06-29
Correct path for xhgui and mongo scripts

deviantintegral:
This isn't working yet as it looks like there's an error installing php-mongodb:

```
    homestead-d8: Installing shared extensions:     /usr/lib/php/20131226/
    homestead-d8: chmod:
    homestead-d8: cannot access '/usr/lib/php/20160303/mongodb.so'
    homestead-d8: : No such file or directory
    homestead-d8: Reading package lists...
    homestead-d8: Building dependency tree...
    homestead-d8:
    homestead-d8: Reading state information...
    homestead-d8: php7.0-dev is already the newest version (7.0.33-8+ubuntu18.04.1+deb.sury.org+1).
    homestead-d8: 0 upgraded, 0 newly installed, 0 to remove and 12 not upgraded.
    homestead-d8: Configuring for:
    homestead-d8: PHP Api Version:         20151012
    homestead-d8: Zend Module Api No:      20151012
    homestead-d8: Zend Extension Api No:   320151012
    homestead-d8: Installing shared extensions:     /usr/lib/php/20151012/
    homestead-d8: chmod: cannot access '/usr/lib/php/20160303/mongodb.so': No such file or directory
```

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -334,11 +334,11 @@ class Homestead
 
           if site['xhgui'] == 'true'
             config.vm.provision 'shell' do |s|
-              s.path = script_dir + '/install-mongo.sh'
+              s.path = script_dir + '/features/mongodb.sh'
             end
 
             config.vm.provision 'shell' do |s|
-              s.path = script_dir + '/install-xhgui.sh'
+              s.path = script_dir + '/features/xhgui.sh'
             end
 
             config.vm.provision 'shell' do |s|


deviantintegral:
As well, it looks like somehow the `cd` in the xhgui script is being preserved in the environment, causing composer to be installed to `/opt/xhgui`.

```
homestead-d8: Composer (version 1.8.6) successfully installed to: /opt/xhgui/composer.phar
    homestead-d8: Use it: php composer.phar
```

deviantintegral:
The additional composer install is forced by xhgui's install script, so is nothing to worry about: https://github.com/perftools/xhgui/blob/master/install.php

@svpernova09 What's the `chmod` for in the mongo script? https://github.com/laravel/homestead/commit/94835aa413bbaf03d89fd7078a4d7dfb68d814d0#diff-d24827c433f0990a98a73662b4713b3aR39

At a glance, removing those doesn't cause any problems with the extension.

I wonder if we need some cleanup or docs updates here. `xhgui.sh` is in `features`, but turning it on there breaks things. Instead, it must be in the site definition.

Feel free to push commits to my branch, as I likely won't get back to this until tomorrow or Tuesday.

svpernova09:
Resolved via #1203

PR #1204, opened 2019-06-29
Add ability to import CA to host

Fuxy22:
Add ability to import homestead CA into host trusted repository while homestead is running just by setting `import_ca: true` in your `Homestead.yaml` file

- Currently only working on host systems that use `trust` like Arch Linux

If someone want to adapt it for other systems here is a doc https://github.com/rcrowley/certified/wiki/Trust-your-CA-on-Linux fo other flavors i don't have a system to test it on at the moment.

Note: It will remove it whenever you do `vagrant halt` or `vagrant destroy` so there is little risk of it being exploited especially since the certificate get regenerated whenever the VM gets destroyed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ phpunit.xml
 /mysql_backup
 /postgres_backup
 /user-customizations.sh
+/.ca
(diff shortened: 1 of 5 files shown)


Fuxy22:
P.S. This is useful for me in case i want to run the chrome lighthouse plugin on https without it crapping itself that the certificate is not trusted but it's mostly a nice to have.

PR #1205, opened 2019-06-29
Add ability to import CA to host

Fuxy22:
Add ability to import homestead CA into host trusted repository while homestead is running just by setting `import_ca: true` in your `Homestead.yaml` file

- Currently only working on host systems that use `trust` like Arch Linux

If someone want to adapt it for other systems here is a doc https://github.com/rcrowley/certified/wiki/Trust-your-CA-on-Linux fo other flavors i don't have a system to test it on at the moment.

Note: It will remove it whenever you do `vagrant halt` or `vagrant destroy` so there is little risk of it being exploited especially since the certificate get regenerated whenever the VM gets destroyed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ phpunit.xml
 /mysql_backup
 /postgres_backup
 /user-customizations.sh
+/.ca
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -548,12 +548,12 @@ class Homestead
       end
 
       config.trigger.after :up do |trigger|
-        trigger.info = "Copying CA files to share..."
+        trigger.info = "Importing CA files..."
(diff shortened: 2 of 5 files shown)


svpernova09:
Thanks for taking the time to open a PR. This is something that Homestead intentionally does not facilitate because users should have to do the research and know the pros and cons to what's involved in trusting a CA from an untrusted source (Vagrant/Homestead) on their host machines.

Fuxy22:
There is effectively no risk though... the CA cert is constantly changing if you provision often and it is only kept while the VM is running.

The only way to exploit this is if your host gets hacked and the CA cert gets stolen at which point you have bigger problems.

And this is only valid until you next destroy your VM since a new cert will be generated after.

PR #1213, opened 2019-07-02
Change features parser to provide environment variables

### LINKED ISSUE #1212: Problem when installing blackfire feature
## Versions
  - Vagrant: Vagrant 2.2.5
  - Provider: 6.0.8
  - Homestead: master or release.

## Host operating system
Ubuntu 16.04

## Homestead.yaml
```yaml
---
ip: "192.168.10.10"
memory: 2048
cpus: 2
provider: virtualbox

authorize: ~/.ssh/id_rsa.pub

keys:
    - ~/.ssh/id_rsa

folders:
    - map: ~/sites
      to: /home/vagrant/sites

sites:
    - map: homestead.test
      to: /home/vagrant/sites/homestead/public

databases:
    - homestead

features:
    - blackfire: "server_id" "server_token" "client_id" "client_token"
```

## Vagrant destroy & up output
https://gist.github.com/ricardoseriani/1dfe7ebeac467a9d4e86176776ea7b45

## Expected behavior
Should install blackfire with the credentials provided

## Actual behavior
Vagrant up is broken due to errors when parsing the Homestead.yaml

## Steps to reproduce
1. vagrant destroy
2. config Homestead.yaml as defined
3. vagrant up

The current blackfire.sh script uses the variables $1, $2, $3, $4, but AFAIK yaml can't understand the yaml using multiple values in the same key, like in the example.

I can think of some approaches:
 - We can set some kind of "separator" in the parsing of yaml features value, but this can create some problems with the separator being used in the string value
 - Create a new section with the variables, parse it before all the features try to run, and use that in the script, but will make less explicit that this value refers to that feature
 - Create some kind of sub-parameters in each feature, but this will make the features harder to create and to parse

Any thoughts on this ?

ricardoseriani:
Close #1212 

### DIFF
--- a/scripts/features/blackfire.sh
+++ b/scripts/features/blackfire.sh
@@ -13,24 +13,24 @@ wget -q -O - https://packages.blackfire.io/gpg.key | apt-key add -
 echo "deb http://packages.blackfire.io/debian any main" | tee /etc/apt/sources.list.d/blackfire.list
 
 # Install Blackfire
-
-apt-get install -y blackfire-agent blackfire-php
+sudo DEBIAN_FRONTEND=noninteractive apt-get update
+sudo DEBIAN_FRONTEND=noninteractive apt-get install -y blackfire-agent blackfire-php
 
 agent="[blackfire]
 ca-cert=
 collector=https://blackfire.io
 log-file=stderr
 log-level=1
-server-id="$1"
-server-token="$2"
+server-id=${server_id}
+server-token=${server_token}
 socket=unix:///var/run/blackfire/agent.sock
 spec=
 "
 
 client="[blackfire]
 ca-cert=
-client-id="$3"
-client-token="$4"
+client-id=${client_id}
+client-token=${client_token}
 endpoint=https://blackfire.io
 timeout=15s
 "
--- a/scripts/features/elasticsearch.sh
+++ b/scripts/features/elasticsearch.sh
@@ -14,7 +14,7 @@ chown -Rf vagrant:vagrant /home/vagrant/.homestead-features
 # Determine version from config
 
 set -- "$1"
-IFS="."; declare -a version=($*)
+IFS=".";
 
 if [ -z "${version[1]}" ]; then
     installVersion=""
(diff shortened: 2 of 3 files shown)


ricardoseriani:
This allow to use the `features` in the following syntax:

```yaml
features:
    - blackfire: 
        server_id: "server_id"
        server_token: "server_value"
        client_id: "client_id"
        client_token: "client_value"
    - golang: true
    - docker: false
    - elasticsearch:
        version: 7
```

In this case, it will set the variables like this:
 - blackfire.sh: `server_id = server_id`, `server_token = server_token`, `client_id = client_id` and `client_token = client_token`
 - golang.sh: not set any variable, but continue with the installation
 - docker.sh: ignore this, since its set to `false`
 - elasticsearch.sh: `version` will be set to 7

ricardoseriani:
If this is OK, I can check the features scripts and try to adjust then.
PS: Do you think that it's better to change all variables to lowercase, to avoid any misconfiguration (only the key part, since the value can be case sensitive) ?

svpernova09:
> If this is OK, I can check the features scripts and try to adjust then.

Yeah I think this looks good.

> PS: Do you think that it's better to change all variables to lowercase, to avoid any misconfiguration (only the key part, since the value can be case sensitive) ?

Part of me wants env variables to be all uppercase, but that might be confusing when presented to the user in Yaml. I think it would be easier to keep them lowercase. Would be open to hearing other views.

ricardoseriani:
> > If this is OK, I can check the features scripts and try to adjust then.
> 
> Yeah I think this looks good.

I will check the scripts and adjust then.


> > PS: Do you think that it's better to change all variables to lowercase, to avoid any misconfiguration (only the key part, since the value can be case sensitive) ?
> 
> Part of me wants env variables to be all uppercase, but that might be confusing when presented to the user in Yaml. I think it would be easier to keep them lowercase. Would be open to hearing other views.

Well, I'm going to assume that they all be lowercase in the scripts, and if needed, we can convert then before applying the environment variables, ok ?

svpernova09:
> Well, I'm going to assume that they all be lowercase in the scripts, and if needed, we can convert then before applying the environment variables, ok ?

Sounds good! Thanks

ricardoseriani:
AFAIK the only scripts that use variables/parameters are `blackfire.sh` and `elasticsearch.sh`

svpernova09:
thanks!

PR #1214, opened 2019-07-02
Change golang.sh lockfile name to use same pattern as other scripts

ricardoseriani:


ricardoseriani:
This change the lockfile name to use the same name as the script "per se", as in other features files.

svpernova09:
Thanks!

PR #1220, opened 2019-07-05
update stubs to include "features" example

browner12:
this will help users get used to the new "features" syntax for optional software.

I picked the "webdriver" because it's something I imagine a relatively good portion of our users will turn on.

documentation is currently being updated as well for "features".

please confirm my JSON syntax is correct. Wasn't sure if we're supposed to use an object or an array of objects?

### DIFF
--- a/Homestead.yaml.example
+++ b/Homestead.yaml.example
@@ -21,6 +21,8 @@ databases:
     - homestead
 
 features:
+    - mariadb: false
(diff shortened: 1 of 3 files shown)


svpernova09:
JSON looks good from here

Let's add ohmyzsh, mariadb, both set to false, as more examples.

svpernova09:
Thanks

PR #1221, opened 2019-07-05
Moved serve-wordpress.sh to site-types/wordpress.sh

justo:
Looks like it might have just been missed when all the scripts were moved into the site-types folder. `wordpress` should probably be added to the "Site Types" list in the documentation as well: https://laravel.com/docs/master/homestead#adding-additional-sites


svpernova09:
Thanks!

PR #1262, opened 2019-08-19
Minio conditional statement and bucket location change

rayblair06:
Buckets created in the folder "homestead" no longer seem to appear within the Minio GUI, instead changed to create the buckets where directly referenced.

Also changed the conditional statement within the initial script to reflect the recent structure changes to "features" array within .yaml.

### DIFF
--- a/scripts/create-minio-bucket.sh
+++ b/scripts/create-minio-bucket.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-mc mb homestead/$1
-mc policy set $2 homestead/$1
+mc mb /usr/local/share/minio/$1
+mc policy set $2 /usr/local/share/minio/$1
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -518,7 +518,7 @@ class Homestead
     end
 
     # Create Minio Buckets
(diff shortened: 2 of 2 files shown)


svpernova09:
Is this going to be the last minio PR for a bit? Can I get another Minio user to sign off in this?

--
- Joe Ferguson
JoeFerguson.me
osmihelp.org

rayblair06:
I'm confident that should be the last Minio PR for now and I'm happy to wait until another user can sign it off.

svpernova09:
Thanks @rayblair06! I'm on vacation with limited internet so wanted to make sure we didn't ship something that I couldn't quickly fix :D

rayblair06:
Of course, understandable @svpernova09 :) Hope you have a good holiday!

kennith:
@rayblair06 if the script change the policy directly to the bucket in the filesystem, it will not apply the policy when accessing files via Url. 

I think it is better if we set both policy on the filesystem and url in `create-minio-bucket.sh` 

```bash
mc policy set $2 /usr/local/share/minio/$1
mc policy set $2 homestead/$1/
```

Since there is a back and forth on changing the `create-minio-bucket.sh`, I think this will be a good solution to meet both requirements. What do you and @svpernova09 think?

PR #1269, opened 2019-09-04
Feature: swoole support (PHP 7.1+)

webimpress:
As we have support for Zend Expressive website type here would be nice to add also swoole module to PHP. It can be used with: https://github.com/zendframework/zend-expressive-swoole

[Since v4.4.0](https://github.com/swoole/swoole-src/tree/v4.4.0) it supports only PHP 7.1.
Not sure if there is a point to install previous version to support older PHP versions?

Also would be nice to add [inotify](https://pecl.php.net/package/inotify) support but I don't know how to do it for different PHP versions. It is also used with swoole.

I do have also another problem: SOCK_DGRAM and SOCK_STREAM connections are not working. When I am using that socket type and trying to start server I am getting: `Error: Operation not permitted`. I've tried `sudo`, different ports, params... The thing is that on xenial it works without issues, and on bionic I can't get it working for some reason. Maybe someone here will know why and how to fix it?

Thanks!

### DIFF
--- a/scripts/features/swoole.sh
+++ b/scripts/features/swoole.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+export DEBIAN_FRONTEND=noninteractive
+
+if [ -f /home/vagrant/.homestead-features/swoole ]
+then
+    echo "Swoole already installed."
+    exit 0
+fi
+
+touch /home/vagrant/.homestead-features/swoole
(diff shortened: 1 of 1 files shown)


svpernova09:
Going to hold off on this for now, would rather implement as much as we can in Settler, then implement the rest here, but I'd like to spend some time with this before we get it merged so i know what we're getting into. Thanks for the PR

Issue #1278, opened 2019-09-24
share (and ngrok) doesn't work on HyperV


nicolus:
## Versions
Homestead 9.1

## Host operating system
Windows 10

## Expected behavior
Running `share domain` or `ngrok http -host-header=domain --log=stdout` should start ngrok

## Actual behavior
ngrok fails silently, or with the `--log=stdout` argument we can see the error

```
lvl=crit msg="command failed" err="listen tcp 192.168.10.10:4040: bind: cannot assign listen tcp 192.168.10.10:4040: bind: cannot assign requested address
```

The issue is that the [create-ngrok](https://github.com/laravel/homestead/blob/v9.1.0/scripts/create-ngrok.sh) script set the WebUI to use the IP configured in homestead.yaml. When using HyperV this IP is completely ignored by Vagrant and a new IP is assigned to the virtual machine.

set0x:
* Run in terminal for fix
```
LOCAL_IP=$(ifconfig | grep -Eo 'inet (addr:)?([0-9]*\.){3}[0-9]*' | grep -Eo '([0-9]*\.){3}[0-9]*' | grep -v '127.0.0.1')
PATH_NGROK="/home/vagrant/.ngrok2"
PATH_CONFIG="${PATH_NGROK}/ngrok.yml"
sed -i "/^web_addr:/c\web_addr: ${LOCAL_IP}:4040" ${PATH_CONFIG}
```

PR #1279, opened 2019-09-24
Dynamically set ngrok webUI IP (fixes share with HyperV)

nicolus:
This proposal fixes https://github.com/laravel/homestead/issues/1278

I'm not entirely sure this should be merged as it adds a pretty obscure regex and rewrites a config file on the fly just to fix a limitation of Vagrant with HyperV.

Maybe mentioning the issue in the docs and proposing a fix (manually editing the file to replace the IP or set `web_addr` to false to disable the WebUI altogether) would be enough ?

### DIFF
--- a/resources/aliases
+++ b/resources/aliases
@@ -192,7 +192,7 @@ function share() {
         PATH_NGROK="/home/vagrant/.ngrok2"
         PATH_CONFIG="${PATH_NGROK}/ngrok.yml"
 
-        sed -i "/web_addr:/c\web_addr: ${LOCAL_IP}:4040" ${PATH_CONFIG}
+        sed -i "/^web_addr:/c\web_addr: ${LOCAL_IP}:4040" ${PATH_CONFIG}
 
         ngrok http ${@:2} -host-header="$1" 80
     else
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for taking the time to try to fix this. 

The root cause here is Microsoft still doesn't expose networking through the Hyper-V APIs. I'd love to provide a better solution, I'm just extremely worried about obscure regex and not easily being able to debug things. 

I'm all for adding a section to the docs to discuss this as a known issue with Hyper-V

set0x:
* Run in terminal for fix
```
LOCAL_IP=$(ifconfig | grep -Eo 'inet (addr:)?([0-9]*\.){3}[0-9]*' | grep -Eo '([0-9]*\.){3}[0-9]*' | grep -v '127.0.0.1')
PATH_NGROK="/home/vagrant/.ngrok2"
PATH_CONFIG="${PATH_NGROK}/ngrok.yml"
sed -i "/^web_addr:/c\web_addr: ${LOCAL_IP}:4040" ${PATH_CONFIG}
```

PR #1299, opened 2019-10-16
Remove site-specific client_max_body_size

cbj4074:
Homestead would provide greater flexibility if this value were simply removed.

Currently, there is no straightforward means by which to override this hard-coded, site-specific value, which causes large file uploads to fail with HTTP status code 413 in NGINX.

If this directive is simply removed, the end-user is able to add a preferred value to e.g. `/home/vagrant/.config/nginx/nginx.conf`, drop an override file in `/etc/nginx/conf.d`, etc.

There seems to be no obvious downside to removing this value.

### DIFF
--- a/scripts/site-types/cakephp3.sh
+++ b/scripts/site-types/cakephp3.sh
@@ -74,8 +74,6 @@ server {
 
     sendfile off;
 
-    client_max_body_size 100m;
-
     location ~ \.php$ {
         try_files \$uri =404;
         include fastcgi_params;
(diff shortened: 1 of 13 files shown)


svpernova09:
I'm good with this change, but would you mind removing this line from the rest of the site types as well please?

cbj4074:
Sure, no problem! I'll make that change shortly. Thanks for the speedy review @svpernova09 !

svpernova09:
Thanks!

PR #1333, opened 2019-12-07
Databases erase prevention setting

mariopro:
Updates Homestead.yaml.example to include as a default setting the "backup: true". This will prevent databases from being deleted on "vagrant destroy", especially for users that are not familiar with the destroy process and its consequences.
Homestead updates require the "vagrant destroy" to update the machine and many users aren't aware that they will end up losing the databases. For experienced users, it's just a matter of configuring the Homestead.yaml file and set it to false or remove the entry.

### DIFF
--- a/Homestead.yaml.example
+++ b/Homestead.yaml.example
@@ -20,6 +20,8 @@ sites:
 databases:
     - homestead
 
+backup: true
+
 features:
     - mariadb: false
     - ohmyzsh: false


svpernova09:
Thanks for taking the time to submit a PR. Going to hold off at this time because I don't think it's a good idea for Homestead to push files to the host OS by default.

mariopro:
I understand the point of view and perhaps another suggestion to prevent losses would be to update the documentation in [https://laravel.com/docs/6.x/homestead#database-backups](url) to include a warning message like the example shown below.

Homestead-Database-Backups warning

Just a suggestion. Thanks.

PR #1341, opened 2019-12-19
FIX: features' versions are not properly parsed to bash scripts

fidanf:
Hello

I've noticed this bug today. Here's the fix. The example below now work as intended :
```yml
features:
  - elasticsearch: true
    version: 7
  - gearman: true
```

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -243,7 +243,7 @@ class Homestead
         config.vm.provision "shell" do |s|
           s.name = "Installing " + feature_name
           s.path = feature_path
-          s.env = feature_variables
+          s.env = feature
         end
       end
     end


browner12:
To start, your `features` entry is incorrect.  If you look at the docs (https://laravel.com/docs/6.x/homestead#installing-optional-features)

```yaml
features:
    - elasticsearch:
        version: 7
    - gearman: true
```

The `feature_variables` is pulled out of each feature on line 224 above, so I believe the current code is correct.

fidanf:
Okay I've updated my codeblock, had forgotten the dashes.

Have you tried to run `vagrant up` in your example ? You cannot indent the `version` key after elasticsearch without throwing a YAML parsing error when you run either `vagrant validate` or `vagrant up`

Secondly, I can confirm the provision step fails once it runs the elasticsearch script, with the following output : 

> homestead: Reading package lists...
    homestead: E: The repository 'https://artifacts.elastic.co/packages/.x/apt stable Release' does not have a Release file.
    homestead: Reading package lists...
    homestead: Building dependency tree...

svpernova09:
Still unsure of what you're trying to accomplish here.

fidanf:
> 
> 
> Still unsure of what you're trying to accomplish here.

The below example YAML configuration does NOT work. 

```yml
features:
    - elasticsearch: true
       version: 7 
```

I suggest you try it yourself if you doubt it. This PR fixes it and I've explained how and why. That's it.

svpernova09:
> The below example YAML configuration does NOT work.

It's not *supposed* to.

Your block:

```
features:
  - elasticsearch: true
    version: 7
  - gearman: true
```

The block @browner12 gave you and pointed out yours was incorrect:

```
features:
    - elasticsearch:
        version: 7
    - gearman: true
```

Log showing ES installed as expected via the *correct syntax*: https://gist.github.com/svpernova09/2e8a726e5d4ddcd6bbbb42e1734a0e9c

Here's `apt show elasticsearch` which verifies version 7.x of Elastic Search was installed as expected:

```
root@homestead:/home/vagrant# apt show elasticsearch
Package: elasticsearch
Version: 7.5.1
Priority: optional
Section: web
Source: elasticsearch
Maintainer: Elasticsearch Team <info@elastic.co>
Installed-Size: 490 MB
Depends: bash (>= 4.1), lsb-base (>= 4), libc6, adduser, coreutils (>= 8.4)
Conflicts: elasticsearch-oss
Homepage: https://www.elastic.co/
License: Elastic-License
Download-Size: 290 MB
APT-Manual-Installed: yes
APT-Sources: https://artifacts.elastic.co/packages/7.x/apt stable/main amd64 Packages
Description: Distributed RESTful search engine built for the cloud
 Reference documentation can be found at
 https://www.elastic.co/guide/en/elasticsearch/reference/current/index.html
 and the 'Elasticsearch: The Definitive Guide' book can be found at
 https://www.elastic.co/guide/en/elasticsearch/guide/current/index.html

N: There are 13 additional records. Please use the '-a' switch to see them.
root@homestead:/home/vagrant#
```

Your indents on `version: 7` are incorrect. Fix this issue and then `vagrant destroy && vagrant up`

> I suggest you try it yourself if you doubt it. This PR fixes it and I've explained how and why. That's it.

I asked for clarification because I wasn't understanding why you were trying to fix a problem that wasn't a problem. I'm going to assume the best and encourage you to turn on whitespace indicators in your editor so you can more easily spot these issues where you're indention is causing the error you're trying to solve.

browner12:
Can confirm. Documented block works correctly.

browner12:
@fidanf copy the documented code *exactly*.  you still have an extra `true` in your example.

fidanf:
Fair enough, I understand now we both had something working but with a different syntax. I didnt catch the `true/false` value was meant to be replaced with other arguments. It seemed off to me in the first look.

However I would then signal this config can still be writen without using indentation (which is a bit misguiding imho), like so : 

> features:
    - mariadb: false
    - ohmyzsh: false
    - webdriver: false
    - elasticsearch: { version: 7 }

Nonetheless, my apologies for the confusion here. See ya. :)

svpernova09:
@fidanf Glad you got it sorted out!

The ES feature is a special case because the PPA supports multiple versions which not all features are able to support.

Cheers!

PR #1345, opened 2019-12-22
Add generated ca to local cert storage

kesselb:
Tried to connect from service a to service b on the same homestead box but the connection was refused due a self signed certificate. 

This patch will symlink the new ca to local certificate storage and execute the update script. Full disclosure: Tested it once :see_no_evil:

### DIFF
--- a/scripts/create-certificate.sh
+++ b/scripts/create-certificate.sh
@@ -87,6 +87,10 @@ then
         -key "$PATH_ROOT_KEY" \
         -x509 -new -extensions v3_ca -days 3650 -sha256 \
         -out "$PATH_ROOT_CRT" 2>/dev/null
+        
+        # Symlink ca to local certificate storage and run update command
+        ln --force --symbolic $PATH_ROOT_CRT /usr/local/share/ca-certificates/
+        update-ca-certificates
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for taking the time to create a PR.

Not opposed to this but would like another person to test drive it 🙈🙉🙊

kesselb:
> Not opposed to this but would like another person to test drive it

Sure ;) 

1. Box `mv /etc/nginx/ssl /etc/nginx/ssl.old` 
2. Host `vagrant provision` 

Will trigger the code.

browner12:
looks pretty standard to me, based on all the docs I found online.

is there a reason you chose to go with a soft link rather than copying it? All the snippets I saw online were just copying.  Not opposed to the soft link, just curious the reasoning behind it.

kesselb:
> just curious the reasoning behind it.

Not having the same data at two places. `update-ca-certificates` will also create a symlink to `/etc/ssl/certs`. So it's basically a chain of symlinks then.

I'm fine with symlink or a copy. As you prefer ;)

svpernova09:
Thanks!

PR #1355, opened 2020-01-19
Fix docblocks

iamgergo:


svpernova09:
Thanks for taking the time, going to pass on these changes for now.

PR #1379, opened 2020-03-03
[yii] Allow use index-test.php

qRoC:


scripts/site-types/yii.sh
```diff
@@ -71,7 +77,6 @@ block="server {
     location ~ \.php$ {
         fastcgi_split_path_info ^(.+\.php)(/.+)$;
         fastcgi_pass unix:/var/run/php/php$5-fpm.sock;
-        fastcgi_index index.php;
```
svpernova09: I believe we still need this because if `index-test.php` is missing the `location ~ \.php$` won't have an `fastcgi_index` to match.

qRoC: @svpernova09 fixed

svpernova09:
Thanks1

Issue #1392, opened 2020-04-02
Sort DB backups into folders


ethanclevenger91:
## Versions
  - Vagrant: 2.2.7
  - Provider: VBox 6.0.18
  - Homestead: release

It'd be great if DB backups were not just timestamped on the file name, but grouped into timestamped folders. That way it'd be easy to re-import an entire directory of files.

Couldn't find this script in the repo, but could work on a PR if I knew where to look

svpernova09:
The functionality is located in the Homestead Ruby class: https://github.com/laravel/homestead/blob/master/scripts/homestead.rb#L572-L585

As these are inline command scripts I don't know if I'd be in favor of any added complexity.

PR #1393, opened 2020-04-02
Add timestamp directory to mysql backups.

ethanclevenger91:
Addresses #1392 

I read concerns about increased complexity but I'm not sure what qualifies as "complex"


svpernova09:
Database Backup users should chime in, I don't use this feature so I wouldn't be affected.

ethanclevenger91:
At this time, Homestead doesn't offer any utility for importing databases, just exporting, so it definitely comes down to per-user use.

svpernova09:
Leaving open for a few more days, then will merge.

svpernova09:
Thanks!

Issue #1401, opened 2020-04-16
Feature Request - after script for vagrant up


skylerkatz:
Hello!

For a recent project I needed to run some shell scripts each time I started the vagrant machine.

This ended up being pretty easy to implement, but there was not a lot documented around the web about how to do this.

Would you be open to a PR that adds this functionality?

Obviously, you don't have to accept it if I do submit the pr, but wanted to check in before extracting the pr.

Thanks!

svpernova09:
Hi! :D

https://laravel.com/docs/7.x/homestead#extending-homestead

We have `after.sh` (Should be in VCS) and `user-customizations.sh` (Shouldn't be in VCS) to solve this. If what you build would add on these I'd be interested.

I'm always curious about how others implement Homestead.

skylerkatz:
I make pretty heavy usage of after.sh to install other os level packages, but there were a handful of things that I need to run every time I run `vagrant up`.

In my case, I need to make sure that Apache is running and that some os packages start.

I will go ahead and make a quick pr that adds a little more context.

svpernova09:
AH! I missed the `each time I started the vagrant machine`.

> make sure that Apache is running

If you're using site type `apache` it should be running, if not that's a bug and we should ensure that script enables the service.

> that some os packages start

This is something you should be using `service` for. Maybe even `supervisorctl`. 


I'll have a look at your PR later (Short on time atm).

skylerkatz:
No rush on the PR.

There may be a bug, but it likely has to do with SSL certificate generation and not with the site type.

In my use case I am running a handful of api services that all communicate with eachother. One of the services requires wildcard subdomains for per account access. For some reason, the only way to generate a proper wildcard SSL cert, I needed to set the site up twice in homestead.yaml. Once as apache and then again as laravel.
```yaml
-
    map: "my-app.test"
    to: /home/vagrant/repositories/my-app/public
    type: "apache"
    php: "7.3"

    # Used to auto generate the proper wildcard SSL Cert for Sermon Cloud
-
    map: "*.my-appo.test"
    to: /home/vagrant/repositories/my-app
    type: "laravel"
    php: "7.3"
```
When vagrant starts, it sees the laravel site and turns on ngnix. I unfortunately cannot use nginx (even though most of the projects on the box are laravel) because our main app is a legacy project that requires apache.

So I generate both sites and then in the `after.sh` script I make some changes to the apache config to point to the wildcard cert.

[...]

PR #1402, opened 2020-04-16
Feature - Allow after `vagrant up` scripts

skylerkatz:
After `vagrant up` is run, trigger the after-up.sh script to execute on the homestead box.

This is related to #1401.

There are times that I have needed to run scripts every time that I start homestead. 
For my exact use case:

- I re-run npm build scripts
- clear out some logs
- start an os level package
- ensure that apache is running (instead of nginx)

By adding this functionality to an `after-up.sh` script and modifying the `Vagrantfile` I was able to accomplish this pretty easily. I thought maybe it would be useful to others.

I can also see it being a niche feature though, so if it was treated like `user-customizations.sh` or closed altogether I would understand.

Thanks for all of your hard work on this project!


svpernova09:
Closing for now, as we're solving apache's issues #1401 can revist.

PR #1406, opened 2020-04-18
Bugfix: Database backups assumes existing `databases` setting.

juangiordana:
Database backups assumes `settings['databases']` to be non empty.

An error is triggered when `settings['databases']` is empty and `backup`
setting is enabled.


svpernova09:
Thanks for taking the time to do this, I think I'm ok with the current behavior failing hard. I like the all or nothing aspect this forces on the database backup settings.

Issue #1404, opened 2020-04-17
Add entry to XDebug config to work with certain IDEs 


michapietsch:
Hey,

to make XDebug work with Visual Studio Code I always need to apped `xdebug.remote_autostart = 1` to `xdebug.ini`. I do so in my `after.sh`.

Maybe others can confirm this need? Also, can anybody say if there is a potential negative effect on other IDE's settings? If not, I suggest to enhance the default config accordingly. 

My best regards, Micha

svpernova09:
If it's not going to adversely affect anything I'm good w/ making this change. Consider it on the TODO list.

michapietsch:
@svpernova09 Thanks!

I never use a browser extension or bookmarklet like mentioned in the docs. As I mention in the PR, it's as easy as set a breakpoint, start a debug session, and load the page (or run the test).

deviantintegral:
The one advantage of a cookie or other parameter is that you can more easily debug from only a single browser or session, without having to mute and unmute all your breakpoints. Perhaps an advanced use case, but it’s my guess why PHPStorm defaults to it in their docs: https://www.jetbrains.com/help/phpstorm/zero-configuration-debugging.html

You shouldn’t need auto start set for debugging to work at all with VS Code, since the ide shouldn’t care how the session was triggered. Can you confirm it doesn’t work when using a session parameter?

michapietsch:
@deviantintegral 

> having to mute and unmute all your breakpoints

I don't mute my breakpoints. I just stop my VS Code debug session. What exactly do you mean?

> Can you confirm it doesn’t work when using a session parameter?

Previously I used PHPStorm and it worked out of the box. However with VS Code it's my experience that I always need to set this parameter. Nothing else but the scaffolded VS Code debug config and a path mapping.

@svpernova09 

> If it's not going to adversely affect anything

I got this answer from Derick Rethans, creator of Xdebug:
"Not as long as that VM has no access from anything but the host OS. (i.e, no access beyond the subnet that the host and guest are on)"
(https://twitter.com/derickr/status/1251992263049252872)

So I understand it should be disabled during sharing.

svpernova09:
Hrm, that tweet concerns me, I replied.

svpernova09:
[...]

PR #1407, opened 2020-04-19
Set xdebug.remote_autostart = 1

michapietsch:
Implements the suggestion in #1404 to allow IDEs like Visual Studio Code to establish a debug session right away.

Only VS Code config left to do is a path mapping. Otherwise set a breakpoint, and start a debug session, load the page, and the breakpoint should immediately be hit.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -452,6 +452,13 @@ class Homestead
       end
     end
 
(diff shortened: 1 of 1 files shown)


svpernova09:
Closing as described in https://github.com/laravel/homestead/issues/1404#issuecomment-617454481

Thanks for taking the time to open the PR and engaged discussion.

PR #1410, opened 2020-04-22
Add wildcard & use_wildcard SSL Site options

svpernova09:
## TODO:

- [x] Apache Support
- [x] nginx Support

## Documenation:
Create an use wildcard for a site:
```
    - map: foo.domain.test
      to: /home/vagrant/domain
      wildcard: "yes"
      use_wildcard: "yes"
```

Create but do *not* use wildcard for a site:
```
    - map: foo.domain.test
      to: /home/vagrant/domain
      wildcard: "yes"
      use_wildcard: "no"
```

### DIFF
--- a/scripts/create-certificate.sh
+++ b/scripts/create-certificate.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 mkdir /etc/nginx/ssl 2>/dev/null
-
+set -f
 PATH_SSL="/etc/nginx/ssl"
 
 # Path to the custom Homestead $(hostname) Root CA certificate.
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -274,6 +274,14 @@ class Homestead
           s.args = [site['map']]
         end
 
+        if site['wildcard'] == 'yes'
+          config.vm.provision 'shell' do |s|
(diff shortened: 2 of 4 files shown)


svpernova09:
Currently does *not* support Apache

browner12:
What's the scenario when you would want to create the wildcard cert, but not use it?

svpernova09:
> What's the scenario when you would want to create the wildcard cert, but not use it?
The scenario that would 100% happen if I hadn't built that already :D

browner12:
hah, fair enough.

would it make more sense to default it to use the cert, and only make people add another setting if they want to override it off?

svpernova09:
> would it make more sense to default it to use the cert, and only make people add another setting if they want to override it off?

I didn't default to the wildcard because I'm not sure how many users really want wildcards. I've run into needing them a few time myself and it's a bit of a hack. Felt better to formalize the support and allow users to opt in.

deviantintegral:
Have you seen https://github.com/FiloSottile/mkcert ? We're using it on a project now and it is working well.

svpernova09:
I believe this is ready to go. Added Apache support, also squashed the bug where we were not properly disabling the nginx service when installing Apache.

Also refactored to `scripts/restart-webserver.sh` to dynamically figure out *which* web sever we should start.

I'm going to leave this in master for a few days and tag a release likely next week.

PR #1440, opened 2020-06-24
Fixed localized paths for serve alias commands

Braunson:
Fixed the localized paths for all the serve commands as they are located in the `site-types` folder now. Looks like this file was just overlooked when the change was made.


Braunson:
@svpernova09 ??

svpernova09:
> @svpernova09 ??

Whoops. That wasn't the right button...

svpernova09:
Thanks!

Issue #1447, opened 2020-07-02
Homestead.yaml doesn't follow .editorconfig


guoyunhe:
The Homestead.yaml created by by `homestead make` uses 4 spaces indention:
```
ip: 192.168.10.10
memory: 2048
cpus: 2
provider: virtualbox
authorize: ~/.ssh/id_rsa.pub
keys:
    - ~/.ssh/id_rsa
folders:
    -
        map: /home/guo/git/laravel-react-demo
        to: /home/vagrant/code
sites:
    -
        map: homestead.test
        to: /home/vagrant/code/public
databases:
    - homestead
features:
    -
        mariadb: false
    -
        ohmyzsh: false
    -
        webdriver: false
name: laravel-react-demo
hostname: laravel-react-demo

```
But in .editorconfig shipped by `laravel new` command is to use 2 spaces for yaml:
```
root = true

[*]
charset = utf-8
end_of_line = lf
insert_final_newline = true
indent_style = space
indent_size = 4
trim_trailing_whitespace = true

[*.md]
trim_trailing_whitespace = false

[*.{yml,yaml}]
indent_size = 2
```

svpernova09:
> But in .editorconfig shipped by laravel new command is to use 2 spaces for yaml:

@driesvints something awful has happened to `.editorconfig` ^


All jokes aside, I don't currently intend to change the indention of `Homestead.yaml`.

svpernova09:
If you want to use 2 space indention in your `Homestead.yaml` you 100% can and there's nothing there that will impact *how* Homestead works.

driesvints:
Well, 2 spaces has been the default for quite some years already. All yaml files in Laravel follow this (for example: https://github.com/laravel/laravel/blob/master/.styleci.yml). We're probably not changing this any time soon, sorry.

driesvints:
I also don't see why this should impact Homestead. Both can have their own preference.

svpernova09:
@driesvints I'm just poking fun :D

PSR-2 has ruined me, I can hardly stand the 2 space indents in Ruby :X

driesvints:
@svpernova09 hah yeah I know. Was more in reply to @guoyunhe.

PR #1456, opened 2020-07-20
Align editor config and yaml files formatting

bbsnly:


svpernova09:
Thanks for taking the time but we won't be making this change. #1447 has more info if you're curious.

PR #1468, opened 2020-07-31
Features: script improvements proposal.

juangiordana:
Features: script improvements proposal.

* Begin scripts with`set -e` to "Exit immediately if a pipeline, a list, or a compound command, exits with a non-zero status.". Commands with successful non-zero exit statuses can be handled separately.
* Remove "Check if installed" comments since code is self explanatory. 
* Mark scripts as done when all commands where executed successfully.

### DIFF
--- a/scripts/features/python.sh
+++ b/scripts/features/python.sh
@@ -1,19 +1,17 @@
 #!/usr/bin/env bash
 
-# Check If python Has Been Installed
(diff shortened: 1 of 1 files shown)


svpernova09:
Going to pass on this at the moment. Thanks for taking the time.

PR #1469, opened 2020-08-03
Added "https" type to force https when needed

mnaderian:
In some usecases it is necessary to force a project to open in **https** rather than **http**. (like **phpMyAdmin**).   

So with this type we can force a site to redirect to https (port 443) mode when opened with http (port 80).

Annotation 2020-08-03 173402

Example:

```yaml
sites:
    - map: homestead.test
      to: /home/vagrant/homestead/public
      type: "https"
```

### DIFF
--- a/scripts/site-types/https.sh
+++ b/scripts/site-types/https.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+
+declare -A params=$6       # Create an associative array
+declare -A headers=${9}    # Create an associative array
+declare -A rewrites=${10}  # Create an associative array
+paramsTXT=""
+if [ -n "$6" ]; then
+   for element in "${!params[@]}"
+   do
+      paramsTXT="${paramsTXT}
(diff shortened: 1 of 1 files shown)


svpernova09:
I see what you're trying to do and I don't think a new `https` site type is the solution. If your application requires SSL then your application configuration should account for that and you should remove the port 80 configuration from your virtual host. It's not currently possible to do that via Homestead config. And I'm not sure it should.

Apache users can configure this in the application via `.htaccess` or in their Apache virtual host.

I could potentially see this being implemented as a site option that forces SSL by removing the port 80 configuration from the virtual host, but that may be more confusing than leaving the current default.

mnaderian:
@svpernova09 you're right Joe. One might wanna use a different type and in that case he won't be able to use this way. So it has to be implemented as a site option. I'll try to find a way to implement it that way if possible.

Take care,
and take care of that cute kitty 😉 😜

PR #1472, opened 2020-08-05
Feature: Add EventStore script

joemugen:
This script will install and configure EventStore using official documentation https://eventstore.com/docs/getting-started/index.html
There is some options you can add on the Homestead.yaml file

```yaml
#...
features:
    - mariadb: false
    - ohmyzsh: false
    - webdriver: false
    - eventstore:
        version: 5.0.8-1
        run_projections: All
        external_ip: 0.0.0.0
        external_tcp_port: 1113
        external_http_port: 2113
        admin_on_ext: true

ports: # Needed if you want to expose the service on the external IP
     - send: 1113
       to: 1113
     - send: 2113
       to: 2113
#...
```

Then you can access it from http://{VM IP or FQDN}:2113 with the default username and password (`admin` / `changeit`)

### DIFF
--- a/scripts/features/eventstore.sh
+++ b/scripts/features/eventstore.sh
@@ -21,13 +21,13 @@ else
 fi
 
 # Install repository
-curl -s https://packagecloud.io/install/repositories/EventStore/EventStore-OSS/script.deb.sh | sudo bash
+curl -s https://packagecloud.io/install/repositories/EventStore/EventStore-OSS/script.deb.sh | bash
 
 # Install EventStore package
-sudo apt-get install -y eventstore-oss"$installVersion"
+apt-get install -y eventstore-oss"$installVersion"
 
 # Update configuration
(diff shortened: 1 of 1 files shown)


scripts/features/eventstore.sh
```diff
@@ -0,0 +1,45 @@
...
+    installVersion="" # by not specifying we'll install latest
+else
+    installVersion="=${version}"
+fi
+
+# Install repository
+curl -s https://packagecloud.io/install/repositories/EventStore/EventStore-OSS/script.deb.sh | sudo bash
+
+# Install EventStore package
+sudo apt-get install -y eventstore-oss"$installVersion"
```
svpernova09, line 27: We can remove `sudo` here since features will always be executed as `root`

joemugen: Of course :)

PR #1488, opened 2020-09-11
added wsl_username and wsl_user_group env var

litan1106:
added wsl_username and wsl_user_group env var so we can just change the env vars


svpernova09 (approved):
This is exactly what I was hoping someone would PR. Thanks!

PR #1490, opened 2020-09-11
RVM Support

martianatwork:
Details:
Installs stable version of RVM with Latest Ruby & Bundler.

Usage:

    features:
        - rvm: true


svpernova09 (approved):
Thanks!

PR #1504, opened 2020-09-28
Installed and disable mailhog by default

litan1106:
Installed and disable mailhog by default. (we should add `Enable / Disable Services` support to wsl2)


svpernova09:
Thanks!

PR #1505, opened 2020-09-28
Updated wslapplyfeatures to use feature variables.

litan1106:
example: (this will pass the server_id as variables/args)
features:
    - blackfire:
        server_id: "server_id"
        server_token: "server_value"
        client_id: "client_id"
        client_token: "client_value"
    - cassandra: true


svpernova09 (approved):
Thanks!

PR #1506, opened 2020-09-29
Added function to extend scripts and site types

Saij:
I added the option to overwrite scripts (and especially add site types) when using Homestead on a per project basis.

Because I want to use Homestead on a ProcessWire project and I would like to use it with composer not globally but only on this project I searched for a way to include a new site type "ProcessWire" but couldn't find any other than forking the git and add it then. But I also don't want to maintaine a full fork and keep it up to date I had the idea to enable a user script dir.

The directory will be in a folder called "homestead" which lies beneath the homestead.yaml. So for example to add a new site type you just create a file "homestead/scripts/site-types/processwire.sh"

You could also use this feature to overwrite some settings by other scripts/site-types

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -33,7 +33,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         abort "Homestead settings file not found in #{confDir}"
     end
 
-    Homestead.configure(config, settings)
+    Homestead.configure(config, settings, "")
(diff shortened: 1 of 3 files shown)


svpernova09:
Going to pass on these changes. Feel free to PR your custom site type. Just ensure it behaves like the others and accounts for all the inputs.

PR #1520, opened 2020-10-17
Fix Magento site type "upstream sent too big header while reading res…

jeshtan:
…ponse header from upstream" error

Raise `fastcgi_buffer_size` as official Magento Nginx sample recent change (https://github.com/magento/magento2/blame/2.4-develop/nginx.conf.sample#L185) to allow PHP process larger request.


svpernova09 (approved):
Thanks

PR #1527, opened 2020-10-27
Add MySQL as a default feature

Vercoutere:
Since commit 7ef844020750deb012d3596b9b0a9aa97b896977 the MySQL databases are no longer being created by default. This is an undocumented change, and can be quite confusing, considering MySQL has always been enabled by default in Laravel Homestead.

I have enabled the MySQL feature by default in the Homestead.yaml/json configuration files. This still allows for MySQL to be disabled if desired, without being confusing and forgoing sensible defaults.

### DIFF
--- a/resources/Homestead.yaml
+++ b/resources/Homestead.yaml
@@ -21,6 +21,7 @@ databases:
     - homestead
 
 features:
+    - mysql: false
     - mariadb: false
     - ohmyzsh: false
     - webdriver: false
--- a/resources/Homestead.json
+++ b/resources/Homestead.json
@@ -24,11 +24,14 @@
     ],
     "features": [
         {
+            "mysql": true
+        },
+        {
             "mariadb": false
-        }, 
+        },
         {
             "ohmyzsh": false
-        }, 
+        },
         {
             "webdriver": false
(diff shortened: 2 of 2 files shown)


svpernova09:
Thanks for beating me to this.

Vercoutere:
> Thanks for beating me to this.

My pleasure.

resources/Homestead.yaml
```diff
@@ -21,6 +21,7 @@ databases:
     - homestead
 
 features:
+    - mysql: false
```
seanbright, line 24: Should this have been `true`?

svpernova09: The value is only evaluated if you've enabled the backup option. in which case you should enable MySQL if you're using it.

seanbright: I ask because in the same PR the mysql key's value was changed to true in `Homestead.json` - if the two files are not supposed to be the same, then I apologize for the noise.

svpernova09: Not a *huge* deal. IMHO it should default to false, but I can see the reasoning for it to default to true as well.

PR #1537, opened 2020-11-02
[20.04] Add support for php 8

chris-doehring:
This PR add's support for php 8 on `20.04`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,7 @@ before_script:
 
 script:
   - vendor/bin/phpunit
+
+matrix:
+  allow_failures:
+    - php: 8.0snapshot
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: php
 
 php:
   - 7.4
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,15 +2,11 @@ language: php
...
-  - nightly
+  - 8.0snapshot
 
 before_script:
   - phpenv config-rm xdebug.ini || true
   - composer update --no-interaction
 
 script:
   - vendor/bin/phpunit
-
```
svpernova09, line 13: We should allow `8.0snapshot` to fail since 8.0 has not been released yet.

svpernova09 (changes requested):

.travis.yml
chris-doehring: changed :)

svpernova09 (approved):
Thanks!

chris-doehring:
Since php 8 will be released on Thursday, could we get a release with this change? :) @svpernova09

PR #1544, opened 2020-11-13
Create Mysql8 Databases if Mysql8 feature enabled

evanrthompson:
Generating databases on up/provision when using mysql 8 is broken. This fixes that :)


svpernova09 (approved):
Thanks!

PR #1545, opened 2020-11-18
add PostgreSQL as a default feature

iDuriel:
Similar to the PR for MySQL here: https://github.com/laravel/homestead/pull/1527 , this change adds the option to enable PostgreSQL in the features array in the Homestead.yaml file.


svpernova09 (approved):
Thanks

PR #1548, opened 2020-11-22
Move Composer to version 2

jhoskins98:
Move composer to version 2 channel


svpernova09:
Going to hold off for now, feel free to throw this in your `after.sh`. My plan is to do a major base box version bump for defaulting to PHP 8 which will also be when I default to Composer 2.

PR #1553, opened 2020-11-25
Add function to switch to PHP 8.0

DieterHolvoet:
I added the function to switch to PHP 8.0, just like there are the functions for PHP 7.x.

One thing I noticed: the `/usr/bin/phpize` and `/usr/bin/php-config` version-specific binaries don't seem to be installed anymore. Is this normal? If so, we should probably update the aliases.

### DIFF
--- a/resources/aliases
+++ b/resources/aliases
@@ -80,6 +80,12 @@ function php74() {
     sudo update-alternatives --set phpize /usr/bin/phpize7.4
 }
 
+function php80() {
+    sudo update-alternatives --set php /usr/bin/php8.0
+    sudo update-alternatives --set php-config /usr/bin/php-config8.0
+    sudo update-alternatives --set phpize /usr/bin/phpize8.0
+}
+
 function serve-apache() {
     if [[ "$1" && "$2" ]]
     then
--- a/resources/localized/aliases
+++ b/resources/localized/aliases
@@ -63,6 +63,12 @@ function php74() {
     sudo update-alternatives --set phpize /usr/bin/phpize7.4
 }
 
+function php80() {
+    sudo update-alternatives --set php /usr/bin/php8.0
+    sudo update-alternatives --set php-config /usr/bin/php-config8.0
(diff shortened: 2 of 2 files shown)


svpernova09 (approved):
Thanks!

DieterHolvoet:
I'll add it to the aliases file in the `localized` folder as well. What's the difference between the two files?

svpernova09:
> 
> 
> I'll add it to the aliases file in the `localized` folder as well. What's the difference between the two files?

The localized versions get copied when someone is using the per-project installation method as opposed to cloning this repository. https://laravel.com/docs/8.x/homestead#per-project-installation

svpernova09:
> One thing I noticed: the `/usr/bin/phpize` and `/usr/bin/php-config` version-specific binaries don't seem to be installed anymore. Is this normal? If so, we should probably update the aliases.

Hrm, This I'm not sure of. It could be because they weren't installed when 8 RC (or whatever early release is in Homestead atm)

DieterHolvoet:
> > One thing I noticed: the `/usr/bin/phpize` and `/usr/bin/php-config` version-specific binaries don't seem to be installed anymore. Is this normal? If so, we should probably update the aliases.
> 
> Hrm, This I'm not sure of. It could be because they weren't installed when 8 RC (or whatever early release is in Homestead atm)

They aren't being installed for any version of PHP.

svpernova09:
> 
> 
> > > One thing I noticed: the `/usr/bin/phpize` and `/usr/bin/php-config` version-specific binaries don't seem to be installed anymore. Is this normal? If so, we should probably update the aliases.
> > 
> > 
> > Hrm, This I'm not sure of. It could be because they weren't installed when 8 RC (or whatever early release is in Homestead atm)
> 
> They aren't being installed for any version of PHP.

Yikes, I would skip it for this PR sounds like I need to do some digging.

DieterHolvoet:
Okay, I'll create an issue.

allok (approved):

PR #1555, opened 2020-11-25
Allow installing Elasticsearch 5

DieterHolvoet:


browner12:
Can you provide an explanation? Does v5 not work when you pass it as an input?

svpernova09:
> 
> 
> Can you provide an explanation? Does v5 not work when you pass it as an input?

I vaguely remember dropping support for version 5 when I did updates for 7... As long as the files are in the PPA it's fine(probably) to support 5.

browner12:
That definitely rings a bell. All I could dig up, though, was you asking me about it when I added ES6 support.

https://github.com/laravel/homestead/pull/767

DieterHolvoet:
Without this, I get the following error:

Screenshot

svpernova09:
> 
> 
> Without this, I get the following error:
> 
> Screenshot

Of the top of my head: Run `sudo apt-get update`, then `apt-cache search elasticsearch`, anything returned in the list should be installable as long as the full version matches. IE: if you want version 5, you need to pick the full 5.x.x version number.

It's possible this is broken and needs to be fixed.

DieterHolvoet:
I looked it up and this is how you list all versions:
```bash
sudo apt-get update
apt-cache madison elasticsearch
```
I tested it again with the following config in `Homestead.yml`
```yml
elasticsearch: 
  version: 5.6.16
```
and it seems to work now. Trying to find out what I did wrong, I read the Homestead docs about Elasticsearch and I found this piece:
> You may specify a supported version of Elasticsearch, which may be a **major** version or an exact version number (major.minor.patch).

There's the problem: just entering a major version is not enough because you will get this:
```
sudo apt-get -y install elasticsearch=5
Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Version '5' for 'elasticsearch' was not found
```

DieterHolvoet:
If one of you can confirm this, I'll create a PR on the laravel/docs repo.

browner12:
Is that issue specific to v5?  Can you specify just the major version for 6 or 7?

DieterHolvoet:
No, you can't.

browner12:
Confirmed, major versions don't work.

It looks like our example is correct, but our explanation is not.  Yah, I think a docs PR is a good idea.

svpernova09:
+1 for a docs update instead.

DieterHolvoet:
Docs are updated, I'll close this PR. Thanks for the review!

PR #1557, opened 2020-11-26
Updating magento template to support params

minioak:
MAGE_RUN_CODE and MAGE_RUN_TYPE fastcgi params are required for multi domain magento installations. This adds support for them


svpernova09 (approved):
Thanks!

PR #1562, opened 2020-12-02
Update MongoDB PHP driver to PHP8 compatible v1.9.0 

litan1106:
Update MongoDB PHP driver to PHP8 compatible v1.9.0


svpernova09:
Thanks!

PR #1568, opened 2020-12-15
Add php version to cron schedule

skylerkatz:
Hello Joe!

I came across an issue where I have multiple sites running on the same box with different php versions. The cron schedule was always using the current php version of the CLI, which causes issues when a site has a different version configured.

This update appends the php version to the cron command if it is set.

### DIFF
--- a/scripts/cron-schedule.sh
+++ b/scripts/cron-schedule.sh
@@ -6,7 +6,8 @@ fi
 
 SITE_DOMAIN=$1
 SITE_PUBLIC_DIRECTORY=$2
+SITE_PHP_VERSION=$3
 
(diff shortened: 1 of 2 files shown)


browner12:
Probably a good idea. Looks like Forge is going this direction, too.

A little off topic, but out of curiosity, is your Cron just constantly running on your Homestead box?

skylerkatz:
> Probably a good idea. Looks like Forge is going this direction, too.
> 
> A little off topic, but out of curiosity, is your Cron just constantly running on your Homestead box?

For the most part I have cron running all the time. At work we have a number of microservices that in development all live on one homestead virtual machine. Most of the microservices are queue and scheduled job based, so we run the cron schedule and work commands .

svpernova09 (approved):
Thanks!

Issue #1569, opened 2020-12-15
VMWare not recognizing cpus or memory in Homestead.yaml


yazeed:
So, I just switched from vagrant virtualbox to vmware, and for some reason everything is working as intended except when I do vagrant up, the vagrant box is up with 1 processor and 2GB RAM even though I have it defined otherwise in my Homestead.yaml.

## Expected behavior
To start vagrant box with 4 cpus and 12GB ram.

## Actual behavior
vagrant box started with 1 cpu and 2GB ram.

## Versions
  - Vagrant: 2.2.14.
  - Provider: VMWare Fusion 12.1.0.
  - Homestead: Tag:v11.4.0
  - Box: laravel/homestead (vmware_desktop, 10.1.1)

**A lot of issues can be resolved by simply updating vagrant, provider or homestead.**
>**Note:** Virtualbox users, please upgrade to `6.x`. You'll need to download a newer version of [Virtualbox](https://www.virtualbox.org/wiki/Downloads).

## Host operating system
MacOS 10.15.7 (19H114)

## Homestead.yaml
```yaml
---
ip: "192.168.10.10"
memory: 12288
cpus: 4
provider: vmware_desktop

authorize: ~/.ssh/id_rsa.pub

keys:
    - ~/.ssh/id_rsa

folders:
    - map: ~/Sites
      to: /home/vagrant/sites
.
.
.
```

svpernova09:
Open this file and find this line: https://github.com/laravel/homestead/blob/master/scripts/homestead.rb#L59

`['vmware_fusion', 'vmware_workstation'].each do |vmware|`

and change the line to:

`['vmware_fusion', 'vmware_workstation', 'vmware_desktop '].each do |vmware|`

Then `vagrant destroy && vagrant up`

yazeed:
Thanks @svpernova09

svpernova09:
> 
> 
> Thanks @svpernova09, added PR for it.

Thanks! I'll merge once Travis-CI catches up!

PR #1570, opened 2020-12-15
missing vmware_desktop directive in homestead.rb

yazeed:
In reference to #1569, this addresses the missing **vmware_desktop** directive in homestead.rb.


svpernova09 (approved):
Thanks!

PR #1600, opened 2021-01-27
Increase maximum size of the client request body

sdespont:
https://www.cyberciti.biz/faq/linux-unix-bsd-nginx-413-request-entity-too-large/


svpernova09:
Thanks!

PR #1620, opened 2021-02-20
Update readme.md

LewisCraik:
Corrected version table to match main.


svpernova09:
Thanks, but our release process is to merge into `main` then into `release` when a new release is tagged.

PR #1630, opened 2021-03-21
Removed trailing slash from apache config

firojghimire:
The trailing slash on `fcgi://localhost/` inside apache config file was causing problems while loading some static assets. This PR fixes that problem.


svpernova09:
Thanks

PR #1631, opened 2021-03-22
Apache corrupts files on mounted folder

skyosev:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=900821

I encountered this issue under hyper-v. The proposed fix disables mmap, which seems to be broken on cifs.


svpernova09:
Thanks!

PR #1633, opened 2021-03-23
Fix Minio Buckets Being Setup Incorrectly

vpillinger:
Creating new buckets in Minio causes the folders to be created under `root`.
While the fact that the Minio CLI behaves this way is extremely questionable, this is a one-line fix for the permission issue it causes.


vpillinger:
To clarify, buckets created under the `root` user will error out when attempting to use them because the minio user does not have permissions to edit the folder.

svpernova09 (approved):
Thanks!

Issue #1590, opened 2021-01-07
Upgrade TimescaleDB to 2.0


yazeed:
Hi,

I'd like to submit a PR to upgrade the default Timescale shipped with Homestead from v1.7.4 to v2.0 as it's nearing a final release, RCv4, but thought I'd ask here first to see if this would be a welcome change.

There are some great new enhancements and features bundled with v2.0 and it would be nice to have it by default on the box.

The way to install it, below, won't be changing with the upcoming full release.

```
# Add our PPA
sudo add-apt-repository ppa:timescale/timescaledb-ppa
sudo apt-get update

# Now install appropriate package for PG version
sudo apt install timescaledb-2-postgresql-12
```

Reference: https://docs.timescale.com/latest/getting-started/installation/ubuntu/installation-apt-ubuntu

svpernova09:
This *would* be a welcome change. I know 2.0 just went GA recently. As long as it's in the repos and available.

JonoB:
Note: Timescale only works on Postgres 11 or 12, and Homestead's default Postgres version is 13 at the moment.

yazeed:
> Note: Timescale only works on Postgres 11 or 12, and Homestead's default Postgres version is 13 at the moment.

Well, yes, it will be supported in [Feb 2021](https://github.com/timescale/timescaledb/issues/2779#issuecomment-769783709).

PR #1635, opened 2021-04-02
Update timescaledb2 to support postgres 13

yazeed:
This PR is a followup to #1590 to install the latest timescaledb2 version which provides support for Postgres 13, which is now default.


svpernova09 (approved):

svpernova09:
Thanks!

PR #1636, opened 2021-04-04
[Meilisearch] Add executable path and default homestead IP on ExecStart

andreia:
Update Meilisearch `ExecStart` adding the executable path and the default homestead IP.


svpernova09 (approved):
Thanks

PR #1643, opened 2021-04-25
Meilisearch accept all incoming ip addresses

RemCom:
Listen to all incoming connections instead of specific ip


svpernova09 (approved):

svpernova09:
Thanks!

jeffochoa:
Can this IP address `192.168.10.10` be picked up from the Homestead.yaml instead?

PR #1644, opened 2021-04-28
Add ERB parsing to YAML and JSON settings files

ziadoz:
This PR enables support for ERB template parsing of the `Homestead.yaml` and `Homestead.json` settings files before they are parsed into a hash object in the `Vagrantfile`.

I added ERB parsing because in my case, it enables me to pass through the `COMPOSER_AUTH` environment variable from my host `~/.zshrc` file to my Homestead VM, which I need to install private packages, by loading it in the variables block like this:

```
variables:
  - key: COMPOSER_AUTH
    value: '<%= ENV["COMPOSER_AUTH"] %>'
```

I'm not a Ruby person, so I don't know if this is the "done thing", but I think I've seen this behaviour in Rails. It would allow a bit more flexibility in the configuration files, which would be useful (e.g. scanning a directory for projects to map).

Thanks!

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,6 +3,7 @@
 
 require 'json'
 require 'yaml'
+require 'erb'
 
 VAGRANTFILE_API_VERSION ||= "2"
 confDir = $confDir ||= File.expand_path(File.dirname(__FILE__))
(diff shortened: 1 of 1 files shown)


svpernova09:
I see what you're doing here, but I think I might pass on this change. I don't feel like giving PHP devs the ability to inject Ruby syntax into Yaml is a great idea.

IMHO your application shouldn't be sourcing things from your host's `~/.zshrc`. Any environment variables required to run your application should be bundled in its run time (Application config or even VirtualHost configuration)

Thanks for taking the time to open  PR, I'm glad you found a workaround for your use case.

svpernova09:
In your specific use case `COMPOSER_AUTH`, I would add that to your `after.sh` to set that variable in the `vagrant` user's profile (bash/zsh/etc) so it's embedded in the VM, not based on getting sourced from your host.

ziadoz:
Thanks for reviewing this PR so quickly. I knew this was a long shot, but at the very least I figured the code might help somebody else out in the future.

I actually ended up putting `COMPOSER_AUTH` in the `user-customizations.sh` script, which works nicely, as it isn't version controlled, and our team each have their own personal tokens.

Thanks again.

PR #1658, opened 2021-05-31
Add option to force SSL

DieterHolvoet:
I added the `force_ssl` option to force SSL for all sites, with the possibility to add overrides per site. The default value is `false` since I'm not sure whether this could break existing installations.

What it does is redirecting all insecure requests to their secure counterparts using NGINX. I considered adding the `Strict-Transport-Security` as well, but in the end I didn't because I'm not sure it has added value for a website on a local development machine.

I only changed the `laravel` site type because it's the only one I'm using and because I'm not sure if you guys want to add this feature. Please let me know what you think.

### DIFF
--- a/Homestead.yaml.example
+++ b/Homestead.yaml.example
@@ -3,6 +3,7 @@ ip: "192.168.10.10"
 memory: 2048
 cpus: 2
 provider: virtualbox
+force_ssl: true
 
 authorize: ~/.ssh/id_rsa.pub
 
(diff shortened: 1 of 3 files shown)


svpernova09:
Thanks for taking the time to open a PR. Going to hold off on this change for now. This sounds like it should be part of the application, not the environment. (that's easier to accomplish with Apache than Nginx w/ default Laravel shipping with `.htaccess`) SSL is a very confusing topic related to Homestead for many people and I'm not sure adding this change would help that.

PR #1661, opened 2021-06-12
Add ngrok feature

Olray:
"ngrok provides a real-time web UI where you can introspect all HTTP traffic running over your tunnels. Replay any request against your tunnel with one click." (Source: https://ngrok.com/) And the best part is that the basic services are even free.

Script /scripts/features/ngrok.sh installs a snap with ngrok and writes a basic configuration file. It also adds a cmd to ~/.bashrc

It expects the authtoken as a mandatory parameter and an optional region (e.g. "us" or "eu"). Example added to Homestead.yaml.

Usage: inside the Homestead VM write:

```
rock yourdomain.local
```
that's it. If you run into problems with the predefined rock() function you can still use the ngrok command directly.
There is a web console with information about your request at http://<vmip>:4040

Configure the feature in Homestead.yaml like this:
```
features:
    - ngrok:
        authtoken: <token>
        region: eu
```

Here's the ngrok server running:
ngrok running on laravel homestead vm

The web interface is most useful, allowing you to inspect the response of your web server in detail

Web interface with most useful http request information

_This is my very first contribution with Git and Github. Please bear with me while I get familiar with Git, thanks :)_


svpernova09:
Homestead already provides ngrok support via the `share` command, which is a bash alias. https://laravel.com/docs/8.x/homestead#sharing-your-environment

Olray:
Inventing the wheel twice sounds just like me :)
Thanks again for the info

Issue #1614, opened 2021-02-11
MySQL DB Provisioning Error in Provision CLI Output


RossIV:
## Versions
  - Vagrant: 2.2.14
  - Provider: Virtualbox 6.1.16
  - Homestead: 11.0.0 (latest from `release` branch)

## Host operating system
macOS 10.14.6

## Homestead.yaml
https://gist.github.com/RossIV/13909bf231541caabefa0ffed51685ea

## Vagrant destroy & up output
https://gist.github.com/RossIV/5ddf55180f73a5ec67aba8b082ad5164

## Expected behavior
Creation of databases listed in Homestead.yaml file in MySQL upon initial box provisioning, and output status of creation of each database listed to the console.

## Actual behavior
Databases were created, but the CLI output indicated otherwise.

[Specific section of `vagrant up` output](https://gist.github.com/RossIV/5ddf55180f73a5ec67aba8b082ad5164#file-out-txt-L101):
```
==> homestead: Running provisioner: Creating MySQL / MariaDB Database: homestead (shell)...
    homestead: Running: script: Creating MySQL / MariaDB Database: homestead
    homestead: We didn't find a PID for mariadb, skipping $DB creation
==> homestead: Running provisioner: Creating MySQL / MariaDB Database: test2314 (shell)...
    homestead: Running: script: Creating MySQL / MariaDB Database: test2314
    homestead: We didn't find a PID for mariadb, skipping $DB creation
==> homestead: Running provisioner: Creating MySQL / MariaDB Database: testagain (shell)...
    homestead: Running: script: Creating MySQL / MariaDB Database: testagain
    homestead: We didn't find a PID for mariadb, skipping $DB creation
```

Also of note, `mysql` shows as an invalid feature [in the same output](https://gist.github.com/RossIV/5ddf55180f73a5ec67aba8b082ad5164#file-out-txt-L63) - not sure if that's related but figured it's worth mentioning.
```
    homestead: Running: inline script
    homestead: Invalid feature: mysql
```

## Steps to reproduce
1. `cd ~/Homestead`
2. `git fetch`
3. `git pull origin release`
4. `bash init.sh` (answer `y` to prompts)
5. `vagrant box update`
6. Update `Homestead.yaml` to add additional entries to the list of databases
7. `vagrant up`
8. `vagrant ssh`
9. `mysql -u homestead -p`
10. `show databases;`

svpernova09:
From your `Homestead.yaml`, add the following:

```
services:
[...]

PR #1662, opened 2021-06-13
Update to separate the MariaDB / MySQL provisioning scripts

4sitemedia:
This pull request is related to issue #1614

This update separates the MariaDB / MySQL provisioning scripts to prevent the display of a failure message that is accurate but misleading / confusing. Assuming that only one of MariaDB or MySQL is enabled the following message currently will always be displayed because the other process will not exist:
```
homestead: We didn't find a PID for [mariadb | mysqld], skipping $DB creation
```

Updates include:
* separate the `create-mysql.sh` script into `create-mariadb.sh` for MariaDB and `create-mysql.sh` for MySQL
* update the `homestead.rb` script to execute the appropriate script
* update the skip messages to not escape the database name so that the database name is displayed instead of `skipping $DB creation`

For example, when enabling MariaDB with the following `Homestead.yaml` configuration:
```
databases:
    - homestead

features:
    - mysql: false
    - mariadb: true
```

the current output is:
```
==> homestead: Running provisioner: Creating MySQL / MariaDB Database: homestead (shell)...
    homestead: Running: script: Creating MySQL / MariaDB Database: homestead
    homestead: We didn't find a PID for mysqld, skipping $DB creation
```

With the proposed update the output is:
```
==> homestead: Running provisioner: Creating MariaDB Database: homestead (shell)...
    homestead: Running: script: Creating MariaDB Database: homestead
```

Similarly when enabling MySQL with the following `Homestead.yaml` configuration:
```
databases:
    - homestead

features:
    - mysql: true
    - mariadb: false
```

the current output is:
```
==> homestead: Running provisioner: Creating MySQL / MariaDB Database: homestead (shell)...
    homestead: Running: script: Creating MySQL / MariaDB Database: homestead
    homestead: We didn't find a PID for mariadb, skipping $DB creation
```

With the proposed update the output is:
```
==> homestead: Running provisioner: Creating MySQL Database: homestead (shell)...
    homestead: Running: script: Creating MySQL Database: homestead
```

### DIFF
--- a/scripts/create-mariadb.sh
+++ b/scripts/create-mariadb.sh
@@ -11,12 +11,12 @@ cp /root/.my.cnf /home/vagrant/.my.cnf
 
 DB=$1;
 
-mysql=$(pidof mysqld)
+mariadb=$(pidof mariadbd)
 
-if [ -z "$mysql" ]
+if [ -z "$mariadb" ]
 then
-    # Skip Creating MySQL database
-    echo "We didn't find a PID for mysqld, skipping $DB creation"
+    # Skip Creating MariaDB database
+    echo "We didn't find a PID for mariadbd, skipping $DB creation"
 else
     mysql -e "CREATE DATABASE IF NOT EXISTS \`$DB\` DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci";
 fi
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -545,9 +545,17 @@ class Homestead
       end
 
       settings['databases'].each do |db|
-        if (enabled_databases.include? 'mysql') || (enabled_databases.include? 'mysql8') || (enabled_databases.include? 'mariadb')
+        if enabled_databases.include? 'mariadb'
           config.vm.provision 'shell' do |s|
-            s.name = 'Creating MySQL / MariaDB Database: ' + db
+            s.name = 'Creating MariaDB Database: ' + db
+            s.path = script_dir + '/create-mariadb.sh'
+            s.args = [db]
+          end
+        end
+
+        if (enabled_databases.include? 'mysql') || (enabled_databases.include? 'mysql8')
+          config.vm.provision 'shell' do |s|
+            s.name = 'Creating MySQL Database: ' + db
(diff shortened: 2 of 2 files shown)


svpernova09:
Looking back at https://github.com/laravel/homestead/issues/1614, my assessment was the OP missed the changes in the upgrade notes from version to version.

I freely admit the MySQL/MariaDB implementation is not ideal, but it's the best and most reproducible I've been able to find. This PR seems to be trading BASH script complexity for more Ruby complexity, in this tradeoff I'd *much* rather have BASH complexity as PHP developers will be much less likely to have Ruby experience than BASH. Thanks for taking the time, but I'd rather stick with the current implementation for now.

4sitemedia:
Thank you for the feedback! If the preference is to confine complexity to the bash scripts, would something along these line be acceptable? (not fully tested)

`homestead.rb`:
```
        if (enabled_databases.include? 'mysql') || (enabled_databases.include? 'mysql8') || (enabled_databases.include? 'mariadb')
          config.vm.provision 'shell' do |s|
            s.name = 'Creating MySQL / MariaDB Database: ' + db
            s.path = script_dir + '/create-mysql.sh'
            s.args = [db, enabled_databases.join('|')]
          end
        end
```

`create-mysql.sh`:
```
DB=$1;
ENABLED=$2;

if [[ $ENABLED =~ ^.*mysql.*$ ]]
then
    mysql=$(pidof mysqld)

    if [ -z "$mysql" ]
    then
        # Skip Creating MySQL database
        echo "We didn't find a PID for mysqld, skipping $DB creation"
    else
        mysql -e "CREATE DATABASE IF NOT EXISTS \`$DB\` DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci";
    fi
fi

if [[ $ENABLED =~ ^.*mariadb.*$ ]]
then
    mariadb=$(pidof mariadbd)

    if [ -z "$mariadb" ]
    then
        # Skip Creating MariaDB database
        echo "We didn't find a PID for mariadb, skipping $DB creation"
    else
        mysql -e "CREATE DATABASE IF NOT EXISTS \`$DB\` DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci";
    fi
fi
```

PR #1667, opened 2021-06-28
add support for vagrant-goodhosts

thedavidthomas:
Adds support for Vagrant goodhosts plugin https://github.com/goodhosts/vagrant


svpernova09:
From the docs:

> This vagrant plugin adds host file entries to the host pointing to the guest VM, using the GoodHosts cli tool. This plugin is based on vagrant-hostsupdater and aims to be compatible with the same config parameters.

We already support `vagrant-hostsupdater` and if the same syntax works there I'd prefer to not support another plugin.

thedavidthomas:
[vagrant-hostsupdater](https://github.com/agiledivider/vagrant-hostsupdater) is unfortunately no longer maintained

svpernova09:
Thanks!

PR #1672, opened 2021-07-08
Update after.sh

stebogit:
Updates Node version

### DIFF
--- a/resources/after.sh
+++ b/resources/after.sh
@@ -17,14 +17,7 @@
 #sudo rm -rf /usr/lib/node_modules/npm/lib
 #sudo rm -rf //etc/apt/sources.list.d/nodesource.list
 
-# Install Node.js v12.x
-#curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
-#sudo apt-get install -y nodejs
-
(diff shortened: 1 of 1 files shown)


svpernova09 (changes requested):
Lets change this block to the following to give more clarity and context:

```
# If you're not quite ready for the latest Node.js version,
# uncomment these lines to roll back to a previous version

# Remove current Node.js version:
#sudo apt-get -y purge nodejs
#sudo rm -rf /usr/lib/node_modules/npm/lib
#sudo rm -rf //etc/apt/sources.list.d/nodesource.list

# Install Node.js Version desired IE: 13
# More info:
#   https://github.com/nodesource/distributions/blob/master/README.md#debinstall
#curl -sL https://deb.nodesource.com/setup_13.x | sudo -E bash -
#sudo apt-get install -y nodejs
```

We're currently installing [node 14](https://github.com/laravel/settler/blob/master/scripts/provision.sh#L30) so the previous version would be 13. The link also gives users enough context to find older versions with the proper instructions for Ubuntu 20.04

svpernova09 (approved):
Thanks!

PR #1684, opened 2021-08-23
Removed older php version in wsl-init in favor of the "Feature scripts for old PHP versions"

litan1106:
Changes:

- Removed older php version in wsl-init in favor of the "Feature scripts for old PHP versions" in https://github.com/laravel/homestead/releases/tag/v12.3.1


svpernova09:
Thanks!

PR #1690, opened 2021-09-03
Fix shebang for `php81.sh` script

om4csaba:
Hi, 

Restore interpreter declaration.
See: https://github.com/koalaman/shellcheck/wiki/SC1104

_Thanks for making this sw!_


svpernova09 (approved):
Thanks!

PR #1691, opened 2021-09-03
feat(php) enable php-fpm on box startup

irineujunior:
After the box restart, the php version do not start automatically.
This pull request revert this.

Thanks


svpernova09 (approved):
Thanks, this has been on the back burner because I'm not quite ready for people to use these.

PR #1714, opened 2021-10-16
Enforce `fastcgi.conf`

juangiordana:
Enforce `fastcgi.conf`.

> In 0.8.30 (released: 15th of December 2009) Igor then included
> `fastcgi.conf` which was the same as `fastcgi_params` except
> including the improved SCRIPT_FILENAME `fastcgi_param`. This meant
> that the community could now start recommending people include
> `fastcgi.conf` instead of recommending moving `SCRIPT_FILENAME` into
> `fastcgi_params`.

https://blog.martinfjordvald.com/nginx-config-history-fastcgi_params-versus-fastcgi-conf/

### DIFF
--- a/scripts/site-types/cakephp3.sh
+++ b/scripts/site-types/cakephp3.sh
@@ -76,12 +76,13 @@ server {
 
     location ~ \.php$ {
         try_files \$uri =404;
-        include fastcgi_params;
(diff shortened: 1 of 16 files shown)


svpernova09:
Thanks for taking the time to do this. I'm going to pass, I'm not clear on what problem this is supposed to be fixing.

Issue #1722, opened 2021-10-30
Mailhog & Golang are still x86_64 on the Homestead ARM Build


Fossil01:
Just installed Homestead for M1 ARM and it's working beautifullly. I just noticed that Mailhog is still x86 instead of ARM64 :-)

`$ file /usr/local/bin/mailhog /usr/local/bin/mailhog: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=5bd20561544187a6b7e9f36d96e025f323b45628, with debug_info, not stripped`

[EDIT]
Oh the same with Golang apparently: 

`/usr/local/go/bin/go: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked`

Fossil01:
Ok so there is no ARM64 release of Mailhog but you can just do `go get github.com/mailhog/MailHog` after installing Go and then the ARM64 binary is found in `go/bin`. Seems to work fine on my install.

svpernova09:
thanks for reporting this stuff!

svpernova09:
Mailhog will require a new base box build. Go has been updated in `main` branch here in Homestead.

Fossil01:
Just a note: I tried that ARM release from the Mailhog Github releases page but it wouldn't run in Ubuntu ARM. That's why I compiled it from the source. Maybe I did something wrong, don't know but writing this comment just to be sure :-)

svpernova09:
I've fixed the golang to use the correct binary. I'm also seeing what you're reporting with Ubuntu ARM Mailhog not working. I guess an alternative would be to include go and use `go get` might be a good option. Going to give that a spin in settler.

svpernova09:
🤔ARM binaries not running in a virtualized ARM OS from an ARM host sounds like a provider bug. Are there any known Parallels issues?

I would hate to have to pull source code and compile these dependencies because **go lang was supposed to fix this**!

Fossil01:
Might be because it's probably armv7 and not arm64/armv8. When I did `go get github.com/mailhog/MailHog` i got a working binary.

Now I have another issue where curl segfaults, but only on one URL (so far) and only over https... `curl https://webservice.fanart.tv` beats me. wget does it too. Looks like there's a bug maybe in the arm64 version of openssl.

Fossil01:
@svpernova09 ngrok also is  x86-64 still :-)

svpernova09:
> 
> 
[...]

PR #1734, opened 2021-11-26
Update minio feature to be aware of CPU arch

a-yasui:
I thought This problem is a kind of the aware of cpu arch.

The same as https://github.com/laravel/homestead/issues/1722 and https://github.com/laravel/homestead/pull/1723


svpernova09 (approved):
Thanks!

bobwwilson:
Hello, does Minio support Apple Silicon in this last update and if not, will it in the near future?

PR #1738, opened 2021-12-05
Add test aliases for both Pest and PHPunit dynamically

lloricode:
This is from @freekmurze blog https://freek.dev/2142-a-bash-function-to-run-tests-for-both-phpunit-and-pest


svpernova09 (approved):
Thanks

PR #1747, opened 2021-12-20
Wordpress - Enable debug mode automatically

dominiccarrington:
Due to homestead being a development platform, enabling WP_DEBUG automatically makes sense.


svpernova09 (approved):
Thanks

Issue #756, opened 2017-12-09
Default php version?


steveheinsch:
I know you can set the php version on a per site basis, but is there a way to set the default php version for all sites to use so it doesn't use whatever is the most current version as default if you don't specify a version per site?

svpernova09:
Not at this time.

--
- Joe Ferguson
JoeFerguson.me
MemphisPHP.org

piotr-cz:
I'd find it very handy, as setting site-specific PHP version doesn't work for `apache` type (#630).
That way I could at least choose global version.

workaround, assuming you want switch to php 7.0:
1) add required php version to one of the sites and run provision which will install it: `vagrant reload --provision`
2) to change CLI, change symbolic link in `/etc/alternatives/php` to `/usr/bin/php7.0`
3) switch apache modules
```sh
sudo a2dismod php7.2
sudo a2enmod php7.0
sudo service apache2 restart
```

piotr-cz:
This has been resolved by https://github.com/laravel/homestead/issues/892, but still on have to run `vagrant reload --provision` to recreate sites.

PR #1756, opened 2022-01-26
Use Homestead file's PHP setting for site default

ethanclevenger91:
Per #756

Current proposal uses the same PHP currently used for CLI default. Could make it a separate setting, though this seems more sensible. Doesn't break existing configurations. May unexpectedly change site PHP versions for folks who are changing their CLI default but seems like a risk worth taking?

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -328,6 +328,7 @@ class Homestead
         load_balancer = settings['load_balancer'] ||= false
         http_port = load_balancer ? '8111' : '80'
         https_port = load_balancer ? '8112' : '443'
+        site_php = settings.has_key?('php') ? settings['php'] : '8.1'
 
         if load_balancer
           config.vm.provision 'shell' do |s|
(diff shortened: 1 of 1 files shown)


scripts/homestead.rb
```diff
@@ -328,6 +328,7 @@ def self.configure(config, settings)
         load_balancer = settings['load_balancer'] ||= false
         http_port = load_balancer ? '8111' : '80'
         https_port = load_balancer ? '8112' : '443'
+        site_php = settings.has_key?('php') ? settings['php'] : '8.1'
```
ethanclevenger91, line 331: I guess `||=` would work here. The above lines don't suggest a style preference. I wasn't sure if it would throw an undefined of some kind. Don't write enough Ruby.

svpernova09:
Thanks for taking the time, but I'm going to pass on this. This would be changing a pretty fundamental part of what to expect in regards to what PHP version defaults are in which version of Homestead/Settler.

ethanclevenger91:
@svpernova09 Creating a unique setting would solve that. Something like `php_fpm_version` (though I'm making an assumption about the implementation of PHP there I guess)

PR #1773, opened 2022-03-10
ignore *.log

marco-acorte:
to avoid log file in root project (ex. virtualbox log files)

immagine


svpernova09:
Going to pass on this, but you can certainly add it to your global git ignore configuration on your system.

PR #1792, opened 2022-08-12
Drop MySQL feature example

browner12:
the "mysql" feature no longer exists, so we'll drop it from the example config files.

#1582 dropped the `mysql8.sh` file.

confirmed running with the "mysql" feature produces `homestead: Invalid feature: mysql`

### DIFF
--- a/resources/Homestead.json
+++ b/resources/Homestead.json
@@ -24,9 +24,6 @@
     ],
     "features": [
         {
-            "mysql": true
-        },
-        {
             "mariadb": false
         },
         {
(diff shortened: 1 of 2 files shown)


svpernova09:
Sorry for taking so long on this, what I wanted to dig into (that I'm just now getting to) is that I believe our database creation at provision uses these features to try to figure out/guess what services need databases created.

Now that we've moved the create database scripts to only create if they find PIDs for the respective service it might be safe to remove that Ruby code.

svpernova09:
I am merging with the other Services changes for the next major release.

PR #1798, opened 2022-09-14
Feature oh my zsh : set theme and plugins from config

UnSeulT:
With this we can set (and keep after a `vagrant destroy` command) at least the theme and plugins for our oh my zsh config. If any of `theme` or `plugins` is not set, we keep the default config.


svpernova09 (approved):

svpernova09:
Thanks for this!

PR #1808, opened 2022-10-12
add Laravel Pint alias

browner12:
Pint now comes pre-installed with every new Laravel installation. Adding an alias so we can reference it more easily.


svpernova09:
Thanks!

Issue #1816, opened 2022-11-14
MariaDB crashes vagrant up/provision when creating databases specified in Homestead.yaml


elcheco:
MariaDB crashes vagrant up/provision when creating databases specified in Homestead.yaml

## Versions
  - Vagrant: 2.3.2
  - Provider: Parallels
  - Homestead: 13.3.2

## Host operating system
macOS Ventura

## Expected behavior
Databases should be created and up/provision script should continue to run

## Actual behavior
When the databases should be created, the script dies with the message:
```
homestead: Running: script: Creating MySQL / MariaDB Database: homestead
    homestead: We didn't find MySQL ($mysql), skipping $DB creation
    homestead: ERROR 1698 (28000): Access denied for user 'root'@'localhost'
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
>
```

## Steps to reproduce
1. specify some databases in Homestead.yaml and change the config to use MariaDB:
```
databases:
    - homestead
    - projectdb

features:
    - mariadb: true
    - mysql: false
    ...
```
2. run vagrant up

## Problem and solution
I have found the setup for MySQL/MariaDB in the file `scripts/create-mysql.sh`. Obviously, there's some issue for the `root` user.

For the test, I changed the command under MariaDB and everything worked:
```
mysql -u homestead -psecret -e "CREATE DATABASE IF NOT EXISTS \`$DB\` DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci";
```
I tested to call `mysql` command under `homestead` user and it worked, but under `root` user it fails. Both users have `.my.cnf` files, but under root it doesn't work, all the time the same error:

```
~ sudo su -l
root@homestead:~# mysql
ERROR 1698 (28000): Access denied for user 'root'@'localhost'
root@homestead:~# mysql -u root -p
Enter password: 
ERROR 1698 (28000): Access denied for user 'root'@'localhost'
```

elcheco:
Same was reported in the issue #1812 
but the fix [1813](https://github.com/laravel/homestead/pull/1813/files) - doesn't solve it at all. I have this in my latest version. The fix just add a .my.cnf file for homestead user, but the create db command is called under root user -> that's what the error message basically says - login to mysql under root@localhost fails...

elcheco:
I created the hotfix in pull request - #1817

[...]

PR #1817, opened 2022-11-15
Fix MariaDB set-up script - issue #1816 - vagrant up crashes on create database…

elcheco:
After installation set-up in create-mysql.sh script is failing on MariaDB , because the root@localhost doesn't have the password. This fixes it and the databases specified in the Homestead.yaml are then properly created


svpernova09 (approved):

svpernova09:
Thanks for this change. It's appreciated.

PR #1827, opened 2023-01-03
[WIP] Add feature for php8.2

csaba215:
I also added php8.2-swoole package for use with Laravel Octane.


svpernova09:
Thanks for taking the time to do this. I'd rather not include Swoole in Homestead. This would be a perfect addition to your `after.sh`.

PR #1831, opened 2023-01-07
[WIP] Add basic libvirt support

csaba215:
For storage I used virtiofs. (NFS didn't work for me, maybe it was because my host os, virtiofs is way faster anyway)

Example config:
https://gist.github.com/csaba215/b96e63377835db50e0cb7cae14047ebf
(Using the box I built for libvirt)
At the moment I get 2 network interfaces, 1 received the ip in the config, the other received ip from the default network.


svpernova09 (approved):
Thanks

PR #1834, opened 2023-01-09
Update meilisearch for aarch64 support

josezenem:
This updates meilisearch to work with aarch64, related to this thread: https://github.com/meilisearch/meilisearch/discussions/2048, the if statement may not be needed as the sh command would do the check and install, however I left it in as I can only test it using the arm aarch64 variation.


svpernova09:
Thanks for this! You can remove the if conditional check completely if the package name is the same on both arch.

josezenem:
@svpernova09 all set, removed the if statement check.

svpernova09 (approved):

PR #1843, opened 2023-01-17
sync up the example file

browner12:
@svpernova09 can you explain what the difference is supposed to be between `Homestead.yaml` and `Homestead.yaml.example`?

### DIFF
--- a/Homestead.yaml.example
+++ b/Homestead.yaml.example
@@ -21,7 +21,9 @@ databases:
     - homestead
 
 features:
+    - mysql: true
     - mariadb: false
+    - postgresql: false
     - ohmyzsh: false
     - webdriver: false
 
@@ -31,9 +33,16 @@ services:
 #    - disabled:
 #        - "postgresql@11-main"
 
-# ports:
-#     - send: 50000
-#       to: 5000
-#     - send: 7777
-#       to: 777
-#       protocol: udp
+#ports:
+#    - send: 33060 # MySQL/MariaDB
+#      to: 3306
+#    - send: 4040
+#      to: 4040
+#    - send: 54320 # PostgreSQL
+#      to: 5432
+#    - send: 8025 # Mailhog
+#      to: 8025
+#    - send: 9600
+#      to: 9600
+#    - send: 27017
+#      to: 27017


svpernova09:
> @svpernova09 can you explain what the difference is supposed to be between `Homestead.yaml` and `Homestead.yaml.example`?

`Homestead.yaml.example` should be an example configuration that is mostly ready to use. At a minimum, you'd have to map a real project folder.

browner12:
could we get rid of the `Homestead.yaml` file, and then copy the example file when we init?  This could help avoid the duplication, and things getting out of sync.

svpernova09:
> could we get rid of the `Homestead.yaml` file, and then copy the example file when we init? This could help avoid the duplication, and things getting out of sync.

https://github.com/laravel/homestead/blob/120b71524e961aa0b6738de158b1dfb36e5c0995/tests/MakeCommandTest.php#L179

`Homestead.yaml.example` is what gets used when using the example flag, we could move that to use the same files as the make command which is https://github.com/laravel/homestead/blob/main/resources/Homestead.yaml or https://github.com/laravel/homestead/blob/main/resources/Homestead.json 

Then we could remove `Homestead.yaml.example`.

browner12:
You understand that stuff a lot better than I do, so I'll defer to you on that.

svpernova09:
> You understand that stuff a lot better than I do, so I'll defer to you on that.

I'm honestly surprised we haven't had to dive into that code keep PHP compatibility, but maybe no one is testing it yet :D 

IMHO leave the `Homestead.yaml.example` until someone has a good enough reason to dive into the make command / example code and update that.

svpernova09 (approved):
Thanks!

browner12:
is that all "per project" Homestead installation code? I honestly have no clue what it does.

I've always done global Homestead installs.

svpernova09:
> is that all "per project" Homestead installation code? I honestly have no clue what it does.
> 
> I've always done global Homestead installs.

Yep! it's a PHP implementation of creating the `Homestead.yaml` that you'd only use w/ per project installs.

roberto-aguilar:
I can confirm! It installs Homestead and sets up either the Yaml or JSON based on the usage or lack of [the json option](https://github.com/laravel/homestead/blob/main/src/MakeCommand.php#L58) in a per-project basis.

I implemented / refactored that code like 6 years ago, but I'll be happy to help if you guys want to make any changes to it 😃 

Miss you a lot, guys ❤️ hope you are doing great!

PR #1846, opened 2023-01-20
update for Ngrok v3

browner12:
- adjust call to start ngrok
- update path to Ngrok config

## To Do

- I think the `wsl-init` file will need to be updated to v3, but I'm a little confused which download it should be using.
- I think we need a better solution in the `create-ngrok.sh` script because I believe a default config file will be created in the base box, but we still need to add the `web_addr` config option. Not sure how to do that, so looking for some help.

### DIFF
--- a/resources/aliases
+++ b/resources/aliases
@@ -233,7 +233,7 @@ function serve-pimcore() {
 function share() {
     if [[ "$1" ]]
     then
-        ngrok http ${@:2} -host-header="$1" 80
+        ngrok http ${@:2} $1:80
     else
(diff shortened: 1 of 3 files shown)


svpernova09 (approved):
Thanks for the update!

browner12:
sorry, I should have marked this "Draft" because I don't think it works completely as is.  I believe the config file requires a `version` and `region` field that we are not currently adding.

It's very easy to manually update with the `ngrok config upgrade` command, but I think we still want to make it work out of the box for users.

PR #1847, opened 2023-01-26
Drop Yarn

browner12:
I know Yarn was a good stop-gap years ago for when there were serious shortcomings with NPM, but now that NPM has gotten its act together, I think we can drop Yarn support.

Yarn users are welcome to add this into their own custom aliases, but I think we're covering the 99% use case here.

### DIFF
--- a/bin/wsl-init
+++ b/bin/wsl-init
@@ -186,7 +186,6 @@ apt-get install -y nodejs
 /usr/bin/npm install -g npm
 #/usr/bin/npm install -g gulp-cli
 #/usr/bin/npm install -g bower
-/usr/bin/npm install -g yarn
 #/usr/bin/npm install -g grunt-cli
 
 ## Install rabbitmq-server and its dependencies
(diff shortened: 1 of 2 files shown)


svpernova09:
Thanks!

ankurk91:
bower must be removed as well.

svpernova09:
Just because we're dropping the installs in Homestead does not mean we drop them in Settler. This way if you still need Yarn, it'll be in the [base](https://github.com/laravel/settler/blob/master/scripts/amd64.sh#L607) [box](https://github.com/laravel/settler/blob/master/scripts/arm.sh#L607) install.

browner12:
Joe, I'm a little confused here I guess. Were these scripts doing anything, then, if it was already in the base box?

svpernova09:
`resources/aliases` is where we store the default aliases for Homestead. By removing the Yarn aliases they won't work out of the box anymore, but the Yarn binary will still be in the base box, so if users need Yarn, it's still there for them.

The `bin/wsl-init` script is to build a settler/homestead-like environment for Windows Subsystem for Linux, it's outdated and needs to be updated w/ the latest Settler changes.

browner12:
Gotcha, thanks!  Yah, the WSL script was the one throwing me off. I've sent in another PR for that.

PR #1849, opened 2023-01-27
remove duplicate installs

browner12:
these things are already installed in the base box

https://github.com/laravel/settler/blob/master/scripts/arm.sh#L602 https://github.com/laravel/settler/blob/master/scripts/amd64.sh#L602


svpernova09:
`bin/wsl-init` Does not use the base box at all, this script is what turns a Ubuntu 20.04 installation into Homestead-ish. Just merged #1850 to update this file to Settler 13

PR #1851, opened 2023-01-30
Libvirt: fix vm naming

csaba215:
Currently the vm name is based on the current directory and name parameter.(e.g. homestead_homestead when directory is homestead and vm name is homestead) It should be based on the name parameter only.


svpernova09 (approved):
Thanks

PR #1856, opened 2023-02-13
Always allow symlink creation in shared folders when using VirtualBox provider on a Windows host OS

Sn0wCrack:
This saves people from having to edit their own Vagrantfile when on Windows and need to use symlinks.

My main reason for proposing this is for people who are using Homestead directly from the repository, as the main Vagrantfile is tracked in git, which can make upgrading Homestead versions, or setting up a new installation, more annoying.

As far as I'm aware there is no detriment to having this enabled always when using Windows.

According to the [VirtualBox documentation](https://www.virtualbox.org/manual/ch04.html#sharedfolders), this option is actually required on all host operating systems if you wish to create symbolic links inside of a shared folder, but it is noted in the [Homestead documentation](https://laravel.com/docs/9.x/homestead#symbolic-links-on-windows) this is only needed for Windows, thus why I've gated this to solely windows.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -55,6 +55,10 @@ class Homestead
       if settings.has_key?('paravirtprovider') && settings['paravirtprovider']
         vb.customize ['modifyvm', :id, '--paravirtprovider', settings['paravirtprovider'] ||= 'kvm']
       end
+      
+      if Vagrant::Util::Platform.windows?
+        vb.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate/v-root", "1"]
+      end
     end
 
     # Override Default SSH port on the host


svpernova09:
Thanks for taking the time for this change but I'm going to pass, the Vagrantfile should be regenerated as needed and Homestead has protections built in to the best effort to prevent people from running incompatible base boxes.

Sn0wCrack:
@svpernova09 Any chance you could further explain the "Vagrantfile should be regenereated as needed" part? I'm assuming you mean edited by the end user as part of their own setup?

Like I mentioned in the initial comment this is was more intended for people who are probably going to use Homestead directly from the repository with multiple projects at once, rather than as a package in composer for individual projects.

If the concern is that we're taking away the option from the user to enable symlinks, would this maybe be better then do add as an option to the configuration file so people are able to enable and disable it as they choose?

Or maybe a way to add in a list of acceptable directories that symlinks can be created in?

For example:
```yaml
symlinks: true

// or...

symlinks:
 - /home/vagrant/website1
 - /home/vagrant/website2
```

svpernova09:
Will look into this again. Thanks.

svpernova09:
Apologies, I was thinking this was doing something in the `Vagrantfile`.

svpernova09:
Thanks.

PR #1859, opened 2023-02-18
Libvirt: Use virtiofs for default when no folders are defined.

csaba215:
Use virtiofs for default when, no folders are defined.
Also allow changing the type of user mounts. e.g.:
```
folders:
    - map: ~/code
      to: /home/vagrant/code
      type: rsync
    - map: ~/code2
      to: /home/vagrant/code2
```


svpernova09 (approved):
Thanks

Issue #1860, opened 2023-02-20
Feature Request: Add Mailpit integration


litan1106:
Mailpit is the new replacement for Mailhug but much faster.

The laravel  v10 env now uses mailpit https://github.com/laravel/laravel/blob/10.x/.env.example#L32

https://github.com/axllent/mailpit

svpernova09:
I'd take a PR accepting this but have no immediate plans to do this.

Sn0wCrack:
I'd be happy to take a crack at this potentially.

A couple ways I could see this going is:

- Install mailpit and a systemd unit in Settler and then use the "services" section in the Homestead configuration file to enable an disable the systemd unit for mailhog and mailpit.
- Similar to the the MariaDB feature script, add a mailpit feature script that uninstalls mailhog and replaces it with mailpit.

If there's a preference either way I'm happy to accommodate and go that route.

PR #1864, opened 2023-03-06
Add feature for installing Mailpit and disabling Mailhog

Sn0wCrack:
Resolves https://github.com/laravel/homestead/issues/1860

This just uses the install script provided on mailpit's repository which will download the correct binary based on the OS and Architecture you're running, it supports arm64 so should work fine on Apple Silicon devices.

It will also disable mailhog's systemd service, however the mailhog binary will remain installed on the system if the user wishes to switch back.

### DIFF
--- a/scripts/features/mailpit.sh
+++ b/scripts/features/mailpit.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+
+if [ -f ~/.homestead-features/wsl_user_name ]; then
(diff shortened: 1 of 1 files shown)


svpernova09 (approved):
Looks good. Thanks

litan1106:
@Sn0wCrack I love this feature approach. People can still use mailhog as default but can opt-in mailpit with features.

PR #1865, opened 2023-03-09
publish composer autocompletions on provision

browner12:
https://getcomposer.org/doc/03-cli.md#bash-completions

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -657,6 +657,13 @@ class Homestead
       s.privileged = false
     end
 
+    # Publish Composer Bash Completions On Every Provision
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for looking into this. Instead of having to run this each time we run `vagrant up`, let's move this code to https://github.com/laravel/settler/blob/master/scripts/amd64.sh#L491-L502 and https://github.com/laravel/settler/blob/master/scripts/arm.sh#L491-L502 so it happens in Settler. 

Until the next base box release, I'd recommend putting this in `after.sh`

svpernova09:
Closing for now.

PR #1866, opened 2023-03-10
Faster first startup of feature mariadb and fixing upgrading packages

marcheffels:
Removing mysql-common upgrades many other packages due to package dependencies. This costs a lot of time and leads to unintentional changes in other packets. Removing the mysql-common package is not required for installing mariadb-server and mariadb-client, the package just needs to be updated from the mariadb repository.

Current first Startup (vagrant up) takes 29:05 minuntes. [Logfile](https://gist.github.com/marcheffels/af410d8aefb3996d60e65d079ce68bc9)
With this PR first Startup (vagrant up) takes 06:05 minuntes. [Logfile](https://gist.github.com/marcheffels/6985931935fba7585f4a82d49a6394bf)

to reproduce you only need this change in your Homestead.yaml
`features:`
`    - mysql: false`
`    - mariadb: true`

### DIFF
--- a/scripts/features/mariadb.sh
+++ b/scripts/features/mariadb.sh
@@ -25,7 +25,7 @@ service apparmor stop
 update-rc.d -f apparmor remove
 
 # Remove MySQL
-apt-get -o Dpkg::Options::="--force-confnew" remove -y --purge mysql-server mysql-client mysql-common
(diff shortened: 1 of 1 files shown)


svpernova09 (approved):
Thank you *very* much for tracking this down. Much appreciated.

svpernova09:
Took 1:50 on my M1 mac mini w/ Parallels.

```
vagrant up: 01:50
```

PR #1876, opened 2023-05-23
Install rabbitmqadmin CLI tool when enabling rabbitmq feature

Sn0wCrack:
This tool is useful for administering rabbitmq locally, allowing you to create exchanges, queues and bindings on the CLI.

I've also adding in topic permissions for the default homestead user to ensure they have access to all topics on all exchanges.


svpernova09 (approved):
Thanks!

PR #1877, opened 2023-05-23
Add Logstash feature

Sn0wCrack:
This is mostly duplicated from the Elasticsearch feature since Elastic uses the same apt repository for all their software.

Unsure if any issues could arise if you select a different version of Elasticsearch from Logstash, but generally speaking this would cause issues running them anyways as the major versions of these two generally need to match.


svpernova09 (approved):
Thanks

PR #1879, opened 2023-05-24
Prevent duplicate features on provision 

LlamaDuckGoose:
We have multiple sites in the homestead yaml that use the same PHP version and during the provision it will try and install the same PHP version over and over again.

This should only get the unique features and install them.


svpernova09 (approved):
Thanks!

PR #1889, opened 2023-09-15
Permits specifying MariaDB version

sdespont:
By using the following description, the specified version of MariaDB will be installed.

```
features:
    - mysql: false
    - mariadb:
        version: 10.6
services:
    - enabled:
        - "mysql"
```

Source : https://computingforgeeks.com/how-to-install-mariadb-on-ubuntu-lts-linux/

### DIFF
--- a/scripts/features/mariadb.sh
+++ b/scripts/features/mariadb.sh
@@ -33,8 +33,19 @@ rm -rf /var/lib/mysql/*
 rm -rf /var/log/mysql
 rm -rf /etc/mysql
 
+# Determine version from config
+
+set -- "$1"
+IFS=".";
+
 # Add Maria PPA
-curl -LsS https://downloads.mariadb.com/MariaDB/mariadb_repo_setup | sudo bash
+if [ -z "${version}" ]; then
+    curl -LsS https://downloads.mariadb.com/MariaDB/mariadb_repo_setup | sudo bash  
+else
+    curl -LsS -O https://downloads.mariadb.com/MariaDB/mariadb_repo_setup | sudo bash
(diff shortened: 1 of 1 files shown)


svpernova09:
We did this once for Redis and I regretted it. Going to pass but thanks for the PR

sdespont:
@svpernova09 ok, it works well for version 10.6 by the way. How would you go about installing the specific version of MariaDB 10.6?

LlamaDuckGoose:
@sdespont
As a work around I've manually updated the feature script for mariadb
```# Add Maria PPA
curl -LsS https://downloads.mariadb.com/MariaDB/mariadb_repo_setup | sudo bash -s -- --mariadb-server-version=mariadb-10.11.5 --skip-eol-check
```

@svpernova09 I'd like this to be relooked since being able to set versions will be handy when trying to keep it the same as you'd have in a production cluster, also software is a bit of a pain when installing the newest version (https://github.com/TablePlus/TablePlus-Windows/issues/622)

svpernova09:
Have fun storming the castle. Please don't document this as it may break :D :D :D

svpernova09:
I'm attempting to build new base boxes, this has been merged to `main` but I may not tag a release until after I can get an updated base box.

LlamaDuckGoose:
Just did a quick test and it's definitely working

```
features:
    - mysql: false
    - mariadb:
        version: 10.11
 ```

image

Side note, I'm not sure how possible it is in the base box but it would be awesome if it could pick mirrors closer to the user instead of always using the US ones.

sdespont:
@LlamaDuckGoose Thanks for the feedback ! Working well with version 10.6 too.

PR #1890, opened 2023-09-15
Permits SSH logging with login and password

sdespont:
By adding the following keys in `homestead.yaml` file, it would permit to log into the box using login and password method.

```
ssh_username: vagrant
ssh_password: vagrant
```

image


browner12:
I'm confused. Do you want to set different username/password combos? 

The default is already `vagrant:vagrant`.

sdespont:
@browner12 Oh ! Didn't know that.... sorry

browner12:
No worries. Glad it does what you need.

PR #1891, opened 2023-09-21
Apache proxy site-type

LlamaDuckGoose:
This works the same way the the nginx proxy site-type but instead uses apache bases on the apache site type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ phpunit.xml
 wsl-init.log
 .backup/*
 .DS_Store
+/ext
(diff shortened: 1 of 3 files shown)


LlamaDuckGoose:
If we're happy with this, I'll create a PR for the docs to include this under the site type section :)

svpernova09 (approved):
Proxying with Apache? Sounds terrible. Have fun storming the castle. Thanks.

svpernova09:
> If we're happy with this, I'll create a PR for the docs to include this under the site type section :)

Go for it, feel free to tag me in the docs PR.

PR #1892, opened 2023-09-21
Ngrok default param for apache share 

LlamaDuckGoose:
It looks like trying to share an apache site won't work without adding this param.
```
--host-header=rewrite
```

From what I've tested this has no effect on nginx but will fix it working for apache so defaulting this won't have any negative effects and will work for both.

I've also included an alias for share over port 443 as well `sshare` (secure share) but if you're not happy with this I can remove it.

### DIFF
--- a/resources/aliases
+++ b/resources/aliases
@@ -227,7 +227,7 @@ function serve-pimcore() {
 function share() {
     if [[ "$1" ]]
     then
-        ngrok http ${@:2} $1:80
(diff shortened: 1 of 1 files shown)


svpernova09 (approved):
Seems safe as long as users avoid https://manpages.ubuntu.com/manpages/focal/man1/sshare.1.html :D

Thanks

PR #1903, opened 2023-10-26
Fix apache proxy header forwarder

LlamaDuckGoose:
Looks like I missed including these two options when I created this, which causes this to fail in some cases


LlamaDuckGoose:
These two should now function the same
https://github.com/laravel/homestead/blob/main/scripts/site-types/proxy.sh#L44-L49

svpernova09 (approved):

svpernova09:
Thanks.

PR #1905, opened 2023-10-29
Add Rustc feature

LlamaDuckGoose:
Allows for setting rustc as a feature and it'll install so it's possible to use cargo and rustc
https://www.rust-lang.org/tools/install


svpernova09 (approved):
Thanks!

PR #1906, opened 2023-11-02
Fix RVM installer

LlamaDuckGoose:
Looks like the current key server it tries doesn't exist anymore and I've updated it to match https://rvm.io/rvm/install
```
gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
gpg: directory '/home/vagrant/.gnupg' created
gpg: keybox '/home/vagrant/.gnupg/pubring.kbx' created
gpg: keyserver receive failed: No name
```

after that it looks like `source /home/vagrant/.rvm/scripts/rvm` doesn't exist and it wants to use `source /usr/local/rvm/scripts/rvm`

```
    homestead: 1 gem installed
    homestead: 
    homestead:   * To start using RVM you need to run `source /usr/local/rvm/scripts/rvm`
    homestead:     in all your open shell windows, in rare cases you need to reopen all shell windows.
    homestead: /tmp/vagrant-shell: line 27: /home/vagrant/.rvm/scripts/rvm: No such file or directory
went wrong.
```

### DIFF
--- a/scripts/features/rvm.sh
+++ b/scripts/features/rvm.sh
@@ -20,7 +20,7 @@ touch /home/$WSL_USER_NAME/.homestead-features/rvm
 chown -Rf $WSL_USER_NAME:$WSL_USER_GROUP /home/$WSL_USER_NAME/.homestead-features
 
 # Install RVM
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks.

Issue #1912, opened 2023-11-14
There is no InfluxDB v2.x


PerikiyoXD:
Why is Is InfluxDB v2.x not already added?

Is there any inconvenience on having that specific package?

I work with that and I'd find it useful if we got it too.

svpernova09:
I'd accept a PR updated our InfluxDB install to install V2

PerikiyoXD:
@svpernova09 there you go

svpernova09:
Thanks for the PR!

PR #1914, opened 2023-11-19
Upgrade InfluxDB feature to v2 (v2.7.4)

PerikiyoXD:
#1912 

This is the PR to upgrade the current InfluxDB feature script.


PerikiyoXD:
Redoing PR as it should use `main` branch instead of `release`

PR #1915, opened 2023-11-19
Upgrade InfluxDB feature to v2 (v2.7.4)

PerikiyoXD:
Upgrade InfluxDB feature script to install InfluxDB v2.x (v2.7.4) instead of v1.x

Source issue: #1912


svpernova09 (approved):
Thanks!

PR #1917, opened 2023-11-24
before.sh

ALameLlama:
There have been cases where I want to make changes to Ubuntu before any of the homestead provisioning stuff has taken place.

I've manually done this for my own use cases but figured others might want to do the same with a more official way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 phpunit.xml
 /Homestead.yaml
 /Homestead.json
+/before.sh
 /after.sh
 /aliases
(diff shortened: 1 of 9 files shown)


svpernova09:
Thanks for taking the time for this but I'm going to pass for now. One of the foundations of Homestead is that you're booting up a VM from a known state. If you tinker with that state before Homestead runs you run the risk of breaking Homestead or causing other issues within the VM.

Glad you found something that worked for you but I'm pessimistic about the support issues and debugging this could bring.

PR #1923, opened 2024-01-14
Update wordpress.sh

vsmash:
escape $ in location redirect.
I found that the $1 was in the nginx config on vagrant machine. (arm64/parallels)
I escaped the $1 to fix. I also removed the https:// before ${11} (but not in this pr) as i prefer to specify the scheme in the yaml

### DIFF
--- a/scripts/site-types/wordpress.sh
+++ b/scripts/site-types/wordpress.sh
@@ -59,7 +59,7 @@ block="server {
     }
 
     location @prod_site {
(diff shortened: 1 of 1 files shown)


svpernova09:
Thanks for taking the time for this.

> prefer to specify the scheme in the yaml

The `Homestead.yaml` isn't aware of a scheme and shouldn't be.

PR #1925, opened 2024-01-17
set "mysql" as an enabled database

browner12:
MySQL is kind of a weird case, because it's our default database, and it cannot (shouldn't?) run side by side with MariaDB.

We kind of cheated this previously by having "mysql" as a key in the "features" configuration, but it's technically not a "feature" as it doesn't have a file in `/scripts/features/`.  this resulted in an "Invalid feature: mysql" warning in the provision script.

This change will allow us to remove `mysql: true` from our docs and example files, while still having MySQL work correctly.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -604,6 +604,11 @@ class Homestead
         end
       end
 
+      # Enable MySQL if MariaDB is not enabled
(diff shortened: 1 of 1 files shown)


svpernova09 (approved):
Thanks.

PR #1926, opened 2024-01-17
drop "mysql8" reference

browner12:
this is a remnant of long ago in our v5 to v8 transition.  we now *only* install MySQL8, so we don't really need this anymore.  if people still have it in their `homestead.yaml` it'll still be fine with my other PR.

this also syncs it up with the duplicated code we have farther down.


svpernova09 (approved):
old cruft!

PR #1930, opened 2024-01-17
mark installation as "complete" at the end of script

browner12:
we use an empty file in `~/.homestead-features` to mark a "feature" as installed. we should move to the end of the file, so we know the install has fully completed successfully.

I ran into this issue with the "webdriver" install. because Chrome does not have an ARM version, the install would fail, but the file would already be created, the provision scipt thought it succeeded.


svpernova09:
I don't want to care about the state of the installation. That's just out of scope for Homestead IMHO. 

I don't want to try to catch anything that might go wrong during an installation. That's a feature problem, not a Homestead problem.

PR #1932, opened 2024-01-19
Update apt-key usages in features

svpernova09:


scripts/features/r-base.sh
```diff
@@ -19,6 +19,8 @@ fi
 touch /home/$WSL_USER_NAME/.homestead-features/r-base
 chown -Rf $WSL_USER_NAME:$WSL_USER_GROUP /home/$WSL_USER_NAME/.homestead-features
 
-apt-key adv --keyserver keyserver.ubuntu.com --recv-keys E298A3A825C0D65DFD57CBB651716619E084DAB9
-add-apt-repository 'deb https://cloud.r-project.org/bin/linux/ubuntu focal-cran40/'
+wget -qO- https://cloud.r-project.org/bin/linux/ubuntu/marutter_pubkey.asc | sudo gpg --dearmor -o /etc/apt/keyrings/r-project.gpg
```
browner12, line 22: switch to `curl` for consistency?

scripts/features/rabbitmq.sh
```diff
@@ -19,53 +19,41 @@ fi
...
-## Team RabbitMQ's main signing key
-sudo apt-key adv --keyserver "hkps://keys.openpgp.org" --recv-keys "0x0A9AF2115F4687BD29803A206B73A36E6026DFCA"
-## Launchpad PPA that provides modern Erlang releases
-sudo apt-key adv --keyserver "keyserver.ubuntu.com" --recv-keys "F77F1EDA57EBB1CC"
-## PackageCloud RabbitMQ repository
-sudo apt-key adv --keyserver "keyserver.ubuntu.com" --recv-keys "F6609E60DC62814E"
+# Import signing keys
+curl -1sLf "https://keys.openpgp.org/vks/v1/by-fingerprint/0A9AF2115F4687BD29803A206B73A36E6026DFCA" | sudo gpg --dearmor -o /etc/apt/keyrings/com.rabbitmq.team.gpg
+curl -1sLf "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xf77f1eda57ebb1cc" | sudo gpg --dearmor -o /etc/apt/keyrings/net.launchpad.ppa.rabbitmq.erlang.gpg
+curl -1sLf "https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey" | sudo gpg --dearmor -o /etc/apt/keyrings/io.packagecloud.rabbitmq.gpg
```
browner12, line 27: switch to `curl -1fsSl` for consistency?

is the "1" required? is this server that far behind?

scripts/features/cassandra.sh
```diff
@@ -20,8 +20,9 @@ touch /home/$WSL_USER_NAME/.homestead-features/cassandra
 chown -Rf $WSL_USER_NAME:$WSL_USER_GROUP /home/$WSL_USER_NAME/.homestead-features
 
 # Install Cassandra and driver dependencies
-echo "deb http://www.apache.org/dist/cassandra/debian 311x main" | sudo tee -a /etc/apt/sources.list.d/cassandra.sources.list
-wget -q -O - https://www.apache.org/dist/cassandra/KEYS | sudo apt-key add -
+echo "deb https://debian.cassandra.apache.org 41x main" | sudo tee -a /etc/apt/sources.list.d/cassandra.sources.list
+curl https://downloads.apache.org/cassandra/KEYS | sudo apt-key add -
```
browner12, line 24: i'm confused by this one? there's like 1000 keys

scripts/features/elasticsearch.sh
```diff
@@ -38,11 +38,10 @@ echo "Elasticsearch majorVersion: $majorVersion"
 
 
 # Install Java & Elasticsearch
-
-wget -qO - https://artifacts.elastic.co/GPG-KEY-elasticsearch | sudo apt-key add -
+curl -fsSL https://artifacts.elastic.co/GPG-KEY-elasticsearch | sudo gpg --dearmor -o /etc/apt/keyrings/elasticsearch.gpg
 
 if [ ! -f /etc/apt/sources.list.d/elastic-$majorVersion.x.list ]; then
-    echo "deb https://artifacts.elastic.co/packages/$majorVersion.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-$majorVersion.x.list
+    echo "deb [signed-by=/etc/apt/keyrings/elasticsearch.gpg] https://artifacts.elastic.co/packages/$majorVersion.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-$majorVersion.x.list
```
browner12, line 44: do we *want* to append to the source file here? Since it looks like we're already creating separate files for each version, and checking if the file already doesn't exist, I think the `-a` is superfluous.

scripts/features/cassandra.sh
browner12: also, I think we can drop the `-a`.

scripts/features/logstash.sh
```diff
@@ -38,11 +38,10 @@ echo "Logstash majorVersion: $majorVersion"
 
 
 # Install Java & Logstash
-
-wget -qO - https://artifacts.elastic.co/GPG-KEY-elasticsearch | sudo apt-key add -
+curl -fsSL https://artifacts.elastic.co/GPG-KEY-elasticsearch | sudo gpg --dearmor -o /etc/apt/keyrings/elasticsearch.gpg
 
 if [ ! -f /etc/apt/sources.list.d/elastic-$majorVersion.x.list ]; then
-    echo "deb https://artifacts.elastic.co/packages/$majorVersion.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-$majorVersion.x.list
+    echo "deb [signed-by=/etc/apt/keyrings/elasticsearch.gpg] https://artifacts.elastic.co/packages/$majorVersion.x/apt stable main" | sudo tee -a /etc/apt/sources.list.d/elastic-$majorVersion.x.list
```
browner12, line 44: I think we can drop the `-a`.

scripts/features/timescaledb.sh
```diff
@@ -18,15 +18,14 @@ fi
...
+curl -fsSL https://packagecloud.io/timescale/timescaledb/gpgkey | sudo gpg --dearmor -o /etc/apt/keyrings/timescaledb.gpg
+echo 'deb [signed-by=/etc/apt/keyrings/timescaledb.gpg] https://packagecloud.io/timescale/timescaledb/ubuntu/ jammy main' | sudo tee /etc/apt/sources.list.d/timescaledb.list
 
-sudo apt-get -y install timescaledb-2-postgresql-13
+sudo apt-get update
+sudo apt-get -y install timescaledb-2-postgresql-15
 
 sudo timescaledb-tune --quiet --yes
-printf "\ntimescaledb.telemetry_level=off\n" | sudo tee -a /etc/postgresql/13/main/postgresql.conf
+printf "\ntimescaledb.telemetry_level=off\n" | sudo tee -a /etc/postgresql/15/main/postgresql.conf
```
browner12, line 28: I think we can drop the `-a`.

scripts/features/cassandra.sh
svpernova09: This is one I was fighting with so I reverted to the docs' recommendation. There are a few of your comments that have the same scenario. Instead of reinventing the install just go w/ what the docs recommend.

scripts/features/elasticsearch.sh
svpernova09: Updated to drop `-a`

scripts/features/logstash.sh
svpernova09: Updated to drop `-a`

scripts/features/timescaledb.sh
svpernova09: We do want the append here, we're throwing 1 string at the end of the existing configuration.

scripts/features/rabbitmq.sh
svpernova09: I copy/pasted from their docs because they're installing multiple keys and this was cleaner.

scripts/features/r-base.sh
svpernova09: This was weird so I reverted to docs' recommendation.

PR #1933, opened 2024-01-19
drop Mailpit feature

browner12:
this is now included by default in the base box, and doesn't require a feature flag to be activated.


svpernova09 (approved):
Thanks.

PR #1937, opened 2024-01-21
Fix NFS Mount failing on Ubuntu 22.04

Sn0wCrack:
It appears that NFS over UDP is disabled in Ubuntu 22.04 by default:

 - https://github.com/hashicorp/vagrant/issues/12758
 - https://discourse.ubuntu.com/t/jammy-jellyfish-release-notes/24668#udp-disabled-for-nfs-mounts-7

This causes NFS mounts to fail by default with Homestead without adding the `nfs_udp` option to all existing folder mounts, which if you have a lot of folders mounted like I do, can be a bit tedious.

Since Homestead will be using Ubuntu 22.04 by default going forward, I think disabling `nfs_udp` by default going forward might be a good idea.

This PR will automatically apply the `nfs_udp` option with a value of `false` when a folder is using NFS.

### DIFF
--- a/scripts/homestead.rb
+++ b/scripts/homestead.rb
@@ -212,14 +212,20 @@ class Homestead
 
           if folder['type'] == 'nfs'
             mount_opts = folder['mount_options'] ? folder['mount_options'] : ['actimeo=1', 'nolock']
+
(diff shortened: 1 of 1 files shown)


svpernova09 (approved):
Thanks for sorting this out. I certainly don't use NFS these days.

PR #1944, opened 2024-01-28
Bump symfony packages 

ALameLlama:
feat: bump symfony versions

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,9 +3,9 @@
     "description": "A virtual machine for web artisans.",
     "require": {
         "php": "^8.0 || <8.3",
-        "symfony/console": "^5.0 || ^6.0",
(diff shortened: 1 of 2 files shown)


browner12:
there are some version updates in `composer.lock` that cause the major to go down. we might want to double check that.

also, IMO lock files should only be updated by maintainers, so if we want to accept this, I think we should remove that from this commit.

ALameLlama:
Happy to remove it if needed.

svpernova09:
Thanks for taking the time to do this. I have this with another change as well as dropping the 5.0 Symfony versions since that should get us clear of [PHP 7.x](https://github.com/symfony/console/blob/5.4/composer.json#L19)

PR #1945, opened 2024-01-29
Fix Ngrok config

ALameLlama:
Ngrok updated to a newer version since they deprecated v1 which we did in here
https://github.com/laravel/settler/pull/251

but it looks like we missed updating the config.

Trying to use the share aliases without the version will result in
```
ERROR:  Error reading configuration file '/home/vagrant/.config/ngrok/ngrok.yml': `version` property is required.
ERROR:
ERROR:  If you're upgrading from an older version of ngrok, you can run:
ERROR:
ERROR:      ngrok config upgrade
ERROR:
ERROR:  to upgrade to the new format and add the version number.
```

Adding `version: 2` seems to resolve the issue and it will work like before.


svpernova09 (approved):
Thanks!

PR #1958, opened 2024-03-01
WSL Installation Upgrades and Documentation

karmendra:
- bin/homestead
  - Registered WslApplyFolderMapping command
- scripts\features\mariadb.sh 
  - Stop the mysql service before uninstall and remove the installation tracking file.
- scripts\features\mysql.sh 
  - Create feature installation script for mysql.
- scripts\features\postgresql.sh 
  - Removed `systemctl disable postgresq` from the end to make sure it works after the feature install
- src\WslApplyFeatures.php 
  - Changed the command name from `wsl:apply-features` to `wsl:features`.
- src\WslApplyFolderMapping.php 
  - Added new command to do folder mapping
- src\WslCreateDatabaseCommand.php 
  - Changed the command name from `wsl:create-databases` to `wsl:databases`.
- src\WslCreateSiteCommand.php 
  - Changed the command name from `wsl:create-sites` to `wsl:sites`.
  - Remove nginx/sites-enabled/* as well while removing existing nginx sites.
  - Add cron job for laravel schedule:run if mentioned in yaml file.
- wsl.md 
  - Documentation for Installing Homestead on Windows subsystem for Linux


karmendra:
@svpernova09 / @MichaelBelgium , I just fixed a typo that I found in wsl.md. This PR is ready to merge, let me know if there is anything else to be addressed.

@svpernova09, I have checked php8.3.sh, there are only two environment vars (`WSL_USER_NAME` and `WSL_USER_GROUP`) and these are populated from the files storing these values.

MichaelBelgium:
Looks okay for me, I don't see the benefit of the `wsl:folders` command tho. But I see its perhaps required due the change from `wsl_sites` to `sites`, which means the separate mapping for wsl will be gone

So this change is from, eg, 

```
wsl_sites:
  - map: homestead.test
    to: /mnt/c/Users/<user>/Documents/code/public
```
to
```
folders:
    - map: ~/code
      to: /mnt/c/Users/<user>/Documents/code

sites:
    - map: homestead.test
      to: /mnt/c/Users/<user>/Documents/code/public
```

If i'm correct, the command will loop through folders and create a symlink between ~/code and /mnt/c/Users/<user>/Documents/code

Although, if someone does the config from below, the symbolic link will be useless i think?

```
sites:
    - map: homestead.test
      to: /mnt/c/Users/<user>/Documents/code/public
```
in stead of 

```
sites:
    - map: homestead.test
      to: ~/code/public
```
which is the symlink

Lastly the checks you did before in wsl init, those were good, i'd keep those

```
if [[ $EUID -ne 0 ]]; then
    echo "Error: This script must be run as root.">&2

    exit 1
fi

# ....

# Set default
if [ -z "${WSL_USER_NAME}" ]; then
    WSL_USER_NAME=vagrant
fi
if [ -z "${WSL_USER_GROUP}" ]; then
    WSL_USER_GROUP=vagrant
fi

# Validate user and group
if ! id "$WSL_USER_NAME" &>/dev/null; then
    echo "Error: User $WSL_USER_NAME does not exist.">&2

    exit 1
fi
if ! getent group "$WSL_USER_GROUP" &>/dev/null; then
    echo "Error: Group $WSL_USER_GROUP does not exist.">&2

    exit 1
fi
if ! groups "$WSL_USER_NAME" | grep -q "\b$WSL_USER_GROUP\b"; then
    echo "Error: User $WSL_USER_NAME is not a member of group $WSL_USER_GROUP.">&2

    exit 1
fi
```

karmendra:
My idea for wsl:folders was just to replicate the feature from vagrant.
But the problem with approach of code hosted in windows and linked in ubuntu is that it is really slow to render these webpages.

Website load much faster if the code is in Ubuntu's file system. Problem here is code would be lost if wsl is unregistered, unlike the previous approach. 

I will work on wsl:backup command that can easily backup the sites and database in future.

Regarding the validations, I will go ahead and add those validations in wsl-init.

Thanks.

svpernova09:
Giving this a review now. Sorry for the delays. I found a bug in the MariaDB installer so far. Should be good to merge this soon.

svpernova09:
I think we want to keep the WSL config separate from the Homestead Configuration. I traditionally have a giant Homestead.yaml that gets configured and passed around various machines and configurations. 

At some point this WSL stuff will need to get pulled out of Homestead because it's not really.... Homestead in the Vagrant sense.

karmendra:
Hi @MichaelBelgium 

> 
> Lastly the checks you did before in wsl init, those were good, i'd keep those
> 

I updated the wsl-init to add those checks.

P.S. Sorry for the delay, I was on long vacation.

wsl.md
```diff
@@ -0,0 +1,277 @@
...
+kali-linux                             Kali Linux Rolling
+Ubuntu-18.04                           Ubuntu 18.04 LTS
+Ubuntu-20.04                           Ubuntu 20.04 LTS
+Ubuntu-22.04                           Ubuntu 22.04 LTS
+...
+```
+
+To install a distribution, use the following command:
+```powershell
+> wsl --install Ubuntu-22.04
```
sdespont, line 66: Should be wsl --install -d Ubuntu-22.04 isn't it ?

MichaelBelgium:
> Hi @MichaelBelgium
> 
> > Lastly the checks you did before in wsl init, those were good, i'd keep those
> 
> I updated the wsl-init to add those checks.

@karmendra  I don't see the checks in this PR?

wsl.md
MichaelBelgium: @sdespont What does the -d option do? I can't even find it in the docs

[The docs](https://learn.microsoft.com/en-us/windows/wsl/basic-commands#install) say `wsl --install <Distribution Name>`

sdespont: The -d stands for "distribution", it permits to keep compatibility for older systems. Also, I am using Windows 23H2 but I need to specify the -d in the install command.
image

karmendra:
> @karmendra I don't see the checks in this PR?

Thanks @MichaelBelgium  for checking this, I thought I have done the changes but looks like I have forgotten to push my changes.
image
I just pushed my changes, please have a look.

PR #1965, opened 2024-04-12
fix: Ensure php8.3-fpm service is correctly enabled and restarted

minutiae:


svpernova09 (approved):
Thanks for the update

PR #1967, opened 2024-04-19
Fix Apache Proxy Websockets

ALameLlama:
This adds support for web sockets in the Apache proxy.

It defaults to `ws://127.0.0.1` and if you only included a port it will be `ws://127.0.0.1:$2` otherwise, it check `$2` has HTTPS and adds the correct web socket protocol


svpernova09:
Thanks!

PR #1972, opened 2024-04-27
Fix reload provision when using php 5.6

lloricode:
it throws an error when using php 5.6

sample

```yaml
    - map: local.test
      to: /home/vagrant/code/local
      php: "5.6"
```

```
    homestead: Composer 2.3.0 dropped support for PHP <7.2.5 and you are running 5.6.40-68+ubuntu22.04.1+deb.sury.org+1, please upgrade PHP or use Composer 2.2 LTS via "composer self-update --2.2". Aborting.
went wrong.
```


svpernova09 (approved):
Thanks!

PR #1983, opened 2024-07-20
Fix features options setting WslApplyFeatures.php

sdespont:
Because the features options are set in the environnement variables in https://github.com/laravel/homestead/blob/98a798b18a078121252f0ecfe3e10cba050ce504/scripts/homestead.rb#L308-L311

it must be the same in the WSL code


sdespont:
Could you consider merging this request?

sdespont:
@svpernova09 Could you consider merging this request?

